PR #13, opened 2016-06-10
Update `plot_carrier_production`

brynpickering:
requires specification of `demand` when running the `plot_timeseries` function from `plot_carrier_production`, otherwise it will fail on any carrier that isn't `power` (as it defaults to searching for `demand_power`).


sjpfenninger:
Thanks!

PR #14, opened 2016-09-05
Adding eff_ref exception to `get_eff_ref` when loading e_eff from file

brynpickering:
Just a simple exception raised so the user is made aware of the current issues with eff_reff not working from default if e_eff is loaded from file. Previously the warning wasn't specific.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ calliope.egg-info/
 doc/_build/
 dist/
 .cache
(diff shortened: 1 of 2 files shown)


sjpfenninger:
Can you remove commit ef8c893? That seems to have slipped in by accident

brynpickering:
Noticed that you sorted this late July, so not a necessary addition. Will update my master accordingly

PR #19, opened 2016-10-06
create get_distance function

brynpickering:
If metadata file exists and no distances are given for links in any yaml file, then use metadata to infer distances (in integer meters). Function updates config_model during model initialisation


brynpickering:
Closed until working with all tests

PR #28, opened 2016-12-09
Added timeseries constraints & revenue generation

brynpickering:
Can now load from file more than just `r` and `e_eff`. This includes other efficiencies and some other constraint parameters, but also cost and revenue. Revenue can be generated by use of an additional technology setting `revenue` which offers subsidies based on `e_cap` (`sub_cap`), `e_prod` (`sub_var`) and annual output (`sub_annual`).


brynpickering:
We good to merge this in now that there is some profiling capability?

brynpickering:
Closed as out-dated duplicate of PR #30

PR #34, opened 2017-02-10
Move revenue streams under costs

brynpickering:
This PR is really about how you set up the revenue definition, either as its own class, or under costs. It doesn't change model generation time, although it does clean things up a bit:

- we can return to just having one cost class: k, rather than kc and kr
- checking timeseries data and logging it in model.data is nicer

All revenue streams are still prefixed with `sub_`, meaning subsidy, but might be better with a different, clearer prefix.


calliope/config/defaults.yaml
```diff
@@ -85,8 +85,7 @@ techs:
                 om_var: 0  # Variable O&M costs (per kWh of ``es_prod``)
                 om_fuel: 0  # Fuel costs (per kWh of ``r`` used)
                 om_rb: 0  # Fuel costs for secondary resource (per kWh of ``rb`` used)
-        revenue:
-            default:
+                ## Revenue generating subsidies for production
                 sub_var: 0 #variable subsidy (per kWh of ``es_prod`` or ``es_con``)
```
sjpfenninger: Should we use ``rev_`` to make it more generically about revenue rather than just subsidies?

calliope/core.py
```diff
@@ -926,62 +907,60 @@ def read_data(self):
...
+                            option_x = self.get_cost(param, y, k, x=x)
+                            if option != option_x:
+                                df[k].loc[:, x] = self._read_param_for_tech(
+                                                            param, y, time_res,
+                                                            option_x, x=x)
+
+                        self._validate_param_df(param, y, df[k])  # Have all `x` been set?
+                        # Create 3D dataframe with axes ordered as (t, x, k)
+                    df_k = pd.Panel(df).swapaxes(0,1).swapaxes(1,2)
+                    param_data[y] = xr.DataArray(df_k, dims=['t', 'x', 'k'])
```
sjpfenninger: Should not go via `Panel` but something like this, replacing `df` with a more descriptive variable name (`cost_ts`):

```
cost_ts = {k: xr.DataArray(v, dims=['t', 'x']) for k, v in cost_ts.items()}
cost_ts = xr.Dataset(cost_ts).to_array(dim='k')

```

calliope/core.py
```diff
@@ -926,62 +907,60 @@ def read_data(self):
...
-                    # that is different from the generic option, and if so,
-                    # update the dataframe
-                    option_x = self.get_option(k, x=x)
-                    if option != option_x:
-                        df.loc[:, x] = self._read_param_for_tech(param, y, time_res, option_x, x=x)
-
-                self._validate_param_df(param, y, df)  # Have all `x` been set?
-
-                param_data[y] = xr.DataArray(df, dims=['t', 'x'])
+            if 'om' in param or 'sub' in param: # cost constraints
```
sjpfenninger: This whole `if-then` block is a bit of a mess (it already was before) and should probably be rewritten at some point, particularly now that a lot of the code is duplicated... but I guess for now it's ok, as long as it works 🙂

calliope/config/defaults.yaml
brynpickering: Yeah, I'd be happy with that: `rev_var`, `rev_cap`, `rev_annual`, and `rev_export` (if that ends up being included)

calliope/core.py
brynpickering: Happy to update it, haven't really spent time working with xarray DataArrays, so was falling back on my pandas knowledge there.

brynpickering:
I'll update as necessary for the specific comments.
As to generalising, I'll have another go with allowing cost to be any real number. It had convexity complaints from what I remember, but I don't really know why it would at that point when it doesn't at the objective definition point (in which the resulting objective function can be any real number, allowing it to go very negative if it can get a good revenue stream)

sjpfenninger:
Happy for this to be merged into master now unless you want to keep it open while having another go at allowing cost to be any real number

brynpickering:
OK, so costs _can_ be allowed to be any real number, no issues (not sure why I didn't pursue this further, earlier). The only problem is that there are many instances where capacities are given infinite limits, in the knowledge that, for cost minimisation, technology capacity will match maximum production when a cost is put in place. The model fails due to unbound capacities if a negative capacity cost is put in place, which particularity affects r_cap. When `.max` is assigned for the relevant capacities, there are no issues with putting capacity revenue in place, it just means the capacity *will* be the maximum possible capacity!

sjpfenninger:
Am I correct that this is superseded by #34?

brynpickering:
You are, this PR is no longer relevant.

PR #35, opened 2017-02-15
Define revenue as a negative cost

brynpickering:
At model definition, costs can be assigned negative values, providing a revenue stream. This allows removal of the 'revenue' class and the associated cost types. All cost types can be negative, but any relating to capacity have to have that capacity maximum defined.


brynpickering:
Current failing CI is a result of MD5 mismatch, exactly as described here: https://github.com/conda/conda/issues/4642
Hopefully will be resolved soon. This PR passes tests locally.

PR #36, opened 2017-03-30
New technology definitions

brynpickering:
Updating Calliope to have new parent technologies which can be used to create more distinctive sets and constraints, reducing model generation time.


sjpfenninger:
Merged via bd5cbc9

PR #44, opened 2017-05-12
Update map plotting for x-y and lat-lon

brynpickering:
Metadata can be supplied as lat-lon, and is then plotted with Basemap background. But, if you want a quick 'abstract' model (i.e. not geographically located, but there are distinct locations with distances between them) you can define as x-y. This allows for that.

Also updates metadata map boundary to be `[lat1, lon1, lat2, lon2]` instead of `[lon1, lat1, lon2, lat2]` to be in line with the definition of the other location lat-lons (i.e. `[lat, lon]`). For cartesian, this becomes [x1, y1, x2, y2], but isn't actually used at the moment.

### DIFF
--- a/calliope/analysis_utils.py
+++ b/calliope/analysis_utils.py
@@ -150,28 +150,39 @@ def plot_graph_on_map(config_model, G=None,
     import networkx as nx
     from calliope.lib import nx_pylab
 
+    coord_system = config_model.metadata.get('coordinate_system', 'geographic')
+
     # Set up basemap
     if not bounds:
         bounds = config_model.metadata.map_boundary
-    bounds_width = bounds[2] - bounds[0]  # lon --> width
-    bounds_height = bounds[3] - bounds[1]  # lat --> height
-    m = Basemap(projection='merc', ellps='WGS84',
-                llcrnrlon=bounds[0], llcrnrlat=bounds[1],
-                urcrnrlon=bounds[2], urcrnrlat=bounds[3],
-                lat_ts=bounds[1] + bounds_width / 2,
-                resolution=map_resolution,
-                suppress_ticks=True)
-
-    # Node positions
-    pos = config_model.metadata.location_coordinates
-    pos = {i: m(pos[i][1], pos[i][0]) for i in pos}  # Flip lat, lon to x, y!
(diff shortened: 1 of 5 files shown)


sjpfenninger:
All good, except uncertain about map bounds. I think it's more consistent to stick to ``[lon1, lat1, lon2, lat2]`` so that it matches what the Cartesian equivalent would be, ``[x1, y1, x2, y2]`` -- does that not make more sense?

brynpickering:
it would, but only if you change all the entries to [lon, lat], which
they're currently not. Most people think in lat, lon (is there a standard
for it?) so the map bounds being [lat1, lon1, lat2, lon2] seems to be more
intuitive, no? Maybe more opinions are needed on this!

sjpfenninger:
You are right. And there is a standard -- [ISO 6709](https://en.wikipedia.org/wiki/ISO_6709) -- which says that the order should be lat,lon.

So, happy for this to be rebased onto master as is!

brynpickering:
I'd much prefer it to be standard as lon, lat mind you! I can see confusion
arising either way...

sjpfenninger:
Perhaps a more verbose definition of locations: ``region1: {'lat': 40, 'lon': -2}`` or ``region1: {'x': -2, 'y': 40}``, so it doesn't matter which order the keys are in. Then treat `lat` and `y` as synonyms and either let one take precedence if both defined or warn if a region defines both.

Then ``map_boundary`` is specifed via two points defined either by ``x/y`` or ``lat/lon``, e.g.:

```
map_boundary:
    lower_left: {'lat': 35, 'lon': -10}
    upper_right: {'lat': 45, 'lon': 5}

```

brynpickering:
That would probably work better actually. Then in `analysis_utils` you can avoid the ambiguity of saying `bounds[1]`, `pos[i][0]` etc. by saying `bounds.lower_left.lat`, `pos[i].lon` etc.
I'll try it out and update the branch.

brynpickering:
Updated, see what you think @sjpfenninger

sjpfenninger:
Looks good to me! Happy for this to be rebased onto master. It looks like the lint job is failing once again due to some dependency rather than an actual issue, so let's just ignore that.

PR #103, opened 2018-04-27
Objective function cost class/sense definition in run config

GraemeHawker:

* Objective function altered to accept cost_class and sense parameters
* Backend amended to pass through parameters from model run config
* Additional checks on run config

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/calliope/backend/pyomo/model.py
+++ b/calliope/backend/pyomo/model.py
@@ -137,9 +137,16 @@ def generate_model(model_data):
     #         self.add_constraint(load_function(c))
 
     # Objective function
-    objective_name = model_data.attrs['run.objective']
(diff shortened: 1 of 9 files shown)


brynpickering:
Thanks for the PR @GraemeHawker! 

It would be good to have a test that checks that the warnings you've added are triggered. I.e. if you use the ``check_feasibility`` objective it will spit out a warning that you can catch with `pytest.warns()`. Additionally worth adding a random key into `objective_options` to check that it is caught earlier on. Maybe add the latter to ``test_core_preprocessing.py::TestChecks::test_unrecognised_model_run_keys``

PR #109, opened 2018-05-04
Fix negative om_con costs in conversion techs

sjpfenninger:

* Fix negative om_con costs in conversion techs

Reviewer checklist:

- [x] Changelog updated


brynpickering:
@sjpfenninger well done on the first PR!
RE tests:
1. You'll need to update the constraint sets (see ``test.common.constraint_sets.yaml``) to account for the addition of a variable conversion tech cost (`om_con` for boiler).
2. Very odd that the additional boiler variable cost leads to greater production from the boiler - we'll need to track down why that is before merging this PR in...

brynpickering:
Addition of `om_con` to boiler in urban scale example model causes the tests to fail. It fails on asserting non-cost related results (e.g. energy cap), where we might have just expected the cost to change and nothing else. Weirdly, making the boiler more expensive leads to greater boiler purchase. However, the overall objective function value increases in line with the increased cost we would expect from the additional `om_con` cost (~6.7). I'd be happy for you to change the ground truth for this example, such that the tests pass. Something to keep an eye on, but I expect it is simply a quirk of the optimisation.

sjpfenninger:
Ok, updated the failing test accordingly, and we'll monitor for any issues

PR #116, opened 2018-05-31
Add unit_capacity_systemwide_constraint

brynpickering:
``units_max_systemwide`` and ``units_equals_systemwide`` can be applied to an integer/binary constrained technology (capacity limited by ``units`` not ``energy_cap``, or has an associated ``purchase`` (binary) cost). Constraint works similarly to existing ``energy_cap_max_systemwide``, limiting the number of units of a technology that can be purchased across all locations in the model.

Reviewer checklist:

- [x] Documentation updated
- [x] Changelog updated

### DIFF
--- a/calliope/backend/pyomo/constraints/milp.py
+++ b/calliope/backend/pyomo/constraints/milp.py
@@ -10,6 +10,7 @@ Constraints for binary and integer decision variables
 """
 
 import pyomo.core as po  # pylint: disable=import-error
+import numpy as np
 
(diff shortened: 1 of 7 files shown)


brynpickering:
@sjpfenninger docs are automatically updated by the addition of the constraint docstring and the default.yaml entries, so I don't think additional documentation is required

sjpfenninger:
Fair enough, we don't explicitly mention any of the _systemwide constraints in the docs now, but they are there in the auto-built parts

PR #123, opened 2018-06-14
Add function plot_energy_flows

FirstWorldSlardar:
Adding a new plot function showing the energy produced in locs, aside
with the energy transmitted by transmission lines. Energy is ordered by
techs. A slider is used to navigate through timesteps.

* solved_model.plot.energy_flows() 

Reviewer checklist:

Aware of not adding tests for now. As I am still not comfortable with the example models, I don't wan't to make a unvalid test pass.
~~- [ ] Documentation updated~~ (*will be added later*)


FirstWorldSlardar:
Aside the violations, there is in the legend points, corresponding to the invisible markers I created for each transmission line in order to display information about the line on hover. They shouldn't be there, putting a showlegend=False in the dict should be sufficient.

brynpickering:
So the easiest is to push commits to fix the Hound syntax violations, and to update the visibility of the legend. On pushing those commits, Hound will re-run automatically!

FirstWorldSlardar:
Oh cool, I'll do it

brynpickering:
Thanks for the updates @FirstWorldSlardar ! It would be good to have the file completely covered by tests, which only requires 4 lines to be dealt with:
1. (198, 199) just need you to test subsetting to two timesteps,
2. for (205, 206) you can catch the Attribute error with `with pytest.raises(AttributeError) as error`, then use the function `check_error_or_warning()` to see if the error you expect is raised.

I'll leave @sjpfenninger to look in more detail on the actual implemented functionality ;)

FirstWorldSlardar:
I'll do it, no problem

sjpfenninger:
Ok, this looks really great, thanks a lot for this!

I see one immediate problem when testing: techs or transmission lines that exist but are not used in a specific time step are not plotted at all. They pop into existence when they are used, then disappear again, leading the plot area to change size as you move through timesteps. I think that existing but unused components should always be plotted to get around this, probably in a light grey. You can see this issue by plotting the built-in example, ``calliope.examples.national_scale()``.

A more minor issue is the legend: how difficult is it to make the legend markers fixed in size rather than changing with the actual plot elements (maybe something @brynpickering knows about)?

FirstWorldSlardar:
Thank you for your feedback :)
Having markers in the legend of fixed size is definitely needed, I agree.
And indeed, I deleted all the 0 and NaN data resulting in techs not used not being rendered, what I think is a good thing, especially when techs/locs/links increases, as it would pollute the visual representation of the energy flows. A parameter for this would be the best option. 

I think I know now what make the plot changing size: when a techs of name longer than previous displayed techs names pops, the legend increases horizontally in size (national scale: the first time the 'free_transmission:region1-1' appears for example). Actually, the size of the markers in the legend are not the cause of the problem as their overflow is hidden by the legend box (national scale: look at the ccgt techs).

brynpickering:
RE the point on having fixed width lines in the legend: you could duplicate each trace (trace1) with another trace (trace2) that has no data, but has the same `legendgroup` as trace1. Then `showlegend=False` for trace1.

RE long names popping up: there is a helper function in `util.py` that allows you to ensure legend strings longer than X in length are broken across multiple lines. You could use that to try and ensure new legend items don't extend too far

FirstWorldSlardar:
@sjpfenninger yes, an optional setting doesn't cost much to be implemented and could be helpul.
@brynpickering After you comment and a few tests I realised that the default font-family in plotly is not monospaced haha ! Why would I count the length of the string if chars are not displayed the same size ! So I changed my layout.font.family to 'Courier New, Droid Sans Mono' (both are monospaced, maybe they could be use for the other graphs) and, using your function solved this problem.

brynpickering:
@FirstWorldSlardar I have a few changes that I've made locally to this branch, if you'd be happy to accept me pushing them in?

1. I wasn't sure about the monospaced font choice ;) Open Sans is a bit more 'friendly', for those who are looking for the visual appeal over utility. So, I instead applied a fixed plot margin, stopping the plot from changing size with a changing legend size. To stop the legend from ever overlapping the plot, I set the width to be large enough for a large character in Open Sans.

2. I changed the legend items to only appear once for each technology and to use the longhand technology names (e.g. `Combined Cycle Gas Turbine` instead of `ccgt`).

3. I moved about some of the functions to util.py in analysis, as they are shared functions with other files. That way, e.g. name breaking isn't duplicated in different files. I updated util.py to reflect your updates to the functions. 

If that seems reasonable I'll just push the commit directly to the PR. You can always push a new commit to make edits after that!

FirstWorldSlardar:
Do as many changes as you want of course ! ;)

brynpickering:
Great, see what you think!

calliope/analysis/plotting/__init__.py
```diff
@@ -2,3 +2,4 @@
 from calliope.analysis.plotting.capacity import plot_capacity
 from calliope.analysis.plotting.timeseries import plot_timeseries
 from calliope.analysis.plotting.transmission import plot_transmission
+from calliope.analysis.plotting.energy_flows import plot_energy_flows
```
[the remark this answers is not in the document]
sjpfenninger: Can you change the name ``plot_energy_flows`` / ``energy_flows`` to just ``flows`` or ``flow`` to keep with the rest of the plotting function names, which also don't contain "energy" in the name (and considering that you could be modelling carriers that are not energy-related)

calliope/analysis/plotting/energy_flows.py
```diff
@@ -0,0 +1,319 @@
+"""
```
sjpfenninger: Again, can you change the filename to ``flows.py`` or ``flow.py`` (and function names accordingly)?

calliope/analysis/plotting/energy_flows.py
```diff
@@ -0,0 +1,319 @@
...
+        return [line, line_legend, line_info_marker]
+    else:
+        return [line, line_info_marker]
+
+
+def _marker(loc_coordinates, location, carrier, tech, prod, scale_factor,
+            techs_colors, is_initial_timestep, add_legend, name):
+    hover_info = "%s: %.2f of %s (%s)" % \
+        (location, prod, tech, carrier)
+    # ex: "Region1: 3552.65 of pipe_import (gas)"
```
sjpfenninger: Can you move the comment to above the line it is commenting on, and expand to ``Example hover_info: "Region1: ..."``

calliope/analysis/plotting/energy_flows.py
```diff
@@ -0,0 +1,319 @@
...
+
+def plot_energy_flows(model, timestep_cycle=1,
+                      timestep_index_subset=[], **kwargs):
+    """
+    Parameters
+    ----------
+    timestep_cycle : int, optional
+        Shows one of every timestep_cycle timesteps. Default is 1 (all timesteps
+        are shown).
+    timestep_index_subset : list of int, optional
```
sjpfenninger: Could this subsetting be done in a way more similar to the other plotting functions? If not with a generic ``subset`` param (see snippet below from ``plot_timeseries``) at least with the ability to pass a datetime range or a start and end datetime like ``['2018-01-01', '2018-01-31']``?

```
    subset : dict, optional
        Dictionary by which data is subset (uses xarray `loc` indexing). Keys
        any of ['timeseries', 'locs', 'techs', 'carriers', 'costs'].
```

FirstWorldSlardar:
@brynpickering  Neat ! Using longhand names is just clearer, me copying the break_name function was ugly, nice edits
I have remarks not related to your changes:
- too many points in plotly result in poor performance, and I almost doubled the number of required points by adding associated legend markers.
- I just noticed that when there is a flow in both direction on the same transmission line, one line_info_marker gets written over by the other, resulting in one not being able to see info on hovering a line (for one of the direction flow).

@sjpfenninger I'll clean and rename files today. For the parameter displaying non used techs, I'm not sure how to implement this yet; what color should I use and  what is the marker size to set for each techs or transmission line ?

brynpickering:
I think the associated legend markers won't add so much, as they don't have any data associated with them. RE the overlap - well spotted! We could take the net flow for all transmission links instead of considering the two directions separately? Then change `>` to `<` if the direction of flow changes?

p.s. might be nicer to have `->` and `<-` than just  `>` , `<`

calliope/analysis/plotting/energy_flows.py
FirstWorldSlardar: Yes it should be done like this, I agree

FirstWorldSlardar:
Easy enough. I will have a look at this soon.
However, this issue made me think of another one: when two lines of different techs superimpose, the info_marker of just one stays visible (when it is the same tech we can collapse the two flows but for different ones it doesn't make sense and the problem remains).
I don't think we can have a quick solution to this with plotly, that's why not having opacity to 1 is important, so that the user can at least see what is happening.

brynpickering:
Hmm, yeah. I had one idea to solve this overlapping problem at locations:

https://files.gitter.im/brynpickering/YHec/image.png

i.e. create a pie chart of the capacity, centred on a location. As time steps change, the relative contribution of each technology can change on the pie chart. Would probably be difficult to achieve, much like your hover info issue. As a solution for the time being, users can 'hide' technologies (using the legend) to see technologies underneath. Maybe we push these additional ideas to plot.flows v.2 ;)

brynpickering:
@FirstWorldSlardar Did you get anywhere with collapsing flows into one?

FirstWorldSlardar:
@brynpickering Done. It wasn't so easy at the end ;). I tested it on my model, and it removes adjust correctly the flows.

sjpfenninger:
@FirstWorldSlardar @brynpickering ok let me re-review this (maybe Friday, else early next week). Thanks for all the work so far! Should be merged soon now :)

FirstWorldSlardar:
I agree with your findings, it is something to keep in mind if we move towards a Dash interface (and more complex Plotly plots)

sjpfenninger:
Great work @FirstWorldSlardar ! :star: :star: :star:

I squashed this into a single commit to get rid of the various files that were added and then removed again. Looking forward to adding bits and pieces of extra functionality to this and to other ideas you have.

PR #124, opened 2018-06-17
Add `excess_supply` variable when ensuring feasibility

brynpickering:

* System can be balanced due to forced excess supply, by the same mechanism as we currently handle unmet demand. 
* `excess_supply` decision variable is merged into `unmet_demand` on completion, leading to just the one data variable in `model.results`. However, it's possible for `unmet_demand` to be negative (i.e. excess supply).

Reviewer checklist:

- [x] Changelog updated

### DIFF
--- a/calliope/backend/pyomo/constraints/energy_balance.py
+++ b/calliope/backend/pyomo/constraints/energy_balance.py
@@ -106,15 +106,16 @@ def system_balance_constraint_rule(backend_model, loc_carrier, timestep):
 
     """
     prod, con, export = get_loc_tech_carriers(backend_model, loc_carrier)
-    if hasattr(backend_model, 'unmet_demand'):
+    if hasattr(backend_model, 'bigM'):
         unmet_demand = backend_model.unmet_demand[loc_carrier, timestep]
+        excess_supply = backend_model.excess_supply[loc_carrier, timestep]
     else:
-        unmet_demand = 0
+        unmet_demand = excess_supply = 0
 
     backend_model.system_balance[loc_carrier, timestep].expr = (
         sum(backend_model.carrier_prod[loc_tech_carrier, timestep] for loc_tech_carrier in prod) +
         sum(backend_model.carrier_con[loc_tech_carrier, timestep] for loc_tech_carrier in con) +
-        unmet_demand
+        unmet_demand + excess_supply
     )
 
     return backend_model.system_balance[loc_carrier, timestep] == 0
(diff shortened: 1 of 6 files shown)


calliope/backend/pyomo/variables.py
```diff
@@ -92,4 +92,5 @@ def initialize_decision_variables(backend_model):
 
     if model_data_dict['attrs'].get('run.ensure_feasibility', False):
         backend_model.unmet_demand = po.Var(backend_model.loc_carriers, backend_model.timesteps, within=po.NonNegativeReals)
+        backend_model.excess_supply = po.Var(backend_model.loc_carriers, backend_model.timesteps, within=po.NegativeReals)
```
sjpfenninger: I wonder if there's a better name than ``excess_supply`` given that its negative, and an excess implies something >0. Doesn't matter hugely though

calliope/backend/pyomo/constraints/energy_balance.py
```diff
@@ -106,15 +106,16 @@ def system_balance_constraint_rule(backend_model, loc_carrier, timestep):
 
     """
     prod, con, export = get_loc_tech_carriers(backend_model, loc_carrier)
-    if hasattr(backend_model, 'unmet_demand'):
+    if hasattr(backend_model, 'bigM'):
```
sjpfenninger: Are we sure that bigM is only ever an attribute if ``ensure_feasibility`` is set? There should be a comment here at least to explain why this check for bigM happens, since it's not immediately obvious?

calliope/backend/pyomo/variables.py
brynpickering: I wondered about changing them to `demand_slack` and `supply_slack` and then generalising it to `slack` in the results Dataset, but dropped that idea as `slack` is perhaps unintuitive outside the optimisation field. `unused_supply` also came to mind. I'll have a think

calliope/backend/pyomo/constraints/energy_balance.py
brynpickering: I could check for `ensure feasibility` here actually, I'll change that

calliope/backend/pyomo/variables.py
sjpfenninger: ``unused_supply`` seems like it fits the bill

brynpickering: Updated in df65948

calliope/backend/pyomo/constraints/energy_balance.py
brynpickering: Updated in df65948

PR #140, opened 2018-08-30
Update xarray dependency to 0.10.8

brynpickering:

* xarray pinned to 0.10.8
* xarray dataset timestep resampling updated to work with 0.10.8

This doesn't address the issue that @pmmeyourmodel had with 0.10.7, in which the dtype of some variables in the dataset was 'object', causing an error when dumping it to netcdf. I can't reproduce this issue, so don't know if it is fixed or not.

Reviewer checklist:

- [x] Documentation updated
- [ ] Changelog updated


pmmeyourmodel:
Using `xarray 0.10.8` I still get the netCDF error:

```
Error in _nc4_dtype:
unsupported dtype for netCDF4 variable: object
```

brynpickering:
Thanks for checking this @pmmeyourmodel. Could you load the model in Python, so that you can try saving the model to netcdf *before* running it:

```python
model = Calliope.Model(path_to_model.yaml)
model.to_netcdf(path_to_netcdf.nc)
```

Additionally, if you could check which of your data variables are defined as 'object' by running ``model.inputs`` or ``model.results`` (after running the model) that would be great. If it's easier, and you're happy to do so, you could PM me your model ;) I'd be happy to check through it for the source of the issue.

pmmeyourmodel:
There is no error if I save the model before running it. I've tried this a few times now with different override groups.

`model.inputs` after running gives me one data variable which is an object, that's `lookup_loc_techs_conversion_plus`. 

`model.results` after running only gives me data variables of type `float64`.

However I am not certain if this is the right information for you. You will receive an E-Mail from me with my model soon. Damn, I thought it's going to be the other way around ;-) ;-)

PR #150, opened 2018-10-04
Re-add cost key to demand tech_group

FLomb:
Added a cost key (om_prod) to the demand tech_group, as in prior versions of Calliope, so as to allow for modelling of import and export (and possibly curtailment and more) as demand techs

Fixes issue(s) #
- Impossibility to model export as a single entity and curtailment

* om_prod is added as a cost key to the demand tech_group

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/calliope/config/model.yaml
+++ b/calliope/config/model.yaml
@@ -64,7 +64,7 @@ tech_groups:
     demand:
         required_constraints: ['resource']
         allowed_constraints: ['energy_con', 'force_resource', 'resource', 'resource_area_equals', 'resource_scale', 'resource_unit']
-        allowed_costs: ['om_prod']
+        allowed_costs: ['om_con']
         essentials:
             parent: null
         constraints:
--- a/calliope/config/model.yaml
+++ b/calliope/config/model.yaml
@@ -64,7 +64,7 @@ tech_groups:
     demand:
         required_constraints: ['resource']
         allowed_constraints: ['energy_con', 'force_resource', 'resource', 'resource_area_equals', 'resource_scale', 'resource_unit']
-        allowed_costs: []
+        allowed_costs: ['om_prod']
         essentials:
             parent: null
         constraints:


brynpickering:
Hi @FLomb. Thanks for the PR. This will need to have tests associated with it, to make sure a model that has a cost associated with demand correctly passes that cost to the optimisation. Additionally, a test to ensure a demand technology with `force_resource: False` works as expected (e.g. as an 'export' node) would be good. 

It's a good chance to try out adding and running tests locally! If you do so (navigate to your calliope folder from the command line and run 'make test' or 'pytest') you'll find that it fails on one test at the moment, which is specifically checking for the (currently) non-allowed `om_prod` for demand.

sjpfenninger:
@FLomb Thanks also for the PR - and let me know if you need help with the tests for this.

brynpickering:
Hi @FLomb, although a valid method to 'fix' the tests, your update just sidesteps existing tests to slip through the gaps... As I mentioned in the previous comment, two new tests are really needed:

1. An example where a demand cost is used in a model, to check it makes its way through to the Pyomo model and that test model results are as expected.
2. An example where `force_resource` is set to `False` for a demand technology. We had no reason to test this before, because it was not expected behaviour for demand techs. But, your change suggests you want to be able to set this.

p.s. I think having  `om_con` as the relevant cost is correct, so your update makes sense whether or not it was applied to just sidestep the failing test ;)

FLomb:
Hi, some updates about this:
- I managed to apply the change on my local dev version, and from some preliminary tests it seems to work as expected from an energy point of view, i.e. it works as a "sink" of energy (with` force_resource` set to False);
- On the other hand, it looks like the `om_con` costs of the technology are not taken into account by the optimisation function. The "sink" is used by the model if and only if there's such an excess of energy that it cannot be stored or supplied elsewhere. To this regard, I gave a look at "costs.py" in the folder \pyomo\constraints, and it seems to me (lines 211-220) that `om_con` costs are only evaluated for supply and supply_plus techs, plus an apparent redirection to conversion.py and conversion_plus.py (line 54) for the calculation related to the latter. In short, it seems like demand techs are (as one could have expected) not accounted for in the variable cost calculation. Is it like this or did I miss where they are counted?

FLomb:
Fixed on other branch and created new updated pull requst

PR #154, opened 2018-10-08
Add conda_create util; test Py 3.5/3.6 but not 3.7

sjpfenninger:

* Add conda_create util
* Test Py 3.5/3.6 but not 3.7

Reviewer checklist:

- [ ] Tests on Travis continue to work
- [ ] Tests on Appveyor continue to work

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,27 +4,34 @@ build: false
 environment:
   matrix:
     - PYTHON: "C:\\Miniconda3"
-      PYTHON_VERSION: "3.6"
+      PYTHON_VERSION: "3.5"
       PYTHON_ARCH: "32"
+    - PYTHON: "C:\\Miniconda3-x64"
+      PYTHON_VERSION: "3.5"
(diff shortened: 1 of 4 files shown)


brynpickering:
If lots of Python3.6 skips are required in tests, does that mean we're no longer <3.6 compatible? If that's the case, we should probably not test <3.6 versions & we should pin the Python version to 3.6...

sjpfenninger:
It seems only due to some more recently added tests assuming dicts are ordered, which they are not in <=3.5. So I'm skipping those to avoid rewriting them (since anyway, dicts are officially ordered as of 3.7).

Mainly, testing against 3.5 and 3.6 is prep for adding 3.7 again once Pyomo is properly compatible and making sure multi-version testing actually properly works in the meantime.

The docs still say 3.6 is required. We should probably have some official policy, like always supporting the most recent two versions of Python!

brynpickering:
Could you place the function wrapper into the test `utils.py` file, instead of defining it in each file? Otherwise happy for this to be merged!

PR #156, opened 2018-10-17
Fix assert time series

timtroendle:
### Summary of changes in this pull request:

Calliope needs at least one time series otherwise the model preprocess failed with unspecific error messages.

Now there is a check that at least one time series exists and the preprocess fails gracefully with a specific error message in case none exists.

###  Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


timtroendle:
By the way, considering that one _has_ to have a time series, the time series documentation is rather hard to find: it's below `more info` _and_ `advanced technology`. One could argue to move that into the prominent and top section `building a model`.

sjpfenninger:
Thanks @timtroendle!

Issue #182, opened 2019-01-15
(Optionally) communicate non optimal solution through CLI


timtroendle:
## Problem description

I am using Calliope in an automated workflow. To me, a non-optimal solution is a fail and I would like my workflow to stop at that point. But because Calliope CLI returns a `0 / success` my workflow continues, which is ~~sub-optimal~~ not what I want. 

Calliope _could_ return a fail whenever the solution is non-optimal. Would it make sense to build that (potentially as an option) into the CLI?

## Calliope version

0.6.3

brynpickering:
That would be a simple addition. At the moment we capture the termination condition to decide whether plots should be generated or not (if a user used `--save_plots`), it would just require a conditional for something like `--break-on-fail` to return a Calliope `ModelError` rather than returning `None`.

timtroendle:
Yep and I could build it in. Just wanted to know whether you think it makes sense; for example is non-optimal = fail true? For me yes, but for others as well?

timtroendle:
And should it be optional, or is it always a good idea?

sjpfenninger:
I think non-optional = fail could be the default, with an option to turn it off.

brynpickering:
Yeah, you might like your automated runs to keep going and have the lack of optimality caught later down the line (given that we include in the model attributes using `termination_condition` whether it was optimal or not). But, as Pyomo gives no useful information back for a non-optimal run, I agree that the default should probably be fail.

timtroendle:
Implementation hint (for myself): looks like the `click` way of doing this is to raise an exception: https://stackoverflow.com/a/26247420/1856079.

PR #183, opened 2019-01-16
Change exit code to no success when infeasible

timtroendle:
This is a breaking change.
This fixes https://github.com/calliope-project/calliope/issues/182.

Fixes issue(s) #182 

Reviewer checklist:

- [ ] ~~Documentation updated~~
- [x] Changelog updated

### DIFF
--- a/calliope/cli.py
+++ b/calliope/cli.py
@@ -1,5 +1,5 @@
 """
-Copyright (C) 2013-2018 Calliope contributors listed in AUTHORS.
+Copyright (C) 2013-2019 Calliope contributors listed in AUTHORS.
 Licensed under the Apache 2.0 License (see LICENSE file).
 
 cli.py
@@ -25,6 +25,7 @@ from calliope import AttrDict, Model, examples, read_netcdf
 from calliope._version import __version__
 from calliope.core.util.generate_runs import generate
 from calliope.core.util.logging import logger, set_log_level
+from calliope.exceptions import BackendError
 
 _time_format = '%Y-%m-%d %H:%M:%S'
 
@@ -55,6 +56,13 @@ _profile_filename = click.option(
     '--profile_filename', type=str,
     help='Filename to save profile to if enabled --profile.'
 )
+
+
+_fail_when_infeasible = click.option(
+    '--fail_when_infeasible/--no_fail_when_infeasible', is_flag=True, default=True,
+    help='Return fail on command line when problem is infeasible (default True).'
+)
+
 
 if logger.hasHandlers():
     for handler in logger.handlers:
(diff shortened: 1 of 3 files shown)


calliope/cli.py
```diff
@@ -259,7 +268,10 @@ def run(model_file, scenario, save_netcdf, save_csv, save_plots,
                     'Model termination condition non-optimal. Not saving plots',
                     fg='red', bold=True
                 )
-        print_end_time(start_time)
+        if fail_when_infeasible and termination != 'optimal':
+            raise BackendError("Problem is infeasible.")
```
timtroendle: Not sure whether this is the right error class.

calliope/cli.py
```diff
@@ -56,6 +57,13 @@
     help='Filename to save profile to if enabled --profile.'
 )
 
+
+_fail_when_infeasible = click.option(
+    '--fail_when_infeasible/--no_fail_when_infeasible', is_flag=True, default=True,
```
timtroendle: I followed Calliope's convention of using underscores in command line options. But is that something you consider changing at one point, maybe for a 1.0 release? ;-).

brynpickering: I think this could be updated a bit earlier than a 1.0 release ;)

calliope/test/test_cli.py
```diff
@@ -120,3 +123,22 @@ def test_generate_scenarios(self):
...
+            "--override_dict={techs.test_supply_elec.constraints.energy_cap_max: 1}"  # elec supply too low
+        ])
+        assert result.exit_code != 0
+
+    def test_success_exit_code_when_infeasible_and_demanded(self):
+        runner = CliRunner()
+        result = runner.invoke(cli.run, [
+            _MINIMAL_TEST_MODEL,
+            "--no_fail_when_infeasible",
+            "--scenario=supply_purchase,investment_costs",  # without these, the model cannot run
```
brynpickering: supply_purchase means that it is running a MILP model... I would go with 'supply_simple' to have the smallest possible model running (if that allows the model to run, ofc)

changelog.rst
```diff
@@ -8,6 +8,8 @@ Release History
 
 |new| Documentation for developers has been improved to include an overview of the internal package structure and a guide to contributing code via a pull request.
 
+|changed| Exit code for infeasible problems now is 1 (no sucess). This is a breaking change when relying on the exit code.
```
brynpickering: success*

calliope/test/test_cli.py
timtroendle: `supply_simple` doesn't exist but is not necessary anyway: `investment_costs` alone has all we need.

PR #184, opened 2019-01-16
Ignore pyomo deprecation warnings during test

timtroendle:

* Ignore DeprecationWarnings of Pyomo during pytest execution.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,3 @@
 [pytest]
-filterwarnings= ignore:.*is deprecated.*:Warning:pyomo
+filterwarnings =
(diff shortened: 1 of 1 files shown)


timtroendle:
Good point. What about making the filter more specific? See commit above, and [pytest doc](
https://docs.pytest.org/en/latest/warnings.html#deprecationwarning-and-pendingdeprecationwarning).

One _could_ add the line number to make it even more specific, but then it will break as soon as the line number changes.

PR #194, opened 2019-02-28
Feature supply share constraint

timtroendle:
Reviewer checklist:

- [x] Documentation updated
- [x] Changelog updated


calliope/backend/pyomo/constraints/group.py
```diff
@@ -79,3 +91,52 @@ def demand_share_constraint_rule(backend_model, constraint_group, carrier, what)
...
+def supply_share_constraint_rule(backend_model, constraint_group, carrier, what):
+    """
+    Enforces shares of carrier_prod for groups of technologies and locations. The
+    share is relative to ``supply`` and ``supply_plus`` technologies only.
+
+    .. container:: scrolling-wrapper
+
+        .. math::
+
+            \\sum_{loc::tech::carrier \\in given\\_group, timestep \\in timesteps} carrier_{prod}(loc::tech::carrier, timestep) <= / >=
```
timtroendle: Can I build that locally? I have no idea of the syntax and whether this is correct.

brynpickering: If you build the docs (`build html` when in `calliope/docs`) then it should be possible to see the result of that in the locally built html files. Needs a bunch of dependencies installed to build the docs (see the relevant requirements file)

timtroendle: cheers, I'll give it a try

brynpickering:
Note that constraints are no longer built over `backend_model.constraint_groups`, but rather `backend_model.group_names_NAME_OF_CONSTRAINT`. This ensures that Pyomo doesn't try to build a constraint over a group that it doesn't fit within.

brynpickering:
Note that constraints are no longer built over `backend_model.constraint_groups`, but rather `backend_model.group_names_NAME_OF_CONSTRAINT`. This ensures that Pyomo doesn't try to build a constraint over a group that it doesn't fit within.

timtroendle:
Does that impact line 32 and 38 in groups.py?

I was actually surprised to see that I did not have to touch the implementation for the tests to pass. I assume that because I can append constraints to the backend where ever I want, but when adding them to `backend_model.group_names_NAME_OF_CONSTRAINT` there is a special mechanism behind?

brynpickering:
Yep, the new implementation doesn't actually invalidate the use of the old method. However, I would expect the old method to still fail in circumstances where there are more complex combinations of group constraints (which is why the implementation was updated).

sjpfenninger:
@timtroendle Once you fix the failing test and resolve conflicts I will merge this

PR #197, opened 2019-02-28
Feature resource area group constraint

timtroendle:
Reviewer checklist:

- [x] Documentation updated
- [x] Changelog updated

### DIFF
--- a/calliope/backend/pyomo/constraints/group.py
+++ b/calliope/backend/pyomo/constraints/group.py
@@ -99,7 +99,13 @@ def demand_share_constraint_rule(backend_model, group_name, carrier, what):
 
 def resource_area_constraint_rule(backend_model, constraint_group, what):
     """
-    TODO write docstring
+    Enforce threshold of resource_area for groups of technologies and locations.
+
+    .. container:: scrolling-wrapper
+
+        .. math::
+
+            \\boldsymbol{resource_{area}}(loc::tech) \\leq threshold
     """
     model_data_dict = backend_model.__calliope_model_data__['data']
     threshold = model_data_dict['group_resource_area_{}'.format(what)][(constraint_group)]
--- a/calliope/backend/pyomo/constraints/group.py
+++ b/calliope/backend/pyomo/constraints/group.py
@@ -389,7 +389,7 @@ def resource_area_constraint_rule(backend_model, constraint_group, what):
     technologies and locations.
 
     .. container:: scrolling-wrapper
-            
+
         .. math::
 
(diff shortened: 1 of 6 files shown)


timtroendle:
Call me crazy, but I'll ask this question: considering we have this new group mechanism and cost caps as well, do we still need `resource_area`? Why not saying `resource_area` is a cost like any other?

Yes, because it may be more convenient in this explicit way. 

No, because really it is the same thing like GHG emissions. Why give resource area a special treatment?

brynpickering:
Do you mean, have 'area' be a cost class then use the same 'cost_max' style notation that we're about to introduce to set a limit on the available area? I suppose it makes sense when you can directly link something like `energy_cap` to `resource_area` and saves using `resource_area_per_energy_cap`. At the moment, `resource_area` is a decision variable in its own right, but I could certainly imagine changing this to always refer to the cost class `area` (which would need some special treatment so that you can only define `area` as a cost class if you want it to do all these crazy things)

timtroendle:
It may well be that I am wrong here as I don't know a lot about the `resource_area` mechanism. I'll try to make my idea explicit, maybe that'll help me or you. That's how you do it now:

```yaml
techs:
    tech_with_area_demand:
        constraints:
            resource_area_per_energy_cap: 10
locations:
    my_loc:
        available_area: 100
        techs:
            tech_with_area_demand:
```

With the new group constraints, I thought this could be changed to this:

```yaml
techs:
    tech_with_area_demand:
        costs:
            area:
                 energy_cap: 10
locations:
    my_loc:
        techs:
            tech_with_area_demand:
group_constraints:
    area_in_my_loc:
        cost_max:
            area: 100
```

Obviously more verbose, but (at least I see) no need for a special treatment of `area`. So all `area` related things could be removed from Calliope.

From what I understood from what you said, this is only one use case of several of the `area` variable. So maybe this isn't a full replacement. Again, all this is just a thought.

brynpickering:
Yep, that's exactly what I imagined when you described it. However, `resource_area` is used as a possible multiplier for `resource` (i.e. `available_resource = resource_area * resource`) if the `resource_unit` is given as `per_area`. Can you see an easy way of incorporating that?

timtroendle:
Oh, I got it know. Well, the only way I see is to generalise this as well and allow `resource_unit` to be of any cost class. This isn't totally unrealistic for other costs: you could infer power generation from industrial CO₂ emissions for example. But sounds like that goes a little too far.

brynpickering:
Note that constraints are no longer built over `backend_model.constraint_groups`, but rather `backend_model.group_names_NAME_OF_CONSTRAINT`. This ensures that Pyomo doesn't try to build a constraint over a group that it doesn't fit within.

Issue #199, opened 2019-03-04
Group constraints cannot be switched off


timtroendle:
## Problem description

Many other Calliope objects like locs, techs etc can be switched off (in scenarios) using ´exists: False´. This is currently not possible with group constraints. See #198 for a use case.

## Calliope version

0.6.4-dev

PR #200, opened 2019-03-04
Add option to switch off group constraints

timtroendle:
Implements https://github.com/calliope-project/calliope/issues/199.

* allow group constraints to be switched off in scenarios

Reviewer checklist:

- [x] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/calliope/core/preprocess/model_data.py
+++ b/calliope/core/preprocess/model_data.py
@@ -433,7 +433,12 @@ def group_constraints_to_dataset(model_run):
 
     group_constraints = model_run['group_constraints']
 
-    for constr_name in model_run.sets['group_constraints']:
+    group_constraints = {
+        name: data for name, data in group_constraints.items()
+        if data.get("exists", True) is True
+    }
+
+    for constr_name in group_constraints:
         dims = ['group_names_' + constr_name]
         if constr_name in checks.defaults.allowed_group_constraints.per_carrier:
             dims.append('carriers')
--- a/calliope/core/preprocess/model_data.py
+++ b/calliope/core/preprocess/model_data.py
@@ -433,12 +433,7 @@ def group_constraints_to_dataset(model_run):
 
     group_constraints = model_run['group_constraints']
 
-    group_constraints = {
-        name: data for name, data in group_constraints.items()
-        if data.get("exists", True) is True
-    }
-
-    for constr_name in group_constraints:
+    for constr_name in model_run.sets['group_constraints']:
         dims = ['group_names_' + constr_name]
         if constr_name in checks.defaults.allowed_group_constraints.per_carrier:
             dims.append('carriers')
--- a/calliope/test/common/test_model/group_constraints.yaml
+++ b/calliope/test/common/test_model/group_constraints.yaml
@@ -1,6 +1,7 @@
 model:
-        name: Test model for group constraint mechanism
-        timeseries_data_path: 'timeseries_data'
+    name: Test model for group constraint mechanism
+    timeseries_data_path: 'timeseries_data'
+    subset_time: '2005-01-01'
 
 run:
     mode: plan
--- a/calliope/test/common/test_model/group_constraints.yaml
+++ b/calliope/test/common/test_model/group_constraints.yaml
@@ -1,0 +1,55 @@
+model:
+        name: Test model for group constraint mechanism
+        timeseries_data_path: 'timeseries_data'
+
+run:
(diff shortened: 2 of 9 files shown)


calliope/core/preprocess/model_data.py
```diff
@@ -434,9 +434,16 @@ def group_constraints_to_dataset(model_run):
     # Get a DataFrame with constraint names as index, constraint groups as columns
     constraint_df = (
         pd.DataFrame(model_run['group_constraints'])
-          .drop(['techs', 'locs'], axis=0, errors='ignore')
+          .drop(['techs', 'locs'], axis='index', errors='ignore')
     )
 
+    if 'exists' in constraint_df.index:
+        constraint_df.loc["exists"] = constraint_df.loc["exists"].fillna(True)
```
timtroendle: This line creates a Pandas warning:

```
KeyError: '_typ'
Exception ignored in: 'util.is_period_object'
KeyError: ('_typ',)
KeyError: '_typ'
Exception ignored in: 'pandas._libs.lib.is_interval'
KeyError: ('_typ',)
```

When printing `constraint_df` on the command line (e.g. in `pdb`) a similar warning is issued. The reason seems to be the dict within the dataframe. Is it a good idea to store a dict within a Pandas dataframe? Seems brittle.

brynpickering: Agreed, is there any reason this is a df at all? Could it not be kept as a dict?

calliope/test/common/test_model/group_constraints.yaml
```diff
@@ -0,0 +1,55 @@
+model:
```
brynpickering: This seems like an unnecessary file addition... Why not just set 'exists' to False as an override dict in the specific test, rather than have a whole new YAML file?

calliope/test/test_constraint_results.py
```diff
@@ -150,8 +150,38 @@ def override(feasibility, cap_val):
...
+        expensive_generation = (model.get_formatted_array("carrier_prod")
+                                     .to_dataframe()
+                                     .reset_index()
+                                     .groupby("techs")
+                                     .carrier_prod
+                                     .sum()
+                                     .loc["expensive_supply"])
+        assert expensive_generation == 0
+
+    def test_switched_off_group_constraint(self):
```
brynpickering: If using an override dict instead of a new model definition:
```
model = build_model(
           override_dict={'group_constraints.example_demand_share_max_constraint.exists': False},
            model_file="demand_share.yaml",
            scenario="demand_share_max_systemwide"
        )
```

calliope/test/common/test_model/group_constraints.yaml
timtroendle: Agreed, if it was only for this one override, the file wouldn't be necessary.

But I also added that file in #194 with more overrides. The idea being, that there is one test file + one test class for the general mechanism of group constraints: application to several locations, application to several technologies, switchability, etc. Other files and test classes would then test specifically the constraints, not the group constraint mechanism. Does that make sense, or would you rather mix the general tests into the other files?

brynpickering: Aha, OK. It's a bit messy at the moment, since the testing of location/technology subsets happens in the individual constraint model overrides anyway (which is valid, to check that the constraint is working correctly). Let's pass all these through, then once we have all the group constraints merged together, we can look at consolidating if it looks a bit much.

calliope/test/common/test_model/group_constraints.yaml
```diff
@@ -0,0 +1,55 @@
...
+        name: Test model for group constraint mechanism
+        timeseries_data_path: 'timeseries_data'
```
brynpickering: For this and all other models, probably worth just having it run over one day, to minimise optimisation time, unless there's some additional information that comes from running over a longer time period...

timtroendle: Gave it a try just now and that kills 50% of test time. So that's certainly a good idea.

sjpfenninger:
@timtroendle Can you add a bit of text on what this does to the documentation?

timtroendle:
@sjpfenninger Sure, have a look at the commit above.

calliope/test/common/test_model/group_constraints.yaml
sjpfenninger: 👍

Issue #187, opened 2019-01-23
Linear scalarisation of multi-objective optimisation problems (e.g. carbon price)


timtroendle:
## Problem description

This is to track the on-going discussion on this feature request: it would be nice to be able to solve multi objective optimisation problems with a simple linear scalarisation of the objectives.

Amongst others, this would allow to add a carbon price to the model: one could define two cost classes, say `greenhouse_gases` and `monetary` and provide weights for each, say `5€/ton` and `1`. But it would also allow to linearly scalarise any other (real!) multi objective cost function.

## Calliope version

0.6.3

timtroendle:
A workaround that works in `0.6.3` is to use only one cost class and add all other costs manually to this cost class. As this is manual, it is error prone though.

timtroendle:
Simply to capture this somewhere: Gurobi has a feature for that:

http://www.gurobi.com/documentation/8.1/refman/multiple_objectives.html#sec:MultipleObjectives

PR #206, opened 2019-03-14
Weighted objective function

brynpickering:
Fixes issue(s) #187

* `run.objective_options.cost_class` can now be defined as a dictionary of cost_class:weight key:value pairs
* objective will consider all cost classes in `run.objective_options.cost_class` and will weight them according to the weights given. If still given as a string or given as a set, weight(s) will be considered as 1

Reviewer checklist:

- [x] Documentation updated
- [x] Changelog updated


calliope/config/model.yaml
```diff
@@ -14,7 +14,7 @@ run:
     cyclic_storage: true # If true, storage in the last timestep of the timeseries is considered to be the 'previous timestep' in the first timestep of the timeseries
     ensure_feasibility: false # If true, unmet_demand will be a decision variable, to account for an ability to meet demand with the available supply. If False and a mismatch occurs, the optimisation will fail due to infeasibility
     mode: plan  # Which mode to run the model in: 'plan' or 'operation'
-    objective_options: {'cost_class': 'monetary', 'sense': 'minimize'}  # Arguments to pass to objective function. If cost-based objective function in use, should include 'cost_class' and 'sense' (maximize/minimize)
+    objective_options: {'cost_class': {'monetary'}, 'sense': 'minimize'}  # Arguments to pass to objective function. If cost-based objective function in use, should include 'cost_class' and 'sense' (maximize/minimize)
```
sjpfenninger: I think this should be `{'cost_class': {'monetary': 1}, 'sense': 'minimize'}`

brynpickering: I'm not too fussed, they all end up as the same thing in the end.

brynpickering:
Will add tests in next few hours

brynpickering:
So, there are a few issues outstanding with this feature:

1. `calliope.exceptions.ModelWarning` is expected, based on a check for cost classes defined as sets, which are then converted to dicts, with the key as the cost class and the value equal to one. Current tests fail because this isn't being caught, or the warning being raised isn't as expected.

2. The way `AttrDict.union()` combines two dictionaries doesn't work with this feature. Either:
a. the default is `objective_options.cost_class: {'monetary' : 1}` which is not overridden on the user saying, e.g. `objective_options.cost_class: {'emissions' : 1}`, instead you get `objective_options.cost_class: {'monetary' : 1, 'emissions' : 1}`. One would have to explicitly include `'monetary': 0` to remove it.
b. the default is `objective_options.cost_class: 'monetary'` which, on overriding with e.g. `objective_options.cost_class: {'emissions' : 1}`, completely removes other keys in the `objective_options` dict, creating `objective_options: {'cost_class': {'emissions' : 1}}`. I.e. it removes `sense`. This causes the model to fail, since it expects `sense` to be there.

brynpickering:
Note @sjpfenninger and @timtroendle : I've updated all `DeprecationWarning` to `FutureWarning`. The former is ignored by default in Python interpreters, so users would not get these messages. We should keep `DeprecationWarning` for developer facing functions which are deprecated. `FutureWarning` could instead be `calliope.exceptions.ModelWarning`, if you prefer, but then it might be difficult to extract those warnings for ignoring in pytest...

Issue #207, opened 2019-03-14
Fail or warn with unknown tech parameters


timtroendle:
## Problem description

Calliope isn't fool-proof enough for a fool like me. When you forget to properly indent your YAML file, things can go wrong without notice. Consider this tech with high costs:

```yaml
my_tech:
        essentials:
            name: 'My tech'
            parent: storage
            carrier: electricity
        constraints:
            energy_cap_max: inf
            storage_cap_max: inf
        costs:
        monetary: # oh no!
            energy_cap: 112345
```

But because the indentation is wrong, this tech is actually cost free.

Currently, this goes unnoticed. The user is not warned, the execution does not stop, and there are no costs considered. 

## Calliope version

0.6.4.dev

PR #208, opened 2019-03-14
Fix unknown tech parameters

timtroendle:
Fixes issue(s) #207.

* Unknown top level tech parameters stop execution with an error message.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


sjpfenninger:
Thanks for this! We generally raise warnings and continue execution, rather than errors, when unknown items are encountered — so for consistency I think this should behave the same way (but one could argue whether errors should be raised in all instances, rather than just warnings).

Issue #169, opened 2018-11-30
Binary flow direction for storage and transmission technologies


brynpickering:
## Problem description

It is possible for a storage or transmission tech to have non-zero `carrier_con` and `carrier_prod`, allowing nonphysical use of these technologies to artificially dump an energy carrier (if their efficiency is <100%). To avoid this, binary variables and connected constraints need to be added to limit the technology to only flow in one direction in a timestep.

# Steps to reproduce the problem

Model any system with storage/transmission where it is financially beneficial to over-produce a particular energy carrier.

## Calliope version
0.6.3

PR #211, opened 2019-03-18
Binary storage/transmission operation

brynpickering:
Fixes issue(s) #169

* Additional binary decision variable: `operating_switch`
* Additional constraints to limit one of `carrier_con` or `carrier_prod` to operate in a given timestep, using `operating_switch`, for either storage or transmission technologies. 

Reviewer checklist:

- [x] Documentation updated
- [x] Changelog updated


sjpfenninger:
@brynpickering changelog, docs?

calliope/test/test_example_models.py
```diff
@@ -354,7 +354,7 @@ def test_urban_example_results_cap_gurobi(self):
 
     def test_milp_example_results(self):
         model = calliope.examples.milp(
-            override_dict={'model.subset_time': '2005-01-01'}
+            override_dict={'model.subset_time': '2005-01-01', 'run.solver_options.mipgap': 0.001}
```
sjpfenninger: why overriding mipgap here?

calliope/test/test_analysis.py
```diff
@@ -48,9 +48,30 @@ def test_national_scale_plotting(self, national_scale_example):
                 plotly_kwarg_updates={'auto_open': False, 'filename': out_path}
             )
 
+    @python36_or_higher
```
sjpfenninger: not clear what this test is adding

calliope/example_models/urban_scale/scenarios.yaml
```diff
@@ -24,6 +25,9 @@ overrides:
                         energy_cap: 35
                         purchase: 2000
             # boiler-end
+            heat_pipes:
```
sjpfenninger: why is this added here? it seems like in the tests you are also and separately adding in binary_operation: True with direct overrides

calliope/core/io.py
```diff
@@ -82,7 +82,7 @@ def save_csv(model_data, path, dropna=True):
     os.makedirs(path, exist_ok=False)
 
     if ('termination_condition' not in model_data.attrs or
-            model_data.attrs['termination_condition'] == 'optimal'):
+            model_data.attrs['termination_condition'] in ['optimal', 'feasible']):
```
sjpfenninger: again, what makes this necessary?

calliope/config/defaults.yaml
```diff
@@ -44,6 +44,7 @@ default_tech:
         primary_carrier_in: false  # Selects the primary input carrier to associate with costs & constraints (if multiple primary input carriers are assigned)
         primary_carrier_out: false  # Selects the primary output carrier to associate with costs & constraints (if multiple primary output carriers are assigned)
     constraints:
+        binary_operation: false  # if True, carrier_prod and carrier_con cannot both occur in the same timestep
```
sjpfenninger: I wonder if binary operation is the right name to use, since it is quite generic. Is this future-proofing as it might encompass additional features? Or should it be called something like `no_concurrent_prod_con` which is more descriptive? Finally: should there be clearer documentation on which constraints turn on MILP and thus dramatically increase solution time -- possibly even ModelWarning that says "The following constraints you activated are forcing mixed-integer and may affect solution time: ...."?

calliope/backend/pyomo/model.py
```diff
@@ -118,9 +119,6 @@ def generate_model(model_data):
     if hasattr(backend_model, 'loc_techs_conversion_plus'):
```
sjpfenninger: Hm, given this change: would it make more sense to list all `load_constraints` functions by default, if we are checking within the respective `load_constraints` functions for existence of sets anyway, for ease and consistency? Even automagically load all existing constraint submodules and their load_constraints functions so they don't need explicit listing?

calliope/analysis/plotting/plotting.py
```diff
@@ -168,7 +168,7 @@ def __init__(self, model):
     def check_optimality(self):
         termination = self._model._model_data.attrs.get(
             'termination_condition', 'did_not_yet_run')
-        if termination not in ['optimal', 'did_not_yet_run']:
+        if termination not in ['optimal', 'did_not_yet_run', 'feasible']:
```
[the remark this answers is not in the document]
brynpickering: Because a run which stops due to the `MIPGap` condition returns as feasible, not optimal

calliope/test/test_example_models.py
brynpickering: because it takes ages to converge on the LP relaxation, due to the introduction of the binary operation switch

calliope/test/test_analysis.py
brynpickering: It has bi-directional flow, so there's some functionality in `flows` which would otherwise not be triggered (neither national scale nor milp models have bi-directional flow, the latter due to the introduction of the binary operation switch). I think plotting tests could be cleaned up generally, but this was a quick (and dirty?) way of keeping coverage up ;)

calliope/analysis/plotting/plotting.py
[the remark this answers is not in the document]
sjpfenninger: Can you just add a comment saying just that on the line above, for future reference?

calliope/example_models/urban_scale/scenarios.yaml
sjpfenninger: Ok, we'll keep in mind "cleaning up tests" for some undetermined future date!

calliope/config/defaults.yaml
sjpfenninger: Re 2: Sounds sensible to use the subsets, and that could be in `preprocess.checks.check_final` which happens after all sets are built?

brynpickering:
@sjpfenninger I've not addressed everything here, but have changed the parameter name and added in a bunch of tests.

brynpickering:
@sjpfenninger  Added changelog and documentation entries. Also changed decision variable name to `prod_con_switch`.

calliope/backend/pyomo/model.py
brynpickering: See #221

sjpfenninger:
👍

Added the docs to advanced constraints too, otherwise good

Issue #212, opened 2019-03-20
Storages with fixed energy to power ratio


timtroendle:
## Problem description

Currently, storages _always_ have two independent decision variables: `energy_cap` and `storage_cap`. The currently existing `charge_rate` constraint can cap the `energy_cap` based on `storage_cap`, but as long as the `charge_rate` is not 0, the two variables remain largely independent.

There are situations where one would want to tightly link the two variables:

* technological reasons: one may want to assume that batteries can only have a certain energy/power ratio;
* economic reasons: one may know or assume costs per installed power ( or per installed energy ) but only based on a certain energy/power ratio of the storage technology ([example](https://doi.org/10.1038%2Fnenergy.2017.110)).

This is currently impossible.

## Calliope version

0.6.4.dev

brynpickering:
I would argue that the current implementation of charge rate (or C-rate, using terminology from your example) is correct. In your examples, although there are given energy/power ratios for storage technologies, these give the maximum, do they not? Any energy technology that can have a discharge rate of 50% of the maximum stored volume should also be able to discharge at a 25% rate. For batteries, C-rate is calculated based on maximum current draw, but clearly it is possible for the current draw to fluctuate based on the load, provided it doesn't go above the maximum.

Still, there are probably cases where you don't want this to apply, even if it is somewhat unphysical (similar to my use of not allowing PV curtailment at the urban scale ;) ).

timtroendle:
I'm not saying the current implementation of `charge_rate` is wrong. But it cannot do what I am requesting above. I would agree with what you are saying about charge rates, but remember this here is not about currents and powers, but this is about _installed_ power (energy cap).

So as an example for batteries: if you are installing fictional battery type Li-Ion X2000 (which as we all know has a max charge rate of 1) you may want to constrain the charge rate for whatever reason to 0.5, and you can do so using the current `charge_rate` parameter. But don't forget, the battery actually has a max charge rate of 1, so while you are using only 0.5 kW of the installed 1 kWh, you still have installed 1 kW. That cannot be modelled with `charge_rate`.

[...]

PR #213, opened 2019-03-20
Add energy_cap_per_storage_cap constraints

timtroendle:
Fixes https://github.com/calliope-project/calliope/issues/212.

* I added another `storage_cap_xxx` constraint, `storage_cap_ratio`, which ties storage capacity to energy capacity.

Reviewer checklist:

- [x] Documentation updated
- [x] Changelog updated


calliope/config/defaults.yaml
```diff
@@ -80,6 +80,7 @@ default_tech:
         storage_cap_equals: false  # name: Specific storage capacity ¦ unit: kWh ¦ If not defined, ``energy_cap_equals`` * ``charge_rate`` will be used as the capacity and overrides ``_max`` and ``_min`` constraints.
         storage_cap_max: .inf  # name: Maximum storage capacity ¦ unit: kWh ¦ If not defined, ``energy_cap_max`` * ``charge_rate`` will be used as the capacity.
         storage_cap_min: 0  # name: Minimum storage capacity ¦ unit: kWh ¦
+        storage_cap_ratio: false # name: Ratio of storage capacity relative to energy capacity ¦ unit: fraction ¦
```
brynpickering: I'm not sure this is sufficiently descriptive. Perhaps it should be `energy_cap_per_storage_cap`, or 
`charge_rate` should be changed to include a `min`, `equals`, and `max`?

calliope/backend/pyomo/constraints/capacity.py
```diff
@@ -130,8 +132,13 @@ def storage_capacity_constraint_rule(backend_model, loc_tech):
             \\quad \\forall loc::tech \\in loc::techs_{store}
 
     """
-
-    return get_capacity_constraint(backend_model, 'storage_cap', loc_tech)
+    ratio = get_param(backend_model, 'storage_cap_ratio', loc_tech)
+    if po.value(ratio) is not False and po.value(ratio) is not None:
+        return backend_model.storage_cap[loc_tech] == (
+            backend_model.energy_cap[loc_tech] * ratio
```
brynpickering: All the literature defines it the other way, i.e. power to energy ratio, not energy to power ratio. No matter the name given to the parameter, it should probably follow that convention

calliope/test/common/test_model/storage_capacity_ratio.yaml
```diff
@@ -0,0 +1,41 @@
+model:
+    name: Test model for storage capacity ratio
+    timeseries_data_path: 'timeseries_data'
+    subset_time: '2005-01-01'
+
+run:
+    mode: plan
+    solver: glpk
```
brynpickering: cbc now please!

calliope/test/common/test_model/storage_capacity_ratio.yaml
```diff
@@ -0,0 +1,41 @@
...
+    cyclic_storage: False
+
+techs:
```
brynpickering: hmm, how does a model with only storage and demand work? There's no way for the system to get energy to begin with. Maybe add a supply tech?

calliope/test/test_constraint_results.py
```diff
@@ -799,3 +799,46 @@ def test_location_specific_energy_cap_min_constraint(self):
         expensive_capacity1 = capacity.loc[("1", "expensive_supply")]
         assert round(expensive_capacity0, 5) >= 6
         assert expensive_capacity1 == 0
+
+
+class TestStorageCapacityRatio:
+
+    def test_no_ratio_set(self):
```
brynpickering: I would recommend testing these without actually running the optimisation. See the tests in `test_backend_pyomo` which check if a particular variable exists in a constraint (e.g. line 214). In this case, you could check that the storage capacity constraints include the variable `energy_cap`, which they only will if your constraint has been applied

calliope/backend/pyomo/constraints/capacity.py
```diff
@@ -130,8 +132,13 @@ def storage_capacity_constraint_rule(backend_model, loc_tech):
             \\quad \\forall loc::tech \\in loc::techs_{store}
 
     """
-
-    return get_capacity_constraint(backend_model, 'storage_cap', loc_tech)
+    ratio = get_param(backend_model, 'storage_cap_ratio', loc_tech)
+    if po.value(ratio) is not False and po.value(ratio) is not None:
```
brynpickering: This might be better placed in its own constraint. At the moment, if a user sets:

```
storage_cap_max: 10
energy_cap_max: 4
storage_cap_ratio: 2
```

`energy_cap_max` will be applied, so `storage_cap` cannot be higher than 8. But, if you set:

```
storage_cap_max: 8
energy_cap_max: 5
storage_cap_ratio: 2
```

`storage_cap_max` is ignored, allowing `storage_cap` to actually go to 10.

Instead, if both the `energy_cap` and `storage_cap` constraints can do their own thing, the constraint in this PR could be an additional constraint which is applied externally, similar to `charge_rate`. To be honest, i'd see it making most sense that `charge_rate` and this constraint are merged into one constraint, in which the link between the two decision variables is either loose (setting a maximum) or strict (as in this constraint)

calliope/test/common/test_model/storage_capacity_ratio.yaml
timtroendle: Right!

calliope/test/common/test_model/storage_capacity_ratio.yaml
timtroendle: Mhm the whole point of a minimal example is to be minimal. The demand can be fed from a storage, as long as `cyclic_storage` is turned off, which it is. The tests prove it works.

brynpickering: Quite. Although if we ever want these to become part of an examples library online, it does seem a little odd. I'm happy to have it remain as is, but maybe with a comment in the YAML next to the cyclic storage that turning it off is what makes the model feasible.

changelog.rst
```diff
@@ -16,6 +16,10 @@ Release History
 
 |new| Documentation for developers has been improved to include an overview of the internal package structure and a guide to contributing code via a pull request.
 
+|new| Storage capacity can be tied to energy capacity with a new `energy_cap_per_storage_cap_equals` constraint.
+
+|changed| `change_rate` has been renamed to `energy_cap_per_storage_cap_max`. `change_rate` will be removed in Calliope 0.7.0.
```
timtroendle: typo

calliope/core/preprocess/checks.py
```diff
@@ -334,10 +333,11 @@ def _check_tech(model_run, tech_id, tech_config, loc_id, model_warnings, errors,
     # If the technology is supply_plus, check if it has storage_cap_max. If yes, it needs charge rate
     if model_run.techs[tech_id].essentials.parent == 'supply_plus':
         if (any(['storage_cap_' in k for k in tech_config.constraints.keys()])
-            and 'charge_rate' not in tech_config.constraints.keys()):
+            and 'charge_rate' not in tech_config.constraints.keys()
```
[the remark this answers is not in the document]
timtroendle: Any agreement on how to handle this @brynpickering, @sjpfenninger  ? As far as I understand, this is considered as an error only since recently, but Calliope's code base is full of this -- now assumed -- error.

timtroendle:
Ready for review @brynpickering.

calliope/core/preprocess/checks.py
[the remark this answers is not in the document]
sjpfenninger: Unfortunately our codebase is not consistent, but generally we'd like to see line break before binary operator rather than ine break after binary operator. I've added W503 to the ignore list in the flake8 config on master - should have done that a while ago (hound will only re-run if you push another commit to this branch though).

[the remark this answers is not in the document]
timtroendle: Thanks! Will merge and update after Bryn's review.

calliope/core/model.py
```diff
@@ -292,3 +293,41 @@ def info(self):
             times=len(self._model_data.coords.get('timesteps', [])))
         info_strings.append('Model size:   {}'.format(msize))
         return '\n'.join(info_strings)
+
+    def _check_future_deprecation_warnings(self):
```
brynpickering: Any reason for moving this into `model.py`?

calliope/core/preprocess/checks.py
```diff
@@ -334,10 +333,11 @@ def _check_tech(model_run, tech_id, tech_config, loc_id, model_warnings, errors,
...
+            and 'energy_cap_per_storage_cap_max' not in tech_config.constraints.keys()):
             errors.append(
                 '`{}` at `{}` fails to define '
-                'charge_rate, but is using storage'.format(tech_id, loc_id, required)
+                'energy_cap_per_storage_cap_max, but is using storage'.format(tech_id, loc_id, required)
```
brynpickering: I think this should also check for `energy_cap_per_storage_cap_equals`, max or equals is fine, provided one has been defined.

calliope/core/preprocess/model_data.py
```diff
@@ -127,7 +127,7 @@ def _get_set(constraint):
             return 'loc_techs_supply_plus'
         elif 'resource' in constraint: # i.e. everything with 'resource' in the name that isn't resource_cap
             return 'loc_techs_finite_resource'
-        elif 'storage' in constraint or 'charge_rate' in constraint:
+        elif 'storage' in constraint or 'charge_rate' in constraint or 'energy_cap_per_storage_cap_max' in constraint:
```
brynpickering: man, these are such hacks... seeing as 'storage' is in 'energy_cap_per_storage_cap_max' (and '_equals'), this will actually catch those with the first condition (`'storage' in constraint`).

calliope/test/test_constraint_results.py
```diff
@@ -799,3 +799,50 @@ def test_location_specific_energy_cap_min_constraint(self):
...
+class TestEnergyCapacityPerStorageCapacity:
+
+    @pytest.fixture
+    def model_file(self):
+        return "energy_cap_per_storage_cap.yaml"
+
+    def test_no_constraint_set(self, model_file):
+        model = build_model(model_file=model_file)
+        model.run()
+        assert model.results.termination_condition == "optimal"
```
brynpickering: I'm generally more of a fan of xarray syntax, since it is cleaner in these instances. E.g. here it would be: `model.get_formatted_array("energy_cap").loc[{'techs': 'my_storage'}].sum()`

calliope/test/test_core_model.py
```diff
@@ -34,3 +34,9 @@ def test_info_minimal_model(self):
         model = calliope.Model(model_location)
 
         model.info()
+
```
brynpickering: There are future warning tests in `test_core_preprocess` in a class called (now misnamed, perhaps) `TestDeprecationWarnings`. Might be best to keep them all in one place, seeing as much of their addition is a copy-paste job. But then maybe we should just have a new file for Future/Deprecation warning capture?

calliope/backend/pyomo/constraints/capacity.py
```diff
@@ -155,6 +163,36 @@ def energy_capacity_storage_constraint_rule(backend_model, loc_tech):
     )
 
 
+def energy_capacity_storage_constraint_rule(backend_model, loc_tech):
```
brynpickering: Have you tried sending this off to `get_capacity_constraint`? I.e. `get_capacity_constraint(backend_model, 'energy_cap', loc_tech, _equals=energy_cap_per_storage_cap_equals, _max=energy_cap_per_storage_cap_max, _min=0 scale=None)`. I don't actually know if pyomo will complain that there is a decision variable in the upper bound. If it does, then I'm fine with this approach, but it misses a check in `get_capacity_constraint` as to whether the equals is infinite.

But now I think about this, the check would be much easier to put into preprocessing anyway (any variable with `_equals` in the name cannot have inf in its values)...

calliope/core/model.py
timtroendle: Yeah I believe in terms of clear modularisation it belongs into the Model module: after all, the Model class is the only instance calling the function _and_ it's handing over its internals as a parameter. That's why I concluded, (a) it should be in the same module as the only calling instance, and (b) it should really be a method and not a function.

calliope/core/preprocess/checks.py
timtroendle: Right.

calliope/core/preprocess/model_data.py
timtroendle: True... any ideas on how to improve this?

calliope/test/test_constraint_results.py
timtroendle: It looks good and is more concise. But your example returns an 0D Array rather than a scalar value. What is a good way to access the scalar value? The only way I found (after stackoverflowing) is `model.get_formatted_array("energy_cap").loc[{'techs': 'my_storage'}].sum().values[()]` which is not very intuitive. Do you know a better way?

calliope/core/preprocess/model_data.py
brynpickering: For the time being, no... Stefan and I left it as is when putting 0.6 together because we couldn't think of a non-hardcoded way of doing it

calliope/test/test_constraint_results.py
brynpickering: `.item()`?

calliope/test/test_core_model.py
timtroendle: Yeah agreed, that is not ideal. I usually like to have one test module per production module which makes it simple to find the corresponding tests. So if there is a `core/model.py`, there is as well a `tests/core/test_model.py` that contains all tests. That's why I intuitively moved it there.

I like the idea of moving everything into one test module `test_core_future_warnings`.

calliope/test/test_constraint_results.py
timtroendle: Nice!

calliope/backend/pyomo/constraints/capacity.py
timtroendle: I did try based on your recommendation, but this line did not work: `decision_variable = getattr(backend_model, parameter)`.

calliope/test/test_core_future_warnings.py
```diff
@@ -0,0 +1,27 @@
...
+
+@pytest.mark.filterwarnings('always::FutureWarning')
+class TestFutureWarnings:
+
+    def test_group_share_warning(self):
+        override = {
+            'model.group_share.test_supply_elec.carrier_prod_min.electricity': 0.85
+        }
+        expected_warning_msg = '`group_share` constraints will be removed in v0.7.0'
+        with pytest.warns(FutureWarning, match=expected_warning_msg):
```
timtroendle: I removed the version check from this test and the following. I believe it's fine like that: should the warning (and the feature) be removed, this test will fail. We should then remove the test, because we are not expecting anything really. But maybe I am missing something here.

calliope/core/model.py
sjpfenninger: Happy with this change

calliope/test/test_core_future_warnings.py
sjpfenninger: Sounds sensible to me

calliope/test/test_core_model.py
sjpfenninger: Moving to a `test_core_future_warnings` seems good -- and we do need to better organise tests. Corresponding test module for each production module would sound sensible for when we do that.

sjpfenninger:
I think it's ok to ignore the coverage decrease here -- the deprecated functionality is no longer being tested, but we can live with that I think.

calliope/backend/pyomo/constraints/capacity.py
brynpickering: ok, don't worry about it then!

PR #203, opened 2019-03-11
Add group cost constraints

brynpickering:
Fixes issue(s) #188

Summary of changes in this pull request:

* `min`/`max`/`equals` constraints which can be applied to `cost`, `cost_investment` and `cost_var` decision variables on a group level (i.e. any subset of `locs` and `techs`)

Reviewer checklist:

- [x] Test(s) added to cover contribution
- [ ] Documentation updated
- [x] Changelog updated
- [x] Coverage maintained or improved

Issue #193, opened 2019-02-28
Fully implement, document and test group constraints


sjpfenninger:
## Problem description

Things that need to be added that didn't make it into PR #191:

Constraints:

* [x] cost constraints
~* [ ] reserve margin~

Tests and docs:

* [x] test for a list with length greater than one in either 'locs' or 'techs' -- maybe have a model with three locations and three techs, so that the inclusion of two of those three within a constraint can be checked
	* To ensure the run times don't get excessive, it is probably sufficient to build the model with pyomo, but not run it. The constraint can still be checked to exist (`constraint_exists` function) and the upper/lower bound can be checked (e.g. `test_backend_pyomo.py` line 488).
* [x] demand_share_constraint_rule is missing docstring
* [x] 2 tests for these two claims: "When specifying group constraints, a named group can list an arbitrary amount of constraints, and optionally give a subset of techs and locations"
~* [ ] clean up the organisation of tests and test models (includes file naming consistency across e.g. "supply_share.yaml" and "model_cost_cap.yaml")~ →  moved to #240
* [x] group cost constraints need docs (#203)
* [x] Group constraints fail with inexistent location #198

Issue #163, opened 2018-10-26
Better approach to and docs on debugging/troubleshooting


sjpfenninger:
## Problem description

Docs already have a section on [debugging failing runs](https://calliope.readthedocs.io/en/stable/user/ref_advanced_func.html#debugging-failing-runs) and [solver options](https://calliope.readthedocs.io/en/stable/user/ref_advanced_func.html#solver-options), but they are hidden and incomplete as practical guides to dealing with common issues.

In particular, a coherent documentation of strategies to deal with infeasinble/unbounded models would be useful.

## Topics to add to docs

* Trying `run.solver_options.Presolve: 0` as it seems that the large numeric range can cause infeasibility with Gurobi
* With Gurobi, set `run.solver_options.DualReductions: 0` to see whether a model is infeasible or unbounded
* Set --save_logs=outdir and `solver_io: lp` to get an LP file, then use Gurobi to generate an Irreducible Inconsistent Subsystem which then can be analysed: ``gurobi_cl ResultFile=model.ilp tmp_model.pyomo.lp`` (see http://www.gurobi.com/documentation/7.5/refman/solving_a_model.html and http://www.gurobi.com/documentation/7.5/refman/model_file_formats.html)
    * This could also involve a new CLI option such as ``--save_lp`` to save an LP file irrespective of what `solver_io` was set to?

Issue #162, opened 2018-10-26
Make time series documentation more prominent


sjpfenninger:
## Problem description

@timtroendle in #156:

> [...] considering that one has to have a time series, the time series documentation is rather hard to find: it's below more info and advanced technology. One could argue to move that into the prominent and top section building a model.

PR #214, opened 2019-03-25
Documentation cleanup and improvements

sjpfenninger:
# Summary of changes in this pull request

Documentation cleanup and improvements

* Restructure overall table of contents
* Split existing advanced content into "Advanced constraints" and "Advanced configuration"
* Clean up group constraint documentation and change national-scale example model to use group constraints rather than the now deprecated group_share constraints
* Address #151, #159, #162, #163, #193 (partial), #203

## Bugs fixed

* Fix a bug in group constraint implementation that led to Calliope crashing if a group constraint was defined over all regions (by not setting locs:) but some of techs considered by the consstraint did not exist across all locations
* Fix a bug in group constraint implementation that meant only supply, not supply_plus technologies were considered

The existence of these bugs demonstrates that group constraints need better testing (#193).

## Resolves the following issues

* #151 -- Document use for random seed
* #159 -- Document ``resource_area`` constraints in the new "Advanced constraints" section
* #162 -- Update and move time series documentation into the "building a model" section
* #163 -- Add new troubleshooting section and add ``--save_lp`` to the ``calliope run`` command to make troubleshooting easier
* #193 (partial) -- Add docstring to demand_share_constraint_rule
* #203 / #193 (partial) -- Add documentation for group cost constraints

# Reviewer checklist

- [x] Documentation updated
- [x] Changelog updated


calliope/cli.py
```diff
@@ -172,23 +193,63 @@ def new(path, template, debug):
...
 @click.option('--scenario')
+@click.option('--model_format')
+@click.option('--override_dict')
 @click.option('--save_netcdf')
 @click.option('--save_csv')
 @click.option('--save_plots')
 @click.option('--save_logs')
-@click.option('--model_format')
-@click.option('--override_dict')
+@click.option(
```
brynpickering: Any reason to not put in the help strings for the other arguments?

calliope/example_models/national_scale/scenarios.yaml
```diff
@@ -2,8 +2,8 @@
 # Scenarios are optional, named combinations of overrides
```
brynpickering: Do we need these scenarios any more, given that we have our minimal examples?

calliope/test/test_constraint_results.py
```diff
@@ -8,7 +8,7 @@
 
 
 class TestNationalScaleExampleModelSenseChecks:
```
brynpickering: Again, these are essentially duplicating tests now given by the minimal group constraint models, do we need them?

calliope/cli.py
```diff
@@ -172,23 +193,63 @@ def new(path, template, debug):
...
-@cli.command(short_help='Run a model.')
+def _run_setup_model(
+        model_file, scenario, model_format, override_dict):
+    """
+    Build model in CLI commands. Returns ``model``, a ready-to-run
+    calliope.Model instance.
+
+    """
+    # Try to determine model file type if not given explicitly
+    if model_format is None:
```
brynpickering: Is this tested?

doc/user/troubleshooting.rst
```diff
@@ -0,0 +1,52 @@
...
+* Try setting ``run.solver_options.Presolve: 0``, as large numeric ranges can cause the pre-solver to generate an infeasible or numerically unstable model.
+
+.. seealso::
+
+    The `Gurobi Guidelines for Numerical Issues <https://www.gurobi.com/documentation/current/refman/numerics_gurobi_guidelines.html>`_ give detailed guidance for strategies to address numerically difficult optimisation problems.
+
+Debugging model errors
+----------------------
+
+Calliope provides a method to save its fully built and commented internal representation of a model to a single YAML file with ``Model.save_commented_model_yaml(path)``. Comments in the resulting YAML file indicate where original values were overridden.
```
brynpickering: Does this not depend on us sticking with the current method of commented dicts, which depends on sticking with ruamel.yaml?

doc/user/troubleshooting.rst
```diff
@@ -0,0 +1,52 @@
...
+
+To deal with numerically unstable models:
+
+* Try setting ``run.solver_options.Presolve: 0``, as large numeric ranges can cause the pre-solver to generate an infeasible or numerically unstable model.
+
+.. seealso::
+
+    The `Gurobi Guidelines for Numerical Issues <https://www.gurobi.com/documentation/current/refman/numerics_gurobi_guidelines.html>`_ give detailed guidance for strategies to address numerically difficult optimisation problems.
+
+Debugging model errors
```
brynpickering: Perhaps also include `model.backend.access_pyomo_model_inputs`, since that will also include NaN values filled with defaults, basically the final parameter sets that pyomo will draw values from (another reason that group constraints should follow a default convention alongside the other constraints?)

doc/user/troubleshooting.rst
```diff
@@ -0,0 +1,52 @@
...
+* **Saving an LP file without running the model**: The LP file contains the mathematical model formulation of a fully built Calliope model. It is a standard format that can be passed to various solvers. Examining the LP file manually or using additional tools (see below) can help find issues when a model is infeasible or unbounded. To build a model and save it to LP without actually solving it, use:
+
+  .. code-block:: shell
+
+    calliope run my_model.yaml --save_lp=my_saved_model.lp
+
+Understanding infeasibility and numerical instability
+-----------------------------------------------------
+
+Using the Gurobi solver
```
brynpickering: I'll write something about CPLEX too

doc/user/building.rst
```diff
@@ -155,6 +155,50 @@ This will create an ``unmet_demand`` decision variable in the optimisation, whic
...
+        essentials:
+            name: 'Rooftop PV'
+            color: '#B59C2B'
+            parent: supply
+            carrier_out: power
+        constraints:
+            resource: file=pv_resource.csv
+            energy_cap_max: 10000  # kW
+
+All time series data in a model must be indexed by ISO 8601 compatible time stamps (usually in the format ``YYYY-MM-DD hh:mm:ss``, e.g. ``2005-01-01 00:00:00``), i.e., the first column in the CSV file must be time stamps.
```
brynpickering: Not strictly true, a user can also define the timestamp format that their data is stored in...

calliope/cli.py
sjpfenninger: No, but these were already missing before, and this PR already does a lot of stuff ;)

calliope/example_models/national_scale/scenarios.yaml
sjpfenninger: I would leave them until we actually have consistent minimal examples -- right now, the minimal examples are only visible if you actually dig through the source code looking at the tests (and who does _that_?)

doc/user/troubleshooting.rst
sjpfenninger: can you add that?

doc/user/troubleshooting.rst
sjpfenninger: This was just moved from where it previously lived in the documentation, nothing new actually implemented

calliope/test/test_constraint_results.py
sjpfenninger: These tests actually picked up issues with the group constraints that the group constraint tests did not, so they are not duplicating them (again, the group constraints need better tests before 0.6.4 release)

doc/user/building.rst
sjpfenninger: Addressed in ba3f4c5

sjpfenninger:
Updated in 5459ca3 to make default configuration more accessible and visible in the docs. To fully address #161, we would also empty the `required_constraints` lists for all abstract base techs, except leaving `resource` required for `demand` techs. I will hold off on that until we are clear we want to, as it will require updating / fixing a few tests. Does that sound sensible?

calliope/test/test_constraint_results.py
brynpickering: Yeah, I'd prefer that the group constraint tests are kept to their respective minimal examples, and that we don't have to pick up strange edge cases using the national-scale / urban-scale example models... If nothing else, it should keep test time down

brynpickering:
I can see some overlap with the new things you've added and existing documentation on improving solution times (https://calliope.readthedocs.io/en/latest/user/running.html#improving-solution-times). I wonder if they should just be cross-referenced, or whether the information from one should be wholly moved to the location of the other?

calliope/test/test_constraint_results.py
sjpfenninger: We should consider this when finishing up #193

sjpfenninger:
> I can see some overlap with the new things you've added and existing documentation on improving solution times (https://calliope.readthedocs.io/en/latest/user/running.html#improving-solution-times). I wonder if they should just be cross-referenced, or whether the information from one should be wholly moved to the location of the other?

Good point - done.

sjpfenninger:
@brynpickering I've removed most required constraints -- are we happy with that? This also removes the pre-processing check for "at least one constraint must be defined" (which wasn't actually checking for that, it was only checking for required constraints, not constraints generally). There could be smarter checks to add back in, but there are now so many ways in which a tech can be constrained, that it might be easiest to just leave it up to the user to ensure that what they doing is sensible.

brynpickering:
@sjpfenninger Made some updates to docs and to tests. Couple of things:

1. Not a fan of how inline code snippets look. Any white space is missed in the highlighting... 
image

2. Catching exceptions in CLI seems a bit difficult, I've implemented some tests to catch some excepted ValueErrors, but the current method is to hoover up all errors and then print it with a sys exit code of 1. If you know how to tap into that printout to check for the actual errors being tested, that would be good. At the moment, they could just exiting with code of 1 because the files being used don't exist (and not because they should fail before knowing that).

PR #216, opened 2019-04-01
Fix operate checks

brynpickering:
Fixes issue raised on Gitter by @FLomb

* Fixed backend check to allow parameters to be a scalar or array
* Added a test

Reviewer checklist:

- [x] Documentation updated
- [x] Changelog updated


brynpickering:
@FLomb, you could take a look at this PR about how a small change can be made (1 line!) with an associated test and changelog entry. Hope it helps to understand the process a bit better!

FLomb:
@brynpickering Thanks Bryn! So, for what I can see, the issue was wit the .item function, which needed to be replaced by .values with slightly different formulation of the code line. In addition to that, you added an ad-hoc test and updated the changelog. 

The code editing process is really clear, thanks. What I'm still stuck with is problems with git/conda. I tried to repeat the key operation (i.e. the change in the checks.py file) on my locally-installed "dev" version of Calliope, but it looks like code changes implemented on my local repo are not "seen" by my conda environment. I even tried to push changes back to the online repo, but still the conda environment remains identical to itself. I'll try to re-read carefully how to install "dev" versions and see if something's missing from what I've done some months ago.

sjpfenninger:
@FLomb Hm, the key is to use `pip install -e path/to/your/modified/calliope` from within the conda env you want to use for development, and to install only all calliope dependencies but not calliope itself via conda. Seems like that may be the issue in your case?

FLomb:
Thanks @sjpfenninger ! Your hint was correct, problem finally solved!

Issue #172, opened 2018-12-07
Better error if no locations are defined


sjpfenninger:
## Problem description

Better error if no locations are defined

Issue #164, opened 2018-11-30
Check common errors with nested keys in YAML configuration


brynpickering:
## Problem description

These all currently silently fail:

* If defining a cost without a cost class
e.g. `costs.om_annual`
* If defining an override to a tech at a specific loc, without nesting that tech in `techs`
e.g. `locations.X.ccgt.constraints.energy_cap`
* If defining a carrier ratio for a `conversion_plus` tech, but applying it to a carrier that isn't one of the carriers specified by that technology
e.g. `carrier_ratios.carrier_in_2.cooling` when `cooling` isn't a carrier

## Calliope version

0.6.3

PR #219, opened 2019-04-01
Improved checks and errors

sjpfenninger:
Improved checks and errors

Addresses the following issues:

* #164: Check common errors with nested keys in YAML configuration
    * Warn if defining a cost without a cost class: e.g. costs.om_annual
    * Error and stop if defining an override to a tech at a specific loc, without nesting that tech in techs: e.g. locations.X.ccgt.constraints.energy_cap
    * Warn if defining a carrier ratio for a conversion_plus tech, but applying it to a carrier that isn't one of the carriers specified by that technology: e.g. carrier_ratios.carrier_in_2.cooling when cooling isn't a carrier
* #172: Better error if no locations are defined
* Minor cleanup of global static values in checks.py
* Minor modification to AttrDict to work with pathlib.Path

Reviewer checklist:

~- [ ] Documentation updated~
- [x] Changelog updated


calliope/core/preprocess/checks.py
```diff
@@ -212,8 +226,18 @@ def check_initial(config_model):
                 'be defined (tech: {})'.format(t_name)
             )
 
+    # Check whether any techs are erroneously in top-level location definitions
```
brynpickering: I would definitely have preference for checking whether the loc_key subdict has any unexpected keys from `[techs, links, available_area, coordinates]` than this check, since it would cover someone not nesting a tech under `techs`, but misspelling that tech, for instance...

calliope/core/preprocess/locations.py
```diff
@@ -397,15 +397,27 @@ def process_per_distance_constraints(tech_name, tech_settings, locations, locati
...
+
+        # Warning if a cost class is empty
         if not isinstance(tech_config.costs[cost], dict):
             warnings.append(
-                'Deleting empty cost class `{}` for technology `{}`.'.format(cost, tech_id)
+                'Deleting empty cost class `{}` for technology `{}` at `{}`.'.format(
+                    cost, tech_id, loc_or_link)
             )
             tech_config.costs.del_key(cost)
+            # If the cost class is empty, it is not enough to delete it,
```
brynpickering: This comment isn't quite right, it's skipping the rest of the loop because it has deleted it (the rest of the loop expects the cost class to exist...)

calliope/core/preprocess/checks.py
brynpickering: Also, this same check should be added for links, right?

sjpfenninger: Adding check for links, but a more complete solution will have to wait.

PR #221, opened 2019-04-02
Change auto load constraints

brynpickering:

* All constraint files are loaded by default.
* Constraint files are not hardcoded, but loaded automatically from the 'constraints' folder in the pyomo backend.
* Order of constraint application is dictated by the 'ORDER' variable in every constraint file.

Reviewer checklist:

~- [ ] Documentation updated~
- [x] Changelog updated


brynpickering:
Strange test failure as it passes locally and in appveyor, and isn't anything to do with the changes in the PR. Will see if it resolves itself on merging in changes from other PRs (once they're merged into master).

sjpfenninger:
@brynpickering might the test failures be caused by order of constraint loading now being random within the same priority?

brynpickering:
Hmm, could be. I allowed the priority to be random, seeing as for most constraints, the order of application doesn't matter. But I may have added more to the first group than I should have... Will have a look...

brynpickering:
@sjpfenninger should now be good to go

calliope/backend/pyomo/model.py
```diff
@@ -101,32 +102,23 @@ def generate_model(model_data):
...
     # Constraints
     constraints_to_add = [
-        'energy_balance.load_constraints',
-        'dispatch.load_constraints',
-        'network.load_constraints',
-        'costs.load_constraints',
-        'policy.load_constraints',
-        'milp.load_constraints',
-        'group.load_constraints'
+        i.split('.py')[0] for i in os.listdir(constraints.__path__[0])
```
sjpfenninger: `i.strip('.py') for i in os.listdir(constraints.__path__[0]) if i != '__init__.py'` would seem to do the same thing but a little less arcane?

calliope/backend/pyomo/model.py
```diff
@@ -101,32 +102,23 @@ def generate_model(model_data):
...
-        constraints_to_add.append('conversion.load_constraints')
-
-    if hasattr(backend_model, 'loc_techs_conversion_plus'):
-        constraints_to_add.append('conversion_plus.load_constraints')
+    # The list is sorted to ensure that some constraints are added after pyomo
+    # expressions have been created in other constraint files.
+    # Ordering is given by the number assigned to the variable ORDER within each
+    # file (higher number = added later).
+    constraints_to_add.sort(
+        key=lambda x: load_function('calliope.backend.pyomo.constraints.' + x + '.ORDER')
```
sjpfenninger: This would blow up if a constraint file does not have ORDER, right? Maybe should files without an ORDER have a very large ORDER number by default and a warning be raised about that?

calliope/backend/pyomo/model.py
```diff
@@ -101,32 +102,23 @@ def generate_model(model_data):
...
-        constraints_to_add.append('capacity.load_constraints')
-
-    if hasattr(backend_model, 'loc_techs_conversion'):
-        constraints_to_add.append('conversion.load_constraints')
-
-    if hasattr(backend_model, 'loc_techs_conversion_plus'):
-        constraints_to_add.append('conversion_plus.load_constraints')
+    # The list is sorted to ensure that some constraints are added after pyomo
+    # expressions have been created in other constraint files.
+    # Ordering is given by the number assigned to the variable ORDER within each
```
sjpfenninger: I feel a bit uneasy about having lots of files with the same value for ORDER. How about using increments of 10 and starting with 10 for energy_balance, 20 for dispatch, and so on in the order that they are currently loaded in, permitting enough space to hack stuff in between (a bit like logging levels perhaps)?

calliope/backend/pyomo/model.py
sjpfenninger: And I wonder whether this warrants a test?

calliope/backend/pyomo/model.py
brynpickering: I think it is good to be clear about what things need to be loaded in after other things. In reality, only a few files need to be loaded after others (because preceding files attach new objects to the backend model) and it is a lot more clear that those files are different if only they are loaded in later. But changing to increments of 10 makes a lot of sense.

calliope/backend/pyomo/model.py
brynpickering: RE files without order. I can add in the warning and the forced order position.
RE tests, I was trying to think of how to do a test on this, but couldn't create a viable one. Ideally I would capture the debug comment about which files have been loaded, but I couldn't get that to pop up when using the pytest [logging capture function](https://docs.pytest.org/en/latest/logging.html).

calliope/backend/pyomo/model.py
brynpickering: I decided to be more generic to account for the existence of any other non-constraint files in that folder (i.e. files can be hidden from being loaded if they are double underscored). Happy to be more explicit though, if you think this functionality is worthless.

sjpfenninger: I would only specifically exclude `__init__.py` as the other behaviour might be unexpected. Unless there is a convention in Python to ignore double-underscored files?

calliope/backend/pyomo/model.py
sjpfenninger: Ok, so we're moving this to increments of 10, right?

calliope/backend/pyomo/model.py
brynpickering: I also excluded `__pycache__`, since that shows up locally for me.

calliope/backend/pyomo/model.py
brynpickering: Done

calliope/backend/pyomo/model.py
brynpickering: Right

PR #223, opened 2019-04-04
Add new energy_cap_per_storage_cap_min constraint

timtroendle:
In our discussion about `energy_cap_per_storage_cap_x` we did not see a use case for `x=min`. I do have one now. It's similar to the idea of equals and thus not modeling electrochemical relationships. Using this constraint, you can let the optimiser find the optimal energy_cap/storage_cap of a battery for example, but you can limit the installed storage capacity per installed power by setting `energy_cap_per_storage_cap_min` to e.g. 0.25 -- i.e. modelling batteries which are depleted after at most 4h.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


timtroendle:
Agreed. I updated the patch.

PR #225, opened 2019-04-05
Fix the names of the new constraints

timtroendle:

After running the changes of #223 manually, I found it was broken. 😬 Here's a fix.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/calliope/backend/pyomo/constraints/capacity.py
+++ b/calliope/backend/pyomo/constraints/capacity.py
@@ -26,25 +26,25 @@ def load_constraints(backend_model):
         )
 
     if 'loc_techs_energy_capacity_storage_min_constraint' in sets:
-        backend_model.energy_capacity_storage_constraint = po.Constraint(
+        backend_model.energy_capacity_storage_min_constraint = po.Constraint(
             backend_model.loc_techs_energy_capacity_storage_min_constraint,
             rule=energy_capacity_storage_min_constraint_rule
         )
 
     if 'loc_techs_energy_capacity_storage_max_constraint' in sets:
-        backend_model.energy_capacity_storage_constraint = po.Constraint(
+        backend_model.energy_capacity_storage_max_constraint = po.Constraint(
             backend_model.loc_techs_energy_capacity_storage_max_constraint,
             rule=energy_capacity_storage_max_constraint_rule
         )
 
     if 'loc_techs_energy_capacity_storage_equals_constraint' in sets:
-        backend_model.energy_capacity_storage_constraint = po.Constraint(
+        backend_model.energy_capacity_storage_equals_constraint = po.Constraint(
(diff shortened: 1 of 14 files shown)


brynpickering:
🤦‍♂️ That will teach me for merging a PR from my phone...

brynpickering:
@timtroendle I noticed that one of the tests you updated was actually an incorrectly prepared test from a while ago... I've updated that test and made a few other changes; have a look to see if you're happy with them. 

Most major one is removing the requirement for defining one of these new constraints, since we've now removed all required constraints. Instead it now just raises a warning.

timtroendle:
Looks good to me! Only thing I would do differently: I would not issue a warning for storage technologies without an energy_cap_per_storage_cap constraint. Calliope failed before, but only for `supply_plus`, not for `storage` technologies. And I think it is perfectly fine to not wanting to model a relationship between energy_cap and storage_cap.

But I don't think it is very important and I am fine with either way.

brynpickering:
Actually, they were required for both, the only difference is that storage techs had it as a 'required_constraint' in the `model.yaml` config, while supply_plus had it as a required constraint *only* if storage is defined (given that it is an optional component of supply_plus). 

Not having a link between the two is fine, but my only concern is that it isn't necessarily obvious to a user that there is no link by default. But I'm also fine either way :P @sjpfenninger want to cast the deciding vote?

sjpfenninger:
@brynpickering Hm, this makes me think there should be different levels of verbosity for warnings. Warning about not having linked power and energy seems like it is a debug-level warning that one wouldn't normally care about. I'm more inclined to get rid of it because it would show up in many cases where nothing actually needs attention.

timtroendle:
Calliope does use Python's logging system, no? Could this be done using the log levels of the logging system?

brynpickering:
@timtroendle it does, and we have already placed a custom log level in there.

@sjpfenninger I thought this too, although perhaps not as low as DEBUG, since that prints out an unending stream of data from pyomo. Perhaps there should be a model debug and a run debug level?

I seem to remember that custom log levels are kinda frowned upon in published packages, since they could clash with upstream/downstream dependency log levels, but I don't think that's too much of a concern...

timtroendle:
@brynpickering you should be able to use the same level (DEBUG), but filter the logger. So there could be a log that sees all debug log statements from Calliope but not from Pyomo.

brynpickering:
Aha, I should clarify: they're our logs, just that they print out the result from functions in the Calliope backend, such as `utils.get_param`, which ends up being quite verbose...

timtroendle:
Oh I see. Still, I believe one should be able to separate the Calliope internal loggers as well, even on the same log level.

sjpfenninger:
For now, I'll merge this as-is and open a new issue for changes to logging levels.

Issue #210, opened 2019-03-15
Handling of storage and transmission techs in levelised cost


timtroendle:
## Problem description

This is not exactly a problem, but rather a definition issue: Calliope considers generation of storages when calculating `systemwide_levelised_cost` --- personally, I would not. For the per technology level, levelised cost may be interesting. But for the total levelised cost (i.e. all locs and techs),  you'd want to know the costs of actually generated energy, i.e. the amount of energy that equals energy demand excluding storage and losses. At least I would.

Maybe this has been discussed already and there has been a decision to do it the way it is currently done? If yes, maybe there could be a second `systemwide_levelised_cost_net_energy` or so?

## Calliope version

0.6.4.dev

brynpickering:
Agreed that excluding storage and transmission would seem to make the most sense (or, like with the group share constraints, having the ability to either specify `demand` or `supply`?). With the two of them in at the moment, it will be double counting: energy production to charge storage and energy production on discharging the storage (similarly for transmission). Not sure it was a conscious decision to include those two tech types.

timtroendle:
I tried to implement this, but failed because I do not understand where the information about the type of technology is stored. I found some sets, but not for supply technologies.

I also found `model._backend_model.loc_tech_carriers_supply_all` but that's not available in the function and it feels brittle to rely on the backend here.

Any hints on how to properly filter technologies?

timtroendle:
Found it. 🎉

PR #229, opened 2019-04-16
Fix total lcoe

timtroendle:
Fixes issue(s) #210 

Use only supply all technologies when calculating the total lcoe.

Reviewer checklist:

~- [ ] Documentation updated~
- [x] Changelog updated

### DIFF
--- a/calliope/test/common/test_model/energy_cap_per_storage_cap.yaml
+++ b/calliope/test/common/test_model/energy_cap_per_storage_cap.yaml
@@ -30,10 +30,20 @@ techs:
                 energy_cap: 1500000
                 storage_cap: 300000
                 interest_rate: 0.1
+    my_dummy_supply:
+        # Unfortunately, the only supply tech in the region is broken.
+        # (Calliope expects at least one supply technology.)
+        essentials:
+            name: Broken Supply
+            carrier: electricity
+            parent: supply
+        constraints:
+            energy_cap_max: 0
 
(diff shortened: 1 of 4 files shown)


calliope/test/common/test_model/energy_cap_per_storage_cap.yaml
```diff
@@ -30,10 +30,20 @@ techs:
                 energy_cap: 1500000
                 storage_cap: 300000
                 interest_rate: 0.1
+    my_dummy_supply:
+        # Unfortunately, the only supply tech in the region is broken.
+        # (Calliope expects at least one supply technology.)
```
timtroendle: This is not a very smart solution to the problem. I still decided to handle the special case in here instead of handling it in the code, because

1. It's a rare special case that people should not run into.
2. I tried handling it in the code, but Calliope is not prepared for this case and breaks in more than one location. First, the set of supply techs does not exist (I'd expect it to be empty, but it simply doesnt exist). Second, after fixing the set, ``split_loc_techs`` breaks with non existing techs. Instead of changing that as well I decided to keep the special case in the test.

sjpfenninger: That's fine for now. If there is ever a practical application for buildings models without supply techs, this could always be revisited.

sjpfenninger:
Thanks @timtroendle!

Issue #245, opened 2019-05-28
save_commented_model_yaml fails with cryptic error message


timtroendle:
## Problem description

When trying to generate a fill YAML model description from model results, `model.save_commented_model_yaml()` fails with a cryptic `AttributeError: 'NoneType' object has no attribute 'copy'` error message. I assume it's because `model_run` doesn't exist at this point, but that's an implementation detail.

I think having a full model description, potentially months or years after actually running the model, is highly valuable. The netcdf file is great for timeseries but not good for the model description. If the `model_run` cannot be regenerated, maybe it can be stored completely within the netcdf? I used to do that in the past: our model results were stored in a HDF5 file, and we copied the full model configuration as a varstring in a 1x1 array. Maybe not the most elegant solution, but absolutely fulfilled its job.

If the YAML cannot be regenerated or stored, I think there should be a clear error message: first, this is normal and simply cannot be done at this point. Second, pointing the user to a working way of understanding the model that is at the bottom of the model results.

## Steps to reproduce the problem

```Python
import calliope
model = calliope.read_netcdf("my-model-results.nc")
model.save_commented_model_yaml("full-model.yaml")
```

## Calliope version

0.6.4.dev

sjpfenninger:
Fixed in #247

PR #247, opened 2019-05-30
Let model_run and debug data roundtrip through NetCDF

sjpfenninger:
Fixes issue(s) #245

* Let model_run and debug data roundtrip through NetCDF

There seems to be no immediate limit on the size of strings in NetCDF attributes -- a test with a 20 MB large string did not reveal any performance or other issues (except that it adds 20 MB to the size of the resulting NetCDF file).

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/calliope/core/model.py
+++ b/calliope/core/model.py
@@ -113,6 +113,7 @@ class Model(object):
             time_since_start=True
         )
 
+        # Ensure model and run attributes of _model_data update themselves
         for var in self._model_data.data_vars:
             self._model_data[var].attrs['is_result'] = 0
         self.inputs = self._model_data.filter_by_attrs(is_result=0)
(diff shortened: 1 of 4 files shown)


calliope/core/model.py
```diff
@@ -125,9 +126,19 @@ def _init_from_model_run(self, model_run, debug_data):
             name='run_config', observer=self._model_data
         )
 
+        # Attach _model_run and _debug_data to _model_data
+        _model_run_to_save = self._model_run.copy()
+        del _model_run_to_save['timeseries_data']  # Can't be serialised!
+        self._model_data.attrs['_model_run'] = _model_run_to_save.to_yaml()
```
brynpickering: I would have preference for `_model_run` only being added into the attrs at the point where a user saves the model to netcdf, i.e. you never see it in `inputs` or `results`, and it is never actually attached to `_model_data`, only to a copy of it when it is saved...

calliope/core/model.py
```diff
@@ -125,9 +126,19 @@ def _init_from_model_run(self, model_run, debug_data):
...
+        self._model_data.attrs['_debug_data'] = self._debug_data.to_yaml()
+
     def _init_from_model_data(self, model_data):
-        self._model_run = None
-        self._debug_data = None
+        self._model_run = AttrDict.from_yaml_string(
```
brynpickering: If the above comment were agreed upon, then this would mean deleting it from the attrs after attaching it to the model as `_model_run`. Maybe it isn't worth it then

calliope/core/model.py
sjpfenninger: Done

calliope/core/model.py
sjpfenninger: Done

PR #257, opened 2019-06-12
Fix saving to netcdf after reading from netcdf

sjpfenninger:

* Fix saving to netcdf after reading from netcdf

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/calliope/test/test_io.py
+++ b/calliope/test/test_io.py
@@ -86,7 +86,11 @@ class TestIO:
             model.to_netcdf(out_path)
             model_from_disk = calliope.read_netcdf(out_path)
 
+        # Ensure _model_run doesn't exist to simulate a re-run
+        # via the backend
+        delattr(model_from_disk, '_model_run')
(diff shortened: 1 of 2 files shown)


calliope/test/test_io.py
```diff
@@ -80,6 +79,20 @@ def test_solve_save_read_netcdf(self, model):
             model_from_disk = calliope.read_netcdf(out_path)
             # FIXME test for some data in model_from_disk
 
+    def test_save_read_solve_save_netcdf(self, model):
```
brynpickering: Is this test actually working with a model that doesn't have `model_run`? Doesn't `model_run` get added back onto the object on reading it back in, if it exists in the netcdf attrs?

brynpickering: Maybe getting a new model from running with the backend interface, then saving *that* to file would work as another / replacement test?

PR #261, opened 2019-06-18
Fix warning demand cost

FLomb:
Fixes issue(s) related to the recently allowed om_con cost for demand techs. In fact, also group constraint defaults need to be updated to allow generic and variable costs group constraints for demand techs.
Otherwise, demand variable costs will be allowed in general but, if a cost group constraint is defined, checks in the preprocess will exclude demand costs from the cost group constraint and raise a warning.

* Added generic costs and variable costs to group constraints allowed by demand techs
* Existing tests should cover this change already
*

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


brynpickering:
> Existing tests should cover this change already

I'm not sure we have cost_max/_min/_equals group constraint tests any more, since group constraint tests are undergoing an overhaul. But I'll merge this in, so the test rewrites include the fact that demand is included in these constraints!

Issue #270, opened 2019-07-09
Limit net imports


timtroendle:
## Problem description

I would like to limit the net imports of a certain carrier into a location or group of locations over the entire time horizon of my model. I thought I could do that with Calliope 0.6.4 in the following way: add a demand_share_min constraint to all internal supply technologies, so that they generate the expected share.

While this works, it's not the same as limiting the net imports: actual carrier consumption is higher within the locations because there are losses in storage and transmission. Using the method described above, those losses can be fed with net imports from the grid. Thus, there are in fact higher net imports in the end. In a case in which I disallow all net imports, I actually see 5% net imports.

Instead one should probably model net import limitations with a demand_share_max constraint on the transmission links. Transmission links cannot be part of group constraints atm though.

## Steps to reproduce the problem

```yaml
group_constraints:
    import_restriction_0_percent:
        locs: ["1"]
        techs: ["ac_transmission"] # this does not work
        demand_share_max:
            electricity: 0.0
```

## Calliope version

0.6.4

brynpickering:
Yup, it would be good to include transmission technologies in the group constraints. They do, however, need a lot of special treatment. You could have transmission be associated with a new carrier (e.g. AC_transmission) and have a conversion tech to convert that to demanded electricity at each location. Then you could place the group constraint on the conversion tech?

timtroendle:
Thanks a lot for the idea. Sounds like a pragmatic workaround. I'm only afraid it would make result analysis slightly more complicated (I very much got used to `squeeze("carriers")` 😉) and much more importantly, having two carriers and having conversion technology may make the problem more complicated to solve. Certainly it's worth a try.

brynpickering:
Well, you now just need to get used to using `sum("carriers")` 😉. I would be certainly interested to hear what the slow-down is. I think the commercial solvers are quite good at preprocessing these kinds of implementations to the same sized optimisation problem.

timtroendle:
I'll let you know.

timtroendle:
Actually, I fear this doesn't work either without changes in Calliope. It works well for restricting _gross_ imports, but not for _net_ imports. At least I do not see a way of restricting net imports. I can restrict the `carrier_prod` of any supply tech, but I would need to restrict the sum of `carrier_prod` plus `carrier_con` (either of the same tech if possible, or of two techs).

Perhaps this could be solved by optionally choosing demand techs (already discussed in #227):

```yaml
techs:
    to_grid:
        essentials:
[...]

PR #271, opened 2019-07-24
Add net import share contraints

timtroendle:
Fixes issue(s) #270 

* add a new constraint constraining net imports

Reviewer checklist:

- [x] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/calliope/backend/pyomo/constraints/group.py
+++ b/calliope/backend/pyomo/constraints/group.py
@@ -92,6 +92,15 @@ def load_constraints(backend_model):
                 )
             )
 
+        if 'group_net_import_share_{}'.format(sense) in model_data_dict:
+            setattr(
+                backend_model, 'group_net_import_share_{}_constraint'.format(sense),
+                po.Constraint(
+                    getattr(backend_model, 'group_names_net_import_share_{}'.format(sense)),
+                    backend_model.carriers, [sense], rule=net_import_share_constraint_rule
+                )
+            )
+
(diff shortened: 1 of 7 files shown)


timtroendle:
Unfortunately, one cannot explicitly define a transmission tech -- at least I do not manage to. I wrote a failing test. However, if you do not define techs, you'll receive warnings.

sjpfenninger:
I think this is fine, but perhaps the docs should mention that one cannot explicitly select transmission techs with this constraint (only implicitly via carriers). What warnings do you receive?
Also can you update CHANGELOG. Otherwise looks good.

timtroendle:
> What warnings do you receive?

When no tech is specified, one sees this warning:

>   * All technologies were requested for inclusion in group constraint `example_net_import_share_constraint`, but those from tech group(s) `['conversion', 'conversion_plus', 'demand', 'storage', 'supply', 'supply_plus']` have been ignored as one or more of the constraints cannot be applied to technologies in these groups

When a tech is specified, one sees this warning which turns into an error lateron:

> * Group constraint `example_net_import_share_constraint` cannot be applied as there are no valid loc_tech combinations available.

sjpfenninger:
Ok, I'm willing to merge this as-is for now with the intention of it being cleaned up later. @timtroendle can you open an issue linking to this PR to track the facts that (1) we need to deal with transmission techs in `techs:` here properly and (2) we need to revisit what warnings are raised for this kind of constraint?

PR #274, opened 2019-07-31
Fix gross imports

timtroendle:
Fixes feature implemented in #271.

* add tests for gross imports
* fix constraint so as to allow gross imports

My bad, I really should have written a test for gross imports in the first place! In my defence I have to say that this is based on a rather nasty Pyomo behaviour: `sum(imports) + sum(exports)` somehow was translated to `sum(imports)` which is odd. The solution now is `sum(imports + exports)`.

Reviewer checklist:

~~- [ ] Changelog updated~~


brynpickering:
Very odd, I don't see why it should act like that at all!

PR #275, opened 2019-08-01
Fix double logging when using Gurobi

### LINKED ISSUE #272: Solver output is logged twice to console
## Problem description

The solver output is logged twice to console: one time with DEBUG level, one time with INFO level. See below. 

All Calliope logs before and after the solver output are just fine and on INFO level only.

```
[2019-07-24 16:28:02] DEBUG    Academic license - for non-commercial use only
[2019-07-24 16:28:02] INFO     Academic license - for non-commercial use only
[2019-07-24 16:28:05] DEBUG    Parameter OutputFlag unchanged
[2019-07-24 16:28:05] INFO     Parameter OutputFlag unchanged
[2019-07-24 16:28:05] DEBUG    Value: 1  Min: 0  Max: 1  Default: 1
[2019-07-24 16:28:05] INFO        Value: 1  Min: 0  Max: 1  Default: 1
[2019-07-24 16:28:05] DEBUG    Changed value of parameter LogFile to /var/folders/77/hl9n2z_n6059xs8ml53n_tsr0000gp/T/tmp8uah5hc7.log
[2019-07-24 16:28:05] INFO     Changed value of parameter LogFile to /var/folders/77/hl9n2z_n6059xs8ml53n_tsr0000gp/T/tmp8uah5hc7.log
[2019-07-24 16:28:05] DEBUG    Prev: gurobi.log  Default:
[2019-07-24 16:28:05] INFO        Prev: gurobi.log  Default:
```

## Steps to reproduce the problem

Run any model.

## Calliope version

0.6.5-dev

timtroendle:
Fixes #272 (a regression introduced in #248)

* add a test for double logging with Gurobi
* mute Gurobi's direct stdout logging

While I think it's totally fine to mute Gurobi's direct stdout printing and rely on their additional logging (why do they do both?), this solution is not consistent with other loggers. This way, Gurobi logs will be INFO level rather than DEBUG level.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/calliope/backend/pyomo/model.py
+++ b/calliope/backend/pyomo/model.py
@@ -193,6 +193,8 @@ def solve_model(backend_model, solver,
 
     with redirect_stdout(LogWriter(logger, 'debug', strip=True)):
         with redirect_stderr(LogWriter(logger, 'error', strip=True)):
+            # Ignore most of gurobipy's logging, as it's output is
+            # already captured through STDOUT
             logging.getLogger('gurobipy').setLevel(logging.ERROR)
             results = opt.solve(backend_model, tee=True, **solve_kwargs)
     return results
--- a/calliope/backend/pyomo/model.py
+++ b/calliope/backend/pyomo/model.py
@@ -6,6 +6,7 @@ Licensed under the Apache 2.0 License (see LICENSE file).
 
 import logging
 import os
+import io
 from contextlib import redirect_stdout, redirect_stderr
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


brynpickering:
Check [this comment](https://github.com/calliope-project/calliope/issues/272#issuecomment-517741952_) out for a solution that suppresses gurobipy at the info level.

Looking at pyomo, there is also a python interface for CPLEX and MOSEK, but I don't see a logger for CPLEX and MOSEK isn't something we know that anyone uses. So for the time being, I think just dealing with gurobi is fine.

brynpickering:
Also: the gurobipy logger is loaded whether or not a user is solving with gurobi (or using the python solver interface), so it should be fine to just suppress it without there needing to be the condition that they specifically requested gurobi as the solver

PR #297, opened 2020-01-29
SPORES implementation as a Calliope run mode

FLomb:

* Created a "spores" run mode, currently with default settings (e.g. number of SPORES) to be later made user-defined
* Added dimension "spores" to model_data and results AttrDicts to manage the multiple solutions without significantly altering the structure of the post-process
* [**To do**: adjust saving/reading/plotting to avoid current issues with the new "spores" dimension]
* [**To do**: implement checks and tests to ensure internal code reliability]

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


calliope/backend/run.py
```diff
@@ -121,6 +129,115 @@ def run_plan(model_data, timings, backend, build_only, backend_rerun=False):
...
+    For use when mode is 'spores', to allow the model to be built, edited, and
+    iteratively run within Pyomo, modifying, at each iteration, the score of
+    each loc::tech in such a way to generate Spatially explicit Practically Optimal
+    RESults (SPORES).
+
+    """
+    log_time(logger, timings, 'run_start',
+             comment='Backend: starting model run in SPORES mode')
+
+    run_config = UpdateObserverDict(
```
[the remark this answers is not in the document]
brynpickering: this will be used once todos are dealt with

calliope/backend/run.py
```diff
@@ -121,6 +129,115 @@ def run_plan(model_data, timings, backend, build_only, backend_rerun=False):
...
+        initial_yaml_string=model_data.attrs['run_config'], name='run_config', observer=model_data
+    )
+
+    n_spores = 1  # TODO: turn into configurable in run_config
```
brynpickering: There needs to be configurability, which is best within the 'run' configuration settings. E.g. under 'spores_options'.

calliope/core/preprocess/locations.py
```diff
@@ -389,8 +391,14 @@ def process_per_distance_constraints(tech_name, tech_settings, locations, locati
     return tech_settings
 
 
-def check_costs_and_compute_depreciation_rates(tech_id, loc_or_link, tech_config, warnings, errors):
+def check_costs_and_compute_depreciation_rates(tech_id, loc_or_link, tech_config, warnings, errors, run_mode, parent):
     cost_classes = list(tech_config.get('costs', {}).keys())
+    # Adds spores_score details if "spores" run mode is selected
+    if run_mode == 'spores' and parent != 'demand': # not in str(tech_id): # not working: the idea is to find a way to skip demand techs, by checking the parent tech in tech_config, if possible
```
brynpickering: I would try to avoid anything to do with run mode here. The point of 'model' and 'run' being separate is that anything to do with the run mode should not influence anything until a user runs `model.run()`. It's worth thinking about how that influences what is done in this file.

calliope/core/preprocess/locations.py
```diff
@@ -420,11 +428,19 @@ def check_costs_and_compute_depreciation_rates(tech_id, loc_or_link, tech_config
                 for i in tech_config.costs[cost].keys()):
             # MUST define lifetime and interest_rate for these technologies
             if plant_life == 0 or interest is None:
-                errors.append(
-                    'Must specify constraints.lifetime and costs.{0}.interest_rate '
-                    'when specifying fixed `{0}` costs for `{1}`. Set lifetime to 1 '
-                    'and interest rate to 0 if you do not want them to have an effect'.format(cost, tech_id)
-                )
+                if run_mode == 'spores':
```
brynpickering: Why is this needed?

calliope/core/preprocess/model_run.py
```diff
@@ -573,6 +573,9 @@ def generate_model_run(config, debug_comments, applied_overrides, scenario):
     model_run['run'] = config['run']
     model_run['model'] = config['model']
     model_run['group_constraints'] = config.get('group_constraints', {})
+    if model_run['run']['mode'] == 'spores':
```
brynpickering: As with `locations.py`, if we can avoid magically generating the group constraint here, it would be ideal. Is it possible to do it in `run.py` in the backend? Could we access the constraint directly, or force users to set up the group constraint in their yaml config if they want to use spores properly? e.g. they could have to give the name of the cost group constraint in the run configuration for spores:
```
run.spores_options: {score_cost_class: 'spores_score', cost_group: 'systemwide_cost_max', slack: 0.1}
```

calliope/example_models/national_scale/model.yaml
```diff
@@ -17,7 +17,7 @@ model:
 
 # Run configuration: all settings that affect how the built model is run
 run:
-    solver: cbc
+    solver: gurobi
```
brynpickering: change back

calliope/backend/run.py
FLomb: Sure, this was in my TODO list already

brynpickering:
@FLomb the implementation looks good now! We still need a test or two and documentation, though...

brynpickering:
@FLomb this could be updated by formatting it with black: install [black](https://black.readthedocs.io/en/stable/), run it in your local Calliope repository, then commit that. You should then be able to merge in latest master with minimal merge conflicts

brynpickering:
@FLomb so, testing is still necessary on this implementation. I'd say the following could do with being tested:

1. missing parameters on choosing the spores run mode (n. spores, score cost class, slack, pointing to group constraint + existence of group constraint, etc.). This would likely need to be linked to a check in `core.preprocess.checks.py` which fails if everything hasn't been specified as expected.

2. spores run mode is correctly initialised in Pyomo (i.e. `build_only = True`)

3. spores run mode produces expected results on running for 1 or 2 spores. Then you can check that the group constraint and loc_tech spore scores have been correctly assigned as well as the results dataset looking like we expect.

Something we talked about was saving each spore to file on solving it - is that an option we still want to add?

brynpickering (approved):

PR #308, opened 2020-04-14
Update .gitignore

sjpfenninger:
Propose this more subtantial restructure in favour of  #304, with the understanding that we add  add any other OS- and tooling-specific things

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,13 +1,16 @@
+# OS
+.DS_Store
+
+# Tooling
+.cache
 .coverage
+.mypy_cache/
+.pytest_cache/
+.vscode
+*.egg-info/
 *.pyc
-calliope.egg-info/
+
+# Project
+dist/
 doc/_build/
 doc/_static/notebooks/*.html
-dist/
-.cache
-.pytest_cache/
-mprofile_*
-profile_*
-*.profile
-_test
-.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -1,145 +1,31 @@
-# Byte-compiled / optimized / DLL files
-__pycache__/
-*.py[cod]
-*$py.class
-
-# C extensions
-*.so
-
-# Distribution / packaging
-.Python
-build/
-develop-eggs/
-dist/
-downloads/
-eggs/
-.eggs/
-lib/
-lib64/
-parts/
-sdist/
-var/
-wheels/
-share/python-wheels/
-*.egg-info/
-.installed.cfg
-*.egg
-MANIFEST
-
-# Editor
-.vscode
-
-# Installer logs
-pip-log.txt
-pip-delete-this-directory.txt
-
-# PyInstaller
-#  Usually these files are written by a python script from a template
-#  before PyInstaller builds the exe, so as to inject date/other infos into it.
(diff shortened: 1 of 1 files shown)


brynpickering:
We could further extend it by using one of the [GitHub example gitignores](https://github.com/github/gitignore)

brynpickering:
Maybe it's excessive, tbh, but it also "future-proof's" the gitignore... Thoughts @suvayu ?

suvayu:
Hmm, I prefer sticking to tooling only.  There's no right answer here.  I guess the question is do we want to entertain random PRs every now and then to add something specific to a user's environment.  I would vote no there.

Even with the tooling related additions, we can optimise it a bit making any need to update even for tooling, less common.  For example:
- my original proposal was `.mypy_cache`,
- @sjpfenninger added `.pytest_cache`, `.vscode`, etc, and
- maybe in the future we start using `hypothesis` which adds `.hypothesis`.

As we know, tooling changes fast these days, so this is not a particularly far off scenario.  So an "optimisation" could be
```
.*[cC]ache
*.json
```
The first catches most cache/settings directories, and for the ones that are missed, the cache/settings files are ignored with the JSON glob.  We could also add YAML, we can always use `-f` when manually adding JSON or YAML files, not that that is a common occurrence.

I did a quick check. This seems to do the trick quite well:
```
# editors
*~
\#*\#
.vscode

# byte-compiled files
__pycache__

# build output
build
*.egg-info

# typical cache directories
.*[cC]ache*

# cache files/dumps
*.json
*.yaml
*.log
*.txt

# reports
*.html
*.js
*.css

.coverage
.hypothesis

# binary
*.png
*.pdf

# archives/compressed files
*.tar
*.zip
*.gz
*.bj2
*.xz
```
I thought of un-anchoring the cache directory glob (would have included `__pycache__`), but then I thought it's quite possible to write a caching module, so went back on it.

brynpickering:
Hmm, yeah, the things I expect to get through the net in PRs is tooling, editor/IDE, and OS-specific things, since they are often user-dependent. For instance, you missed the macOS `.DS_Store` in your list ;)

It comes back to the question of how much we push onto users to put in their global gitignore. Perhaps we should expect editor and OS-specific options to go in users' global gitignores, and only tooling goes into this one. 

Another note on your selection, ignoring YAML will create some havoc with our internal example models ;)

suvayu:
> It comes back to the question of how much we push onto users to put in their global gitignore. Perhaps we should expect editor and OS-specific options to go in users' global gitignores, and only tooling goes into this one.

Agreed

> Another note on your selection, ignoring YAML will create some havoc with our internal example models ;)

Not really.  The way Git treats ignore patterns is:
- if a file is in the repo, `.gitignore` doesn't matter
- any new files matching a pattern in `.gitignore` are rejected with a warning when you do `git add <file>`, or `git commit -a`.
- any ignored file can be added manually with the `-f` option: `git add -f <file>` (which is actually mentioned in the warning)

suvayu:
Very clear guidelines in [Git docs](https://git-scm.com/docs/gitignore), see the third para under "description".

Issue #294, opened 2020-01-06
Custom constraints by injecting them into the Calliope Pyomo model


timtroendle:
## Feature description

I need to create my own constraints. Because I prefer not to change Calliope’s base code, I use a non-invasive way: I manually inject Pyomo constraints into the Calliope model. Considering the lack of a plug-in system at the current moment, and considering that building such a system would likely require a lot of resources, I am wondering whether this approach or something similar should be made the official or semi-official way of adding own constraints? For me it works well, and I assume I will continue to use this approach in the future. Making this approach official or semi-official requires mainly documentation work.

This may be a duplicate of #100, feel free to close in that case.

The main code is only a few lines, and it is already mentioned briefly in #100:

```python
model = calliope.Model(path_to_model)
model.run(build_only=True)
pyomo_model = model.backend._backend
pyomo_model.my_constraint = po.Constraint(...)
# further modifications
model = model.rerun() # Works, but rerun has some issues. 
# A slightly more complicated version can avoid them.
```

Of course this requires to understand Pyomo, and it requires to understand the Pyomo model that Calliope builds. The former is obvious and I think many people in the field are able to. The latter is possible by studying the “Mathematical formulation” section of the documentation, or the NetCDF model — as it is very close to the Pyomo model. 

Exposing the Pyomo model is a disadvantage. First, because Pyomo becomes part of the official feature set. If Calliope moves away from Pyomo at one point, this feature will break. Second, because the entire Pyomo model becomes part of the Calliope interface and it needs to be maintained accordingly. Third, the user can break the Pyomo model, for example by removing the set `loc_techs` or anything similarly nasty. To overcome theses issues, the method above could be semi-official: it could be advertised in the documentation as a feasible but not officially supported approach.

Only a few things would be necessary to make this an official or semi-official approach: first, most importantly, it should be documented and advertised. Currently it's difficult to know as a user this actually works. Second, either `rerun` should be fixed or the `run` method changed in a way compatible with this approach. Third, especially if the approach is officially supported, the `Calliope.Model` needs to be slightly updated, so that this is part of the official interface and one would not need to access private members of the class.

## Calliope version

0.6.5.dev

brynpickering:
I like the idea of adding it as a stop gap for those who already know some Pyomo. 

[...]

PR #312, opened 2020-05-18
New pyomo backend interface methods

brynpickering:
Fixes issue(s) #294

* New Pyomo backend interface to add a custom constraint
* New Pyomo backend interface to view the variables, parameters, and sets in the existing backend model

Reviewer checklist:

- [x] Documentation updated
- [x] Changelog updated

### DIFF
--- a/calliope/backend/pyomo/interface.py
+++ b/calliope/backend/pyomo/interface.py
@@ -266,9 +266,7 @@ def get_all_pyomo_model_attrs(backend_model):
     }
     # Indices
     objects["Set"] = [
-        i.name
-        for i in backend_model.component_objects()
-        if isinstance(i, po.base.Set)
+        i.name for i in backend_model.component_objects() if isinstance(i, po.base.Set)
     ]
 
     return objects
--- a/calliope/backend/pyomo/interface.py
+++ b/calliope/backend/pyomo/interface.py
@@ -194,23 +194,23 @@ def add_pyomo_constraint(
 
     Examples
     --------
-    To reproduce our maximum carrier consumption constraint, you would do the following:
+    To limit supply technologies' capacity to 90% of their maximum possible capacity,
+    you would do the following:
 
-    constraint_name = 'carrier_consumption_max_constraint'
-    constraint_sets = ['loc_tech_carriers_carrier_consumption_max_constraint', 'timesteps']
+    constraint_name = 'max_capacity_90_constraint'
+    constraint_sets = ['loc_techs_supply']
 
-    def carrier_consumption_max_constraint_rule(backend_model, loc_tech_carrier, timestep):
-        loc_tech = calliope.backend.pyomo.util.get_loc_tech(loc_tech_carrier)
-        carrier_con = backend_model.carrier_con[loc_tech_carrier, timestep]
-        timestep_resolution = backend_model.timestep_resolution[timestep]
+    def max_capacity_90_constraint_rule(backend_model, loc_tech):
 
-        return carrier_con >= (
-            -1 * backend_model.energy_cap[loc_tech] * timestep_resolution
+        return backend_model.energy_cap[loc_tech] <= (
+            backend_model.energy_cap_max[loc_tech] * 0.9
         )
-    # Add the constraint
-    model.backend.add_constraint(constraint_name, constraint_sets, carrier_consumption_max_constraint_rule)
 
-    # Rerun the model with new constraint. Note: model.run(force_rerun=True) will *not* work
+    # Add the constraint
+    model.backend.add_constraint(constraint_name, constraint_sets, max_capacity_90_constraint_rule)
+
+    # Rerun the model with new constraint.
+    # Note: model.run(force_rerun=True) will *not* work, since the backend model will be rebuilt, killing any changes you've made.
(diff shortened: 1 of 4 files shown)


doc/user/advanced_constraints.rst
```diff
@@ -528,3 +528,34 @@ This constraint can be applied to storage or transmission technologies. This exa
...
+            -1 * backend_model.energy_cap[loc_tech] * timestep_resolution
+        )
+
+    # Add the constraint
+    model.backend.add_constraint(constraint_name, constraint_sets, carrier_consumption_max_constraint_rule)
+
+    # Rerun the model with new constraint. Note:
+    new_model = model.backend.rerun()
+
+.. note::
```
timtroendle, line 559: As mentioned in #294 `add_constraint` makes the xarray model and the pyomo model inconsistent. Somehow this should be said explicitly here. Especially, there should be a warning that running `model.build()` (or `model.run()` with or without `build_only`) will remove these changes (it will, right?).

calliope/test/test_backend_pyomo_interface.py
```diff
@@ -216,3 +216,118 @@ def test_rerun_fail_on_operate(self, model):
...
+            model.backend.add_constraint(
+                constraint_name, constraint_sets, energy_cap_time_varying_rule
+            )
+
+        assert check_error_or_warning(
+            excinfo, "Pyomo backend model object has no attribute 'not_a_set'"
+        )
+
+    def test_added_constraint(self, model):
+        """Constraint sets must be backend model sets"""
```
timtroendle, line 307: Wrong comment.

calliope/test/test_backend_pyomo_interface.py
```diff
@@ -216,3 +216,118 @@ def test_rerun_fail_on_operate(self, model):
...
+            loc_tech = calliope.backend.pyomo.util.get_loc_tech(loc_tech_carrier)
+            carrier_con = backend_model.carrier_con[loc_tech_carrier, timestep]
+            timestep_resolution = backend_model.timestep_resolution[timestep]
+            return carrier_con * 2 >= (
+                -1 * backend_model.energy_cap[loc_tech] * timestep_resolution
+            )
+
+        constraint_name = "new_constraint"
+        constraint_sets = [
+            "loc_tech_carriers_carrier_consumption_max_constraint",
```
timtroendle, line 319: As in the documentation, I don't really understand what this set is. Isn't there a `loc_tech_carrier` set that you could use here?

calliope/backend/pyomo/interface.py
```diff
@@ -171,6 +171,107 @@ def rerun_pyomo_model(model_data, backend_model):
...
+        i.e. arguments include the Pyomo backend model (`backend_model`)
+        and pointers to each set in `constraint_sets`.
+        Set arguments need to be in the same order as in constraint_sets.
+        The convention we follow is to name the index without the trailing `s`
+        E.g. if 'timesteps' is in constraint_sets, then 'timestep' should be one of the function arguments;
+        long `loc_techs` set names are abbreviated to `loc_tech`.
+        To see what parameters and variables you have to work with,
+        call the backend interface method `get_all_model_attrs`
+
+    Examples
```
timtroendle, line 200: Same comments as in the user documnentation, in particular I'd go for a simpler example and make crystal clear that any `build` step will kill these changes.

doc/user/advanced_constraints.rst
```diff
@@ -528,3 +528,34 @@ This constraint can be applied to storage or transmission technologies. This exa
...
+        loc_tech = calliope.backend.pyomo.util.get_loc_tech(loc_tech_carrier)
+        carrier_con = backend_model.carrier_con[loc_tech_carrier, timestep]
+        timestep_resolution = backend_model.timestep_resolution[timestep]
+
+        return carrier_con >= (
+            -1 * backend_model.energy_cap[loc_tech] * timestep_resolution
+        )
+
+    # Add the constraint
+    model.backend.add_constraint(constraint_name, constraint_sets, carrier_consumption_max_constraint_rule)
```
timtroendle, line 554: First, you need to `build` the model. This is not mentioned in this example.

doc/user/advanced_constraints.rst
brynpickering: I guess `add_constraint` doesn't cause the inconsistency, but `.rerun()`does? The xarray doesn't include the constraints themselves, only the sets needed to generate them.

calliope/test/test_backend_pyomo_interface.py
brynpickering: that is a `loc_tech_carrier` set, but I'll change it to `loc_tech_carriers_con` for simplicity (I *think* they're equivalent in this instance)

doc/user/advanced_constraints.rst
timtroendle: Probably it's correct to say that the xarray model together with the `build` code is inconsistent to the pyomo model. Both describe an optimisation problem. The former is partly declarative, partly imperative, and the second is purely declarative.

Whatever the names, I think it's important to keep this in mind for later to overcome this inconsistency eventually.

Issue #309, opened 2020-04-18
Explicitly define domain of Pyomo Params


timtroendle:
## Problem description

There are loads of Pyomo warnings in the logs of my model runs:

```
WARNING: DEPRECATED: The default domain for Param objects is 'Any'.  However,
    we will be changing that default to 'Reals' in the future.  If you really
    intend the domain of this Param (resource_unit) to be 'Any', you can
    suppress this warning by explicitly specifying 'within=Any' to the Param
    constructor.  (deprecated in 5.6.9, will be removed in 6.0)
```

I assume it's because (some?) Params in Calliope do not define a domain explicitly.

## Calliope version

0.6.5

## Pyomo version

5.6.9

brynpickering:
Yeah, we don't define the domain for any Params as far as I'm aware. We could probably just make it an easy fix with 'within=Any', but a smarter way would be to match the dtypes in the xarray dataset: 'Object' -> 'Any', 'bool' -> Binary, 'int/float' -> Reals. Could even go one step further and differentiate between `NonNeg` and `Neg` Reals.

The advantage of the smarter solution would be that on updating a Pyomo model to rerun it, we could remove the warning about us having no idea what's going on, since (I expect) Pyomo would simply complain if the dtype of the new data is incorrect.

brynpickering:
Fixed in #313

PR #313, opened 2020-05-18
Fix pyomo param domain warnings

brynpickering:
Fixes issue(s) #309 

* Pyomo `Param`objects are assigned a domain based on the variable dtype in `model_data`
* Domains now make it difficult to use None and np.nan as default values, so I've cleaned that up slightly
* Fixed an issue where updating values to `np.inf` in operate checks leads to large negative integer values if the variable`model_data` dtype is `int`. Those variables instead get assigned a float type.

Reviewer checklist:

~- [ ] Documentation updated~
- [x] Changelog updated

### DIFF
--- a/calliope/backend/checks.py
+++ b/calliope/backend/checks.py
@@ -61,6 +61,11 @@ def check_operate_params(model_data):
             return False
 
     def _set_inf_and_warn(loc_tech, var, warnings, warning_text):
+        if model_data[var].dtype == int:
+            # int -> float to have a correctly defined inf
+            _attrs = model_data[var].attrs
+            model_data[var] = model_data[var].astype(float)
+            model_data[var].attrs = _attrs
         if np.isinf(model_data[var].loc[loc_tech].item()):
             return (np.inf, warnings)
         elif model_data[var].loc[loc_tech].isnull().item():
--- a/calliope/backend/checks.py
+++ b/calliope/backend/checks.py
@@ -61,11 +61,11 @@ def check_operate_params(model_data):
             return False
 
     def _set_inf_and_warn(loc_tech, var, warnings, warning_text):
-        if model_data[var].dtype == int:
-            # int -> float to have a correctly defined inf
-            _attrs = model_data[var].attrs
-            model_data[var] = model_data[var].astype(float)
-            model_data[var].attrs = _attrs
+        # inf will only work on float dtypes
+        # astype doesn't retain attributes: https://github.com/pydata/xarray/issues/2049
+        _attrs = model_data[var].attrs
+        model_data[var] = model_data[var].astype(float)
+        model_data[var].attrs = _attrs
         if np.isinf(model_data[var].loc[loc_tech].item()):
             return (np.inf, warnings)
         elif model_data[var].loc[loc_tech].isnull().item():
--- a/calliope/backend/pyomo/model.py
+++ b/calliope/backend/pyomo/model.py
@@ -21,7 +21,7 @@ import pyomo.environ  # pylint: disable=unused-import,import-error
 # TempfileManager is required to set log directory
 from pyutilib.services import TempfileManager  # pylint: disable=import-error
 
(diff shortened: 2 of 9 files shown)


calliope/backend/checks.py
```diff
@@ -61,6 +61,11 @@ def _is_in(loc_tech, set_or_var):
             return False
 
     def _set_inf_and_warn(loc_tech, var, warnings, warning_text):
+        if model_data[var].dtype == int:
```
suvayu, line 64: I don't think this is a good idea.  I suspect it will fail when the dataset is created and used on different platforms: 32-bit vs 64-bit.  If your intention was to check for an integer in a platform agnostic manner, try the following:
```suggestion
        if model_data[var].dtype.kind == "i":
```
see: [`numpy.dtype` docs](https://numpy.org/doc/1.18/reference/generated/numpy.dtype.kind.html#numpy.dtype.kind).

That said, maybe it is just simplest to unilaterally call `astype(float)`?  But if I haven't quite grasped the larger context, please ignore this second comment.

calliope/backend/pyomo/model.py
```diff
@@ -74,42 +75,23 @@ def generate_model(model_data):
...
-                    mutable=True,
-                    default=backend_model.__calliope_defaults[k],
-                ),
-            )
+        _kwargs = {
+            "initialize": v,
+            "mutable": True,
+            "within": getattr(po, get_domain(model_data[k])),
+        }
+        if not pd.isnull(backend_model.__calliope_defaults.get(k, None)):
```
suvayu, line 83: Is `model.__calliope_default` a dictionary (or mapping)?  Then you could just do:
```suggestion
        if k in backend_model.__calliope_defaults:
```

calliope/backend/pyomo/util.py
```diff
@@ -199,3 +201,22 @@ def loc_tech_is_in(backend_model, loc_tech, model_set):
         return True
     else:
         return False
+
+
+def get_domain(var: xr.DataArray) -> str:
+    def check_sign(var):
+        if "cost" in var.name or var.name in ["resource", "loc_coordinates"]:
```
suvayu, line 208: I guess there are several kinds of `cost`s?  This is not very readable, I would simply use `re.match`.
```suggestion
        if  re.match("resource|loc_coordinates|cost*", var.name):
```
Don't forget to `import re`.

calliope/backend/pyomo/util.py
```diff
@@ -199,3 +201,22 @@ def loc_tech_is_in(backend_model, loc_tech, model_set):
...
+            return ""
+        else:
+            return "NonNegative"
+
+    if var.dtype == bool:
```
suvayu, line 213: I'm guessing this is the same as before, although probably it's safe for booleans.  Best to be consistent though:
```suggestion
    if var.dtype.kind == "b":
```

calliope/backend/checks.py
brynpickering: Yeah, possibly just setting as float without checking would be the least likely to cause issues

calliope/backend/pyomo/model.py
brynpickering: It's still possible for k to be in the dictionary and have a value of None, this is aimed at catching *both* those keys not in the dictionary and those in the dictionary but with a value of 'None' or 'NaN'

calliope/backend/pyomo/util.py
brynpickering: Yeah, I always avoided re so that I didn't have to import it; no longer sure why I had that mentality, it certainly makes the expression more readable

calliope/backend/pyomo/model.py
suvayu: Hmm, whether a config should have a `None` default is a debate for another time.  At least after I fix the `dataconfig` issues :-p

brynpickering:
Hmm, when did we lose a coverage diff check?

Issue #316, opened 2020-06-01
One way transmission and transmission costs can't co-exist


brynpickering:
## Problem description

On setting a production or consumption cost, @ramaroesilva found that one way transmission breaks down because there is no longer a valid tech for the return direction. This makes Pyomo fail when it tries setting a cost for that direction.

The same error was logged by @Stevogallo on gitter, but whether it was caused by transmission costs, I don't know.
 
## Steps to reproduce the problem

```python
override = {
            "links.X1,N1.techs.heat_pipes.constraints.one_way": True,
            "links.N1,X2.techs.heat_pipes.constraints.one_way": True,
            "links.N1,X3.techs.heat_pipes.constraints.one_way": True,
            "techs.heat_pipes.costs.monetary.om_prod": 0.1,
            "model.subset_time": "2005-01-01",
        }
m = calliope.examples.urban_scale(override_dict=override)
m.run()
```

yields:
```python
ERROR: Rule failed when generating expression for constraint
    cost_var_constraint with index ('monetary', 'N1::heat_pipes:X2',
    Timestamp('2005-01-01 00:00:00')): KeyError: "Index
    '('N1::heat_pipes:X2::heat', Timestamp('2005-01-01 00:00:00'))' is not
    valid for indexed component 'carrier_prod'"
ERROR: Constructing component 'cost_var_constraint' from data=None failed:
    KeyError: "Index '('N1::heat_pipes:X2::heat', Timestamp('2005-01-01
    00:00:00'))' is not valid for indexed component 'carrier_prod'"
```
Which comes from generating `cost_var_constraint`.

## Calliope version

0.6.6-dev

sjpfenninger:
@brynpickering Needs to check whether he's already resolved this

ramaroesilva:
You guys rock! :-)

PR #328, opened 2020-09-09
Fix one way om costs

brynpickering:
Fixes issue(s) #316 

* Catch when production costs are being applied to a transmission technology that can't produce energy
* Test

Reviewer checklist:

~~Documentation updated~~
- [x] Changelog updated

### DIFF
--- a/calliope/backend/pyomo/constraints/costs.py
+++ b/calliope/backend/pyomo/constraints/costs.py
@@ -15,7 +15,7 @@ from calliope.backend.pyomo.util import (
     get_param,
     get_timestep_weight,
     loc_tech_is_in,
-    is_valid,
+    invalid,
 )
 
 ORDER = 10  # order in which to invoke constraints relative to other constraint files
@@ -225,7 +225,7 @@ def cost_var_constraint_rule(backend_model, cost, loc_tech, timestep):
 
     if loc_tech_is_in(
         backend_model, loc_tech_carrier, "loc_tech_carriers_prod"
-    ) and is_valid(cost_om_prod):
+    ) and not invalid(cost_om_prod):
         cost_prod = (
             cost_om_prod
             * weight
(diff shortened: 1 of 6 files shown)


calliope/backend/pyomo/util.py
```diff
@@ -217,5 +217,5 @@ def check_sign(var):
         return "Any"
 
 
-def check_value(val: po.base.param._ParamData) -> bool:
-    return val._value == po.base.param._NotValid or po.value(val) is None
+def is_valid(val: po.base.param._ParamData) -> bool:
+    return not (val._value == po.base.param._NotValid or po.value(val) is None)
```
suvayu, line 221: I think renaming `is_valid` to something like `invalid` and reversing the logic would be simpler in terms of readability.  The comparison itself is to `_NotValid`, which is negated, and in `pyomo/constraints/group.py` it is negated again with `not is_valid(...)`. And in `pyomo/constraints/costs.py`, the checks remain very readable: `loc_tech_is_in(..) and not invalid(..)`.

PS: Regarding readability, my mental model for readability for conditionals is "Does this read like a natural sentence?"  On occasion you can also apply this to loops and simplify them.

calliope/test/test_backend_pyomo.py
```diff
@@ -635,6 +639,31 @@ def test_loc_techs_cost_var_constraint(self, tech, scenario, cost):
...
+        has_cost = get_indexed_constraint_body(
+            m._backend_model,
+            "cost_var_constraint",
+            ("monetary", "1::test_transmission_elec:0", m._backend_model.timesteps[1]),
+        ).to_string()
+        has_no_cost = get_indexed_constraint_body(
+            m._backend_model,
+            "cost_var_constraint",
+            ("monetary", "0::test_transmission_elec:1", m._backend_model.timesteps[1]),
+        ).to_string()
```
suvayu, line 663: I think it would be good if you could build the arguments to `get_indexed_constraint_body` in one variable, and vary it to get the second one.  In the current way, you have to look very carefully to spot where the differences are.  So something like this maybe:

```suggestion
        arg1, arg2 = m._backend_model, "cost_var_constraint",
        arg3 = ["monetary", "1::test_transmission_elec:0", m._backend_model.timesteps[1]]
        has_cost = get_indexed_constraint_body(arg1, arg2, arg3).to_string()
        
        arg3[1] =  "0::test_transmission_elec:1"
        has_no_cost = get_indexed_constraint_body(arg1, arg2, arg3).to_string()
```
This way it's very explicit.  If the third argument needs to be a tuple, you can always pass `tuple(arg3)`.

calliope/backend/pyomo/constraints/costs.py
```diff
@@ -227,13 +234,16 @@ def cost_var_constraint_rule(backend_model, cost, loc_tech, timestep):
...
-    elif loc_tech_is_in(backend_model, loc_tech, "loc_techs_supply") and cost_om_con:
+    elif loc_tech_is_in(backend_model, loc_tech, "loc_techs_supply") and is_valid(
+        cost_om_con
+    ):
         energy_eff = get_param(backend_model, "energy_eff", (loc_tech, timestep))
-        if (
-            po.value(energy_eff) > 0
-        ):  # in case energy_eff is zero, to avoid an infinite value
+        # in case energy_eff is zero, to avoid an infinite value
+        if po.value(energy_eff) > 0:
```
suvayu, line 246: I think you missed an `is_valid` check, line 254?

A clear separation between valid and invalid is probably easier to reason, so a separate `if`-clause for the validity check might be cleaner:

    if is_valid(..): # or not invalid(..)
        cost_con = 0
    else:
        if loc_tech_is_in(..):
            ...
        elif loc_tech_is_in(..):
            ...

suvayu (commented):
My comments are mostly a matter of taste.

calliope/backend/pyomo/util.py
brynpickering: makes a lot of sense, I went with it (after doing some cleanup to deal with all the failing tests).

sjpfenninger (approved):

Issue #323, opened 2020-08-27
Split loc_techs for good


timtroendle:
## Problem description

There are various issues with the use of the combined sets `loc_techs` and `loc_techs_carrier` -- in particular, it makes analysing the results more complicated. There were two reasons why this was introduced in the first place:

1) Efficiency on the storage side.
2) Efficiency within Pyomo.

The first issue is probably solvable with the [introduction of sparse support into xarray](http://xarray.pydata.org/en/stable/whats-new.html?highlight=sparse#v0-13-0-17-sep-2019). The second may be solvable with [sparse sets in Pyomo](https://pyomo.readthedocs.io/en/stable/pyomo_modeling_components/Sets.html#sparse-index-sets) (but this needs to be investigated further).

## Calliope version

0.6.5

brynpickering:
For why we did this to begin with, see #69

Issue #322, opened 2020-08-27
Link sets to supersets


timtroendle:
## Problem description

This is an internal issue. Currently, there are loads of subsets in the xarray and in Pyomo which aren't defined as such. Making the link explicit may have two benefits:

1) Ability to infer links programmatically.
2) (Maybe) performance improvements.

I am not sure on the latter, as I haven't tested this. But these subsets can easily contain 1000s of elements, and if they were linked in Pyomo, not copied, this may improve performance. 

## Calliope version

0.6.5

brynpickering:
How this would look in pyomo: `backend_model.techs_supply = po.Set(initialize=mapped_techs, domain=backend_model.techs`, where `mapped_techs` is some mapping function applied to the `techs` dim in the xarray dataset.

sjpfenninger:
No longer needed after the set cleanup

PR #346, opened 2021-01-13
Unconcatenate sets and more

brynpickering:
Fixes issue(s) #323, #322

* sets are now split out from e.g. 'loc_techs' to '(loc, techs)' to simplify the xarray dataset, remove the need for lookup tables, and pave the way for user-defined custom constraints.
* optimisation model sets for constraints, variables, and expressions are defined in a YAML file, instead of hardcoded in the core code.
* xarray model_data is now built more programmatically, with reduced hardcoding of special rules for different parameters (will be further improved with typedconfig).
* Various minor performance improvements to ensure this new functionality doesn't add to time/memory burden and, ideally, improves on v0.6.6 performance. An example is converting all datetimes to integer in the backend model, since there is a memory overhead to converting np.datetime64 to pandas datetime when going from an array dataset to a pandas series when generating the Pyomo backend model.
* 'locs' (i.e. locations) to 'nodes' to avoid all confusion with xarray/pandas slicing ('loc')

Reviewer checklist:

- [ ] Documentation updated
- [x] Changelog updated


brynpickering:
@sjpfenninger any ideas on why CI tests are not being run?

suvayu:
> any ideas on why CI tests are not being run?

Maybe because of the conflicts?

brynpickering:
> > any ideas on why CI tests are not being run?
> 
> Maybe because of the conflicts?

hmm, I'd hoped to deal with all those *after* making review edits. I'll will try dealing with them now instead.

brynpickering:
@suvayu and @timtroendle this PR sets the base layer for working from with new functionality, including typedconfig, custom constraints, updated operate mode, etc. I just realised that I didn't have the latest master merged in (i.e. it was working off a v0.6.6-dev commit), so let me fix that before anything.

There are too many little changes to expect you to review in one go, so I will point you to the main changes to review:

1. calliope/preprocess/model_data.py now builds the xarray model data more programmatically by restructuring the data dictionary, building pandas dataframes from this and then building the xarray dataset from that. We were restructuring the data dictionary before, but this is a bit slimmer, even if the code is perhaps more impenetrable...

2. calliope/backend/imasks.py & calliope/config/sets.py now combine to build the constraint and variable sets in the backend. This removes more configuration from the code (into YAML) and also allows the unconcatenation of sets (e.g. loc_techs), which makes the xarray dataset dimensions much clearer. 

3. files in calliope/backend/pyomo have been updated to expect tuple keys from imasks to build constraints and variables. The choice of which variables and constraints to build now lies solely on the programmatic results from imasks, rather than on checks undertaken in the pyomo backend, reducing the amount of code in there to just the implementation of the constraint mathematics

Known now broken elements include: 

1. operate mode. Although I've mostly fixed it locally, I want to supersede what we have with what we were working on in 2018 on windowsteps, so I'm OK with that failing at the moment.
2. group constraints. These were already implemented in a bit of a messy way, so rather than re-create that, I decided to ignore them in favour of us implementing custom constraints.
3. SPORES, which depend on a group constraint (but that was already a bit of a hack). I'll aim to create a special constraint for that using all this new imask functionality ;)
4. some preprocessing checks are no longer triggered, with the idea that typedconfig will handle those soon anyway.

calliope/test/test_model_data.py
```diff
@@ -0,0 +1,146 @@
+from re import L
```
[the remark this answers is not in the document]
brynpickering: Black added this in, no idea why...

brynpickering:
@suvayu, @timtroendle, and @sjpfenninger this is ready to go :)

calliope/test/test_backend_imask.py
```diff
@@ -0,0 +1,294 @@
...
+
+    @pytest.mark.parametrize(
+        "foreach", (["techs"], ["nodes", "techs"], ["nodes", "techs", "carriers"])
+    )
+    def test_get_valid_index(self, model_data, foreach):
+        imask = imask_foreach(model_data, foreach)
+        idx = get_valid_index(imask)
+        assert isinstance(idx, pd.Index)
+        assert len(idx) == imask.sum()
+        assert all(imask.loc[i] == True for i in idx)
```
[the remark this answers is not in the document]
sjpfenninger: this probably wants a comment explaining why the `==` comparison is used, so it's not accidentally changed in the fugure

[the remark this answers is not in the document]
brynpickering: I might just change it to a comparison to 1, but I want to make it clear that it's expecting to see a binary (but sometimes the dtype is int/float so 'is True' returns False)

timtroendle (commented):
This is a massive pull request, by far the largest I’ve ever seen. It’s in fact so large that my browser has issues displaying it. Even if I was able to understand the reasons behind all changes (which I am not), it would take me days to go through all changes in detail. Therefore, I will constrain myself to giving some high-level and general comments. In addition, I added a few line comments mainly on code style.

Here’s what I find positive on this pull request. The main reason for the pull request is great: pulling apart locs and techs. Large parts of the changes in this pull requests are code removals, which is a confirmation to me that this is a good change. The pull request changes only a small amount of tests, which should avoid regression. These are indications that the pull request can be merged.

Here’s what I think can be improved. The necessary changes of this pull request are massive. I understand that “loc_techs” are a fundamental concept of Calliope, but still think the amount of necessary changes is a warning sign that the code design is not optimal. I do not think this should be solved in this pull request, but we should keep this in mind.

In terms of readability of the changes, there is one thing I want to suggest: smaller functions. Some functions are too large, leading to very complex nesting structures and the need for line comments. The complex nesting structures are difficult to follow. Line comments are usually a warning sign for code that is not self-explanatory enough (see e.g. “Clean Code” by Robert Martin). Breaking functions into smaller parts will improve both.

When you improve readability and when the tests run through, I’ll be fine with merging this — knowing that I do not understand the details of this pull request.

calliope/backend/pyomo/util.py
```diff
@@ -221,6 +164,57 @@ def check_sign(var):
 
 def invalid(val) -> bool:
     if isinstance(val, po.base.param._ParamData):
-        return val._value == po.base.param._NotValid or po.value(val) is None
+        return val._value == po.base.param._NotValid or pd.isnull(po.value(val))
     else:
         return pd.isnull(val)
+
+
+def datetime_to_string(
```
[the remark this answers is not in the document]
brynpickering: You would think yes, but I tested string and int and string was by quicker and didn't have a noticeable memory footprint increase compared to int.

calliope/backend/checks.py
```diff
@@ -46,219 +47,166 @@ def check_operate_params(model_data):
...
         )
 
+        if ((~_is_missing("storage_cap")) & (~_is_missing("energy_cap"))).any():
+            if (
+                (
+                    model_data.storage_cap
+                    * model_data.get("energy_cap_per_storage_cap_max", np.inf)
+                )
+                < model_data.energy_cap
+            ).any():
```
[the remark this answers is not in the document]
brynpickering: I have a plan for these checks that would do away with all of this complexity. For now, they've been reformatted to ensure everything runs with 'locs' and 'techs' separated.

calliope/preprocess/model_data.py
```diff
@@ -43,468 +39,310 @@ def build_model_data(model_run, debug=False):
...
-        c
-        for c in group_constraints_carrier
-        if c
-        in "".join(model_run.model.get_key("group_share", AttrDict()).keys_nested())
-    ]:
-        group_share_data[constraint] = [
-            [
-                model_run.model.get_key(
-                    "group_share.{}.{}.{}".format(techlist, constraint, carrier), np.nan
+    model_locations = (
```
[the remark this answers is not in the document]
brynpickering: Yeah, my renaming locations to nodes didn't make its way through everything. Will fix.

brynpickering:
Thanks for the comments @timtroendle. I agree that the PR is too long, but it ended up being necessary to do a lot of overhaul to make it all work with the unconcatenation. I certainly don't expect anyone to go through this in extreme detail, hence why I gave a summary of the main things to look at earlier in the PR thread.

I will work on smaller functions for the scripts that I expect to stay. Some will end up being stripped out in future PRs, so I won't touch them for now; there is still more to be done ;)

brynpickering:
@timtroendle ignoring Hound's issues with finding flake8, would you be able to check `preprocess/model_data.py` and see if it makes more sense than the previous implementation (which included many loops etc.). I've tried to further generalise it and make everything more 'bitesized'....

timtroendle:
Any chance fixing Hound? It's terrorising my mailbox 😅😬.

brynpickering:
I can't see any reason for it to be an issue on our end, so I'm not sure
there's much we can do...

calliope/preprocess/model_data.py
```diff
@@ -9,17 +9,21 @@
...
+from calliope import exceptions
 from calliope.core.attrdict import AttrDict
 from calliope._version import __version__
 from calliope.preprocess import checks
-from calliope.preprocess.time import add_time_dimension, update_dtypes
+from calliope.preprocess import time
+from calliope.core.util import dataset
 
 
 def build_model_data(model_run, debug=False):
```
timtroendle, line 29: I like the new class-based structure. To be consistent, I'd kill this function too by moving its functionality into `ModelData.init` . The entire interface to the module (and class) from outside the module would then be:

```Python
from calliope.preprocess.model_data import ModelData

model_data = ModelData(model_run)
model_data_original = model_data.model_data_original
model_data = model_data.model_data # FIX name conflict
if debug:
    data_pre_time = model_data.data_pre_time
```

This interface shows, however, the name `model_data` is slightly overstretched. Anyways, I would not worry to much about that at this point.

calliope/preprocess/model_data.py
```diff
@@ -51,325 +55,398 @@ def build_model_data(model_run, debug=False):
...
+            "essentials.carrier",
+        ]
+
+        self.strip_unwanted_keys()
+        self.add_node_tech_sets()
+
+        if model_run_dict.get_key("model.random_seed", None):
+            np.random.seed(seed=model_run_dict.model.random_seed)
+
+    def empty_or_invalid(self, var):
```
timtroendle, line 119: I'd mark all methods that are supposed to be private with an underscore: `def _empty_or_invalid(self, var)`.

In fact, `ModelData` has no public methods apart from the default methods of classes in Python.

timtroendle (commented):
I only gave this a high-level check as there are many new changes. The new structure makes sense to me as it decreases function size and makes the purpose of the module clearer: creating a module_data. With a few more tweaks you can make it even more consistent with the factory pattern.

calliope/backend/imasks.py
```diff
@@ -0,0 +1,228 @@
...
+import functools
+import re
+import ast
+
+import xarray as xr
+import pandas as pd
+from calliope.core.attrdict import AttrDict
+from calliope.core.util.dataset import reorganise_xarray_dimensions
+
+# [storage_cap_min, or, inheritance(storage), and, export=True, and, run.mode=plan]
```
sjpfenninger, line 23: Leftover?

calliope/config/model_data_lookup.yaml
```diff
@@ -0,0 +1,79 @@
+tech_inheritance:
```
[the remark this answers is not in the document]
brynpickering: Sadly, this isn't going to be dealt with by typedconfig. The exact regular expressions will change with typedconfig, but we will still have this issue of having a heavily nested dictionary that needs to be converted to tabular data somehow.

sjpfenninger:
What _is_ wrong with hound?

calliope/backend/imasks.py
```diff
@@ -0,0 +1,226 @@
+"""
+Copyright (C) since 2013 Calliope contributors listed in AUTHORS.
+Licensed under the Apache 2.0 License (see LICENSE file).
+
+masks.py
```
sjpfenninger, line 5: i

calliope/backend/pyomo/constraints/__init__.py
```diff
@@ -0,0 +1,12 @@
+from calliope.backend.pyomo.constraints.capacity import *
```
sjpfenninger, line 1: Maybe we should put all constraint functions into one file so these import statements aren't needed -- these files no longer work as one would expect Python modules to work, since all their contents are dumped into one namespace here anyway. @timtroendle? @brynpickering? And perhaps this should be called "rules" rather than "constraints", since it also contains rules for expressions.

calliope/core/model.py
```diff
@@ -92,47 +86,35 @@ def __init__(self, config, model_data=None, *args, **kwargs):
...
-        self._debug_data = debug_data
         log_time(
             logger,
             self._timings,
             "model_run_creation",
             comment="Model: preprocessing stage 1 (model_run)",
         )
 
-        self._model_data_original = build_model_data(model_run)
+        model_data_factory = ModelDataFactory(model_run)
```
sjpfenninger, line 98: Why do we call several methods on ModelDataFactory here? It seems like those should be called in ModelDataFactory.__init__ instead so that instantiating a model_data_factory gives you all the different data you need in one go, without the need to know about the internal methods of the class.

sjpfenninger (changes requested):

calliope/core/model.py
brynpickering: It's split out for what I assume were the same reasons for splitting out the parts of the process previously: testing. If initialising the ModelDataFactory does *everything* in one go, then it's a pain to test some of the key intermediate steps in the process.

calliope/backend/pyomo/constraints/__init__.py
brynpickering: It's my usual refrain, but I wouldn't concern ourselves with these details for now: with custom constraints implemented we'll be able to re-write these constraints in YAML and can do away with all these files anyway.

sjpfenninger: Ok, let's leave it be for now

timtroendle: Agree with Bryn that it might not be the right moment to do so.

If we were to keep these files for a longer time I'd suggest some sort of a callback mechanism that inverts the current call order. There could be a registry of whatever form in this file here (the simplest being a Python list) into which constraints (or constraint module) add themselves: "Hello, I am a constraint. Call me back whenever you needs constraints."

calliope/core/model.py
sjpfenninger: So move what's in `__init__()` right now into a separate `_initial_init()` or something, then have `__init__()` call the four init functions in turn, but still test them separately?

timtroendle: I agree with Stefan here. This approach seems error prone, because you can easily forget to call one of these methods.

If it's difficult to test that's usually a sign for a design flaw. As fixing the design may be too much here I'd go for a pragmatic, even if slightly ugly, way: call "private" methods in the test. In that case, you'd make all these methods private and call them from within a creation method (could be `build()`, or `create`, but I suggest `__call__`). Then you can test subsets of that creation method in your test by calling private methods instead of the creation method itself. 

The API then looks like:
```Python
model_data_factory = ModelDataFactory(model_run)
model_data = ModelDataFactory()
```

brynpickering: I'm in favour of this change. However, calling ModelDataFactory would need to return four objects, not just `model_data`, which should be reserved as a pointer to the main xarray dataset. So it would be: 

```
model_data_factory = ModelDataFactory(model_run)
model_data_pre_clustering, model_data, data_pre_time, stripped_keys = ModelDataFactory()
```

Is that also OK?

timtroendle: I think that's fine.

brynpickering: OK, I've made that change, although I'm liking ModelDataFactory less and less at this point :P

sjpfenninger (dismissed):

brynpickering:
@sjpfenninger and @timtroendle thoughts on squashing essentials/constraints/switches into one in this commit? I added `switches` as a new component to try and differentiate between config which is optional but is "one of a set" ([True, False], ["energy", "energy_per_cap", "energy_per_area"], etc.) and that which is completely free for a user to define. 

I'm not a huge fan of it and wonder whether it should be sorted out now before it gets accidentally baked in by merging this commit...

changelog.rst
```diff
@@ -3,8 +3,30 @@
...
+
+|changed| |backwards incompatible| The `loc::tech` and `loc::tech::carrier` sets have been removed. Model components are now indexed separately over `node`, `tech`, and `carrier` (where applicable). Although primarily an internal change, this affects the xarray dataset structure and hence how users access data in `model.inputs` and `model.results`. For example, `model.inputs.energy_cap_max.loc[{"loc_techs": "X:pv"}]` in v0.6 needs to be changed to `model.inputs.energy_cap_max.loc[{"nodes": "X", "techs": "pv"}]` in v0.7. This is functionally equivalent to first calling `model.get_formatted_array("energy_cap_max")` in v0.6, which is no longer necessary in v0.7.
+
+|changed| |backwards incompatible| The dimensions of the model data no longer include all possible subsets. E.g. a user can no longer access `loc_techs_supply` to view the location/technology pairs which have defined `supply` as their top-level parent. Instead, the same subset can be supplied by calling `model.inputs.inheritance.str.endswith('supply')` to create a boolean array of technologies with `supply` as their top-level parent.
```
sjpfenninger, line 18: Should there be a more complete code example here (as in the above?)

calliope/test/test_model_data.py
```diff
@@ -0,0 +1,425 @@
...
+    ):
+        nesting = ["techs"]
+        key_to_check = "aA.techs.bB"
+        search_results = model_data._get_key_matching_nesting(
+            nesting, key_to_check, start, end
+        )
+        if search_result is not None:
+            assert search_results.groups() == tuple(search_result)
+        else:
+            assert search_results == None
```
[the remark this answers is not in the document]
sjpfenninger: Is this else clause needed here?

changelog.rst
brynpickering: No, I think this is fine since it's really a starting point for us to be able to update the documentation, right?

PR #353, opened 2021-04-01
Update mechanism for dataset checking; get operate mode working again

brynpickering:

* As with everything else, checks applied directly to the model_data xarray dataset are now given in YAML configs. This allows us to capitalise on already implemented functionality in `subsets` and removes complex conditionals currently found in the dataset checks.
* This has been applied to operate mode checks and the few `model_data` checks.
* Operate mode tests now pass, although operate mode would be better updated to using `windowsteps`, as was tested in another branch a few years ago.

If the functionality seems to make sense (are we going too far with YAML configs....?), then I'll add tests for `checks.py` and move some of these scripts around so they are in more sensible places based on when they get used.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


calliope/postprocess/results.py
```diff
@@ -91,10 +91,15 @@ def capacity_factor(results, model_data, systemwide=False):
...
     # In operate mode, energy_cap is an input parameter
-    if "energy_cap" not in results.keys():
-        energy_cap = model_data.energy_cap
-    else:
+    try:
         energy_cap = results.energy_cap
+    except AttributeError:
+        try:
+            energy_cap = model_data.energy_cap
+        except AttributeError:
```
sjpfenninger, line 99: Is this a smart idea?

calliope/backend/checks.py
```diff
@@ -8,14 +8,43 @@
...
+
 import numpy as np
-import pandas as pd
 import xarray as xr
 from calliope.core.attrdict import AttrDict
-from calliope.core.util.observed_dict import UpdateObserverDict
+from calliope.backend.subsets import imask_where, _combine_imasks, _imask_foreach
+
+
+def _equation_parser(model_data, eq_dict):
```
sjpfenninger, line 20: Should this already go somewhere else assuming it will be used for custom constraints too?

calliope/backend/checks.py
```diff
@@ -32,202 +61,56 @@ def check_operate_params(model_data):
...
+                for exctype, exclist in logs.items():
+                    if exctype in check_config.keys():
+                        exclist.append(check_config[exctype])
+                if "must_exist" in check_config.keys():
+                    _create_var(check_config["must_exist"], model_data, run_config)
+
+    return logs["warning"], logs["error"]
+
+
+def _create_var(var_to_create, model_data, run_config):
```
sjpfenninger, line 83: For functions like this that look pretty impenetrable I think we should take the minimal effort of writing a docstring that says what the function's parameters are and what it returns

calliope/backend/run.py
```diff
@@ -500,14 +515,15 @@ def run_operate(model_data, timings, backend, build_only):
...
+            # Note: Warmstart isn't possible with GLPK (dealt with here to avoid the warning being duplicated)
+
             _results = backend.solve_model(
                 backend_model,
                 solver=solver,
                 solver_io=solver_io,
                 solver_options=solver_options,
                 save_logs=save_logs,
-                warmstart=warmstart,
+                warmstart=warmstart * allow_warmstart,
```
sjpfenninger, line 526: Is this actually needed? i.e. will it error on a solver that cannot warmstart, if warmstart=True?

calliope/backend/run.py
```diff
@@ -525,6 +541,7 @@ def run_operate(model_data, timings, backend, build_only):
                 _results = backend.get_result_array(backend_model, model_data)
             else:
                 _results = xr.Dataset()
+                continue
```
sjpfenninger, line 544: is this a good idea? it wasn't there before and if we need it now, maybe the logic should be rewritten so it's clearer what's happening?

calliope/backend/subsets.py
```diff
@@ -136,7 +142,7 @@ def _inheritance(model_data, tech_group):
 }
 
 
-def _imask_where(
+def imask_where(
```
sjpfenninger, line 145: Again I think a minimal docstring would be very helpful

calliope/backend/subsets.py
```diff
@@ -196,14 +202,29 @@ def __is_equals_statement(imask_string):
     return imask
 
 
-def _combine_imasks(curr_imask, new_imask, _operator):
+def _translate_operator(_operator):
```
sjpfenninger, line 205: This will be a lot [nicer in Python 3.10](https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching)

calliope/backend/subsets.py
```diff
@@ -196,14 +202,29 @@ def __is_equals_statement(imask_string):
...
     if _operator in ["or", "or_"]:
```
sjpfenninger, line 206: this should just be `_operator == "or"`

calliope/backend/subsets.py
```diff
@@ -196,14 +202,29 @@ def __is_equals_statement(imask_string):
...
         _operator = "or_"
     elif _operator in ["and", "and_"]:
```
sjpfenninger, line 208: this should just be `_operator == "and"`

sjpfenninger (changes requested):

sjpfenninger:
Think generally this makes sense and we should proceed down this route of YAMLification!

calliope/backend/run.py
brynpickering: Yes, it will, annoyingly.

calliope/postprocess/results.py
brynpickering: The relative benefit of if/else vs try/except seems to be hotly debated on forums; I would be happy enough using if/else if you think it's smarter. TBH, our operate mode stuff *isn't* very smart, as evidenced by things like this.

brynpickering:
@sjpfenninger: plan of action on this is to clean up Hound's concerns and then merge in. Could you double check that I'm not missing anything major?

brynpickering:
Closing in favour of #539

PR #363, opened 2021-06-03
Fix conda environment creation statement

FLomb:
I think it should be `conda create` instead of `conda env create`; or at least, in my version of conda that's how it works. It doesn't work if I try to use the statement suggested in the documentation.

Fixes issue(s) #

* just edited the documentation about how to install a developer version

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/doc/user/develop.rst
+++ b/doc/user/develop.rst
@@ -30,7 +30,7 @@ Then install all development requirements for Calliope into a new environment, c
 
   .. code-block:: fishshell
 
-   $ conda env create -f requirements.yml -n calliope_dev
+   $ conda create -f requirements.yml -n calliope_dev
    $ conda activate calliope_dev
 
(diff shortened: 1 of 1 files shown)


brynpickering:
The documentation is correct, including for current `conda` and `mamba` versions. What versions are you using to come across this?

FLomb:
@brynpickering what do you mean? as far as I can see, the master branch still has the wrong documentation (https://github.com/calliope-project/calliope/blob/master/doc/user/develop.rst) and that's also what is shown on the website for version 0.6.6-post1. In addition, I don't see any reference to `mamba`. Which version are you looking at?

FLomb:
And you're right, it works on my machine too now. Weird, who knows what was going wrong back then. I'm closing the PR.

PR #375, opened 2022-01-14
version 0.6.8 

brynpickering:

* Fix "demand share per decision timestep" constraint set formulation.
* Add run config relaxation option for "demand share per decision timestep" (strict `equals` can easily lead to infeasibility).
* Change default values from False to None to avoid issues of booleans and zeros being mixed up.
* Catch and deal with issue in saving to NetCDF, whereby some variables end up with mixed dtypes, where one dtype is boolean. Booleans are changed to floats.
* Add more logging steps to logging level `INFO`

Reviewer checklist:

- [x] Documentation updated
- [x] Changelog updated


changelog.rst
```diff
@@ -3,6 +3,23 @@
...
+
+|new| `storage_cap_min/equals/max` group constraints added.
+
+|changed| Updated to Pyomo 6.2, pandas 1.3, xarray 0.20, numpy 1.20.
+
+|changed| parameters defaulting to False now default to None, to avoid confusion with zero.
+
+|changed| `INFO` logging level includes logs for dataset cleaning steps before saving to NetCDF and for instantiation of timeseries clustering/resampling (if taking place).
+
+|fixed| `demand_share_per_timestep_decision` constraint set includes.
```
sjpfenninger, line 18: "included" ?

calliope/test/common/test_model/storage_cap.yaml
```diff
@@ -0,0 +1,106 @@
+
```
sjpfenninger, line 1: Call this something like `group_constraints_storage_cap.yml` so its clearer what it tests?

calliope/test/common/test_model/storage_cap.yaml
```diff
@@ -0,0 +1,106 @@
...
+            techs: ["cheap_elec_storage"]
+    cap_expensive_storage:
+        group_constraints.example_storage_cap_constraint:
+            techs: ["expensive_elec_storage"]
+
+scenarios:
+    storage_cap_max: [cap_max, cap_cheap_storage]
+    storage_cap_min:  [cap_min, cap_expensive_storage]
+    storage_cap_equals:  [cap_equals, cap_expensive_storage]
+    storage_cap_min_max: [cap_max_cheap_storage, cap_min]
```
sjpfenninger, line 106: This last scenario adds two constraints, one of which does not specify any techs. Intended?

I do also wonder whether for models like this there shouldn't be a little bit of documentation saying "with these constraints we are trying to achieve behaviour X, ..." - so that they can be used for the fabled modelling strategies guide?

calliope/test/test_example_models.py
```diff
@@ -76,8 +76,8 @@ def example_tester(self, solver="cbc", solver_io=None):
     def test_nationalscale_example_results_cbc(self):
         self.example_tester()
 
-    @pytest.importorskip("gurobipy")
     def test_nationalscale_example_results_gurobi(self):
+        pytest.importorskip("gurobipy")
```
sjpfenninger, line 80: Why? gurobipy messing with things when you want to use the LP/file interface?

changelog.rst
```diff
@@ -3,6 +3,23 @@
 Release History
 ===============
 
+0.6.8 (dev)
+------------------
+|new| run configuration parameter to enable relaxation of the `demand_share_per_timestep_decision` constraint.
```
sjpfenninger, line 8: Documentation?

calliope/preprocess/checks.py
```diff
@@ -650,35 +650,6 @@ def check_final(model_run):
                 "Time series `{}` contains non-unique timestamp values.".format(k)
             )
 
-    # Warn if loc/tech is defined in group constraint that doesn't exist in the model
```
sjpfenninger, line 653: Why gone?

calliope/backend/pyomo/constraints/energy_balance.py
```diff
@@ -211,7 +211,7 @@ def balance_supply_constraint_rule(backend_model, loc_tech, timestep):
     else:
         available_resource = resource * resource_scale
 
-    if po.value(force_resource):
+    if po.value(force_resource) == 1:
```
sjpfenninger, line 214: Should we comment in cases like this that really this is boolean converted to 0/1? It is not very pythonic...

sjpfenninger (changes requested):
Looks good in general! Some comments, mostly minor or for clarification. Documentation.... have we given up on that?

calliope/config/defaults.yaml
```diff
@@ -29,6 +29,8 @@ run:
         save_per_spore: false  # whether or not to save each SPORE run results separately or as one concatenated NetCDF. If True, "save_per_spore_path" or CLI argument "--to_netcdf" must be defined (to_netcdf will take precendence and be used a the directory name).
         save_per_spore_path: null  # file path for each spore run, which will be used if save_per_spore is used, and CLI command "--to_netcdf" is not defined. Will apply spore number using the python "format" method, so the path should include a "{}" at the position where the spore number should be included (e.g. "/path/to/spores/spore_{}.nc" will save results for SPORE 1 to "/path/to/spores/spore_1.nc"). The cost-optimal solution will be saved by using spore number 0.
         skip_cost_op: false  # whether or not to run the initial cost optimisation model to ascertain the cost-optimal cost and initial spores scores. If True, will take the group constraint and cost_energy_cap values directly.
+    relax_constraint:  # Enable relaxing some equality constraints to be min/max constraints. The extent of relaxation is given as a fraction.
```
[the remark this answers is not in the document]
brynpickering: Yes, this is really a function of necessity; we have the constraint and so we should make sure it can work. I actually don't like having to put this new key in. 

Another option would be to strip out this troublesome constraint entirely and I would just implement it as a custom constraint when I use it (namely, in Euro-Calliope).

calliope/preprocess/constraint_sets.py
```diff
@@ -437,15 +437,10 @@ def generate_constraint_sets(model_run):
         techs = group_constraint.get("techs", allowed_techs)
 
         locs = group_constraint.get("locs", sets["locs"])
-
```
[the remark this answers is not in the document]
brynpickering: they're stripped along with all other non-existent loc::techs later on

calliope/test/common/test_model/storage_cap.yaml
brynpickering: intended. It sets the minimum globally across all storage techs and the maximum only on the cheap tech.

I would hope the modelling strategies guide would come after we deprecate group constraints ;)

calliope/test/test_example_models.py
brynpickering: It tries to import gurobipy, if it fails, it skips the whole test. This is for use on any machine (including the CI) where gurobi isn't available.

changelog.rst
brynpickering: Looks like I forgot to finish the sentence...

calliope/backend/pyomo/constraints/energy_balance.py
brynpickering: It isn't, but is an issue with saving and loading the data to/from netcdf. In some instances, boolean data needs to be stored as floats, hence this (and some other) non-pythonic check...

calliope/preprocess/checks.py
brynpickering: Equivalent warnings come up in 'constraint_sets'

calliope/preprocess/util.py
```diff
@@ -152,7 +152,7 @@ def vincenty(coord1, coord2):
         cosSigma = sinU1 * sinU2 + cosU1 * cosU2 * cosLambda
         sigma = np.arctan2(sinSigma, cosSigma)
         sinAlpha = cosU1 * cosU2 * sinLambda / sinSigma
-        cosSqAlpha = 1 - sinAlpha ** 2
+        cosSqAlpha = 1 - sinAlpha**2
```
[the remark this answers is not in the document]
brynpickering: In fact, Black (v22.1.0) made these exact changes

sjpfenninger (approved):

PR #390, opened 2023-01-10
Add non-test relevant files to exclude path list

brynpickering:
This extends the triggers for PRs and branch-based CI to minimise the amount of testing being done. Only changes to those files which are part of the python package (and the few top-level test config files) should cause the test pipeline to run.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,9 +1,37 @@
 trigger:
-  - master
-  - "*"
+  branches:
+    include:
+      - master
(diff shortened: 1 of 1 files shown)


brynpickering:
OK, I can't be certain that this actually works since there seems to be the issue that if CI is triggered on the first commit of a PR then all subsequent commits will trigger CI. Will need to test this with dummy PRs...

PR #393, opened 2023-01-10
Update and align click between requirement files

brynpickering:
Fixes issue of `click` being pinned in `requirements.txt` and not in `requirements.yml`.


brynpickering:
I should add that updating click adds no features but it aligns with the requirements for the last 3-4 major releases of Black.

PR #396, opened 2023-01-11
Update requirements & fix code to handle them

brynpickering:
Fixes issue of installing v0.6.x on Apple Silicon machines. It also (locally) is blissfully warning-free.

Azure pipelines CI doesn't have an Apple Silicon VM available (yet), so this will need to be tested by those with new macbooks to see if it works. @sjpfenninger @timtroendle.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


brynpickering:
We will have to sacrifice py3.7 to make it work on Apple Silicon devices. We can't go to py3.10 because of some issue between the version of importlib.metadata we would require and the version of xarray we're forced to pin to (2022.03, due to a breaking change they make after that - not sure yet if it's intentional - which makes it impossible to reorder dimensions in a dataset/array).

sjpfenninger (approved):
This works on macos/arm. Good to go!

ollie-bell:
Would be great to get this is in as currently installing calliope on Windows doesn't work on Python 3.9, 3.10 or 3.11. scikit-learn 0.23.2 doesn't have a wheel on pypi for these versions of Python, neither does numpy 1.17.3. It ends up trying to build numpy from source.

brynpickering:
@ollie-bell thanks for the info. I've already removed the troubleshooting option to install calliope from pip - everything should come from `conda-forge`, even if installing a development environment.

Issue #360, opened 2021-05-20
Rewrite Pyomo constraints in custom YAML format


sjpfenninger:
## Problem description

Once #100 is complete, we want to rewrite all built-in constraints to use the custom YAML format.

Issue #100, opened 2018-04-16
Custom constraints


brynpickering:
In 0.6.0 we have (temporarily?) removed the ability to introduce custom constraints via the YAML format*. 

To bring it back in, ideally a user could have one Python script which introduces any new sets, parameters, decision variables, and constraints. If we provide a template for this, it would ensure that a user includes all the necessary information to produce the desired results.

## Steps to introduce the functionality

- [ ] Get list of all data required to build a brand new constraint
- [ ] Build a custom constraint template and test with a new constraint

* A user familiar with Pyomo can currently add a custom constraint after building the backend model (``model.run(build_only=True)``) by accessing the Pyomo ConcreteModel directly via ``model._backend_model``.

sjpfenninger:
Always sets for each `tech` and `tech_group` to allow you to easily attach per-tech custom constraints via those sets?

timtroendle:
Here's a work-in-progress suggestion of how custom constraints could look like in the yaml file. The idea is that next to the user-friendly Calliope structure, one can manually add ´Sets´, `Parameters`, `Variables`, and `Constraints`.

The constraints defined below are the group constraints in Calliope 0.6.5. It requires #322 to be solved first.

``` yaml
custom_constraints:
    sets:
        wind_techs:
            elements:
                - onshore_wind
                - offshore_wind
            within: techs
        ac_transmission:
            elements:
                - ac_transmission
            within: techs
        nordic_countries:
            elements:
                - FIN
                - CHE
                - NOR
            within: locs
        tims_favorite_timesteps:
            elements:
                - "2016-04-01 04:00"
                - "2016-06-16 08:00"
            within: timesteps
        monetary_cost:
            elements:
                - monetary
            within: costs
    parameters:
        carrier_prod_share: 0.25
        net_import_share: 0.7
        energy_cap_share: 0.2
        energy_cap: 200
        cost_cap: 1e6
        cost_var_cap: 1e4
        cost_investment_cap: 1e3
        resource_area_cap: 100
    # TODO add variables
    constraints:
        net_import_share:
            foreach: [nordic_locs, carriers]
            eq: net_import_sum <= net_import_share * demand_sum
            components:
                net_import_sum:
                    sum: carrier_prod + carrier_con
                    over: [ac_transmission, timesteps]
                demand_sum:
                    sum: carrier_con
                    over: [demand, timesteps]

        carrier_prod_share:
[...]

PR #409, opened 2023-03-06
Feature yaml constraints evaluate expressions

brynpickering:
Partial fix for issue(s) https://github.com/calliope-project/calliope/issues/360, https://github.com/calliope-project/calliope/issues/100 to be able to process multidimensional array masking for our new constraints (overall plan detailed in https://github.com/calliope-project/calliope/issues/100):

* Evaluate parsed where strings into masking arrays within the parsing class.
* Evaluate parsed equation expressions into backend objects that can be compared in an equation. NOTE: interface to access and operate on backend objects is currently an empty dummy class.
* Each parsed element has a distinct "as_dict" evaluation method, to enable testing the evaluation without relying on the backend interface (previously "test" and only applied to a few elements).
* Parsing class is now an abstract base class inherited by different optimisation problem parse classes (e.g., to parse constraint dicts, variable dicts, objective dicts, etc.)

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


brynpickering:
> Some files could not be reviewed due to errors:
> ./calliope/backend/equation_parser.py:719:30: B008: Do not perform function c...

Fixed, but Hound does not hide the warning after it has issued it.

PR #414, opened 2023-03-14
Feature: Allow to override constraint.yml with custom constraints

jnnr:
Fixes issue(s) #100 

* Moves the storage-inter-cluster to separate yaml file
*

Reviewer checklist:

- [ ] Documentation updated
- [x] Changelog updated


calliope/config/storage-inter-cluster.yaml
```diff
@@ -1,7 +1,95 @@
...
+          expression: storage[timesteps=$final_step]
+    index_slices:
+      final_step:
+        - expression: roll(lookup_datestep_last_cluster_timestep, datesteps=1)
+
 variables:
   storage:
     foreach: [nodes, techs, timesteps]
-    where: "(inheritance(storage) OR inheritance(supply_plus)) AND include_storage=True AND model.time.function_options.storage_inter_cluster=True AND model.time.function=apply_clustering"
+    where: "(inheritance(storage) OR inheritance(supply_plus)) AND include_storage=True AND model.time.function=apply_clustering"
```
brynpickering, line 71: I think `AND model.time.function=apply_clustering` should not be there (my fault for including it to begin with). If it were to stay then this variable would *only* exist if the user was clustering the timeseries.

Saying that, without this conditional, if a user doesn't use clustering in the end, for whatever reason, the "storage" variable will exist but with a lower bound of -inf. This will cause other issues in a model without clustering...

calliope/config/storage-inter-cluster.yaml
```diff
@@ -1,7 +1,95 @@
+constraints:
+  balance_supply_plus_with_storage:
+    components:
+      carrier_prod: *carrier_prod_with_parasitic
```
brynpickering, line 4: Unnecessary duplication between the YAMLS

calliope/config/storage-inter-cluster.yaml
```diff
@@ -1,7 +1,95 @@
+constraints:
```
brynpickering, line 1: Missing switching off (active: false) of the "storage_max" constraint

calliope/config/storage-inter-cluster.yaml
```diff
@@ -1,7 +1,95 @@
...
+      carrier_prod: *carrier_prod_with_parasitic
+      storage_previous_step: &storage_previous_step
+        - where: timesteps=get_val_at_index(dim=timesteps, idx=0) AND NOT run.cyclic_storage=True
+          expression: storage_initial * storage_cap
+        - where: ((timesteps=get_val_at_index(dim=timesteps, idx=0) AND run.cyclic_storage=True) OR NOT timesteps=get_val_at_index(dim=timesteps, idx=0)) AND NOT lookup_cluster_last_timestep
+          expression: (1 - storage_loss) ** roll(timestep_resolution, timesteps=1) * roll(storage, timesteps=1)
+        - where: lookup_cluster_last_timestep
+          expression: "0"
+
+  set_storage_initial_inter_cluster:
```
brynpickering, line 13: This should have the name `set_storage_initial` so that it replaces the existing `set_storage_initial` constraint instead of existing alongside it.

calliope/config/storage-inter-cluster.yaml
brynpickering: If it is removed, then this matches the where string in the original and can be removed here.

calliope/config/storage-inter-cluster.yaml
```diff
@@ -1,7 +1,95 @@
...
+        - expression: roll(lookup_datestep_last_cluster_timestep, datesteps=1)
+
 variables:
   storage:
     foreach: [nodes, techs, timesteps]
-    where: "(inheritance(storage) OR inheritance(supply_plus)) AND include_storage=True AND model.time.function_options.storage_inter_cluster=True AND model.time.function=apply_clustering"
+    where: "(inheritance(storage) OR inheritance(supply_plus)) AND include_storage=True AND model.time.function=apply_clustering"
+    bounds:
+      min: -.inf
+      max: .inf
```
brynpickering, line 74: Unnecessary duplication between the YAMLS

calliope/config/storage-inter-cluster.yaml
```diff
@@ -1,7 +1,95 @@
...
+          expression: "0"
+        - where: NOT (datesteps=get_val_at_index(dim=datesteps, idx=0) AND NOT run.cyclic_storage=True)
+          expression: storage[timesteps=$final_step]
+    index_slices:
+      final_step:
+        - expression: roll(lookup_datestep_last_cluster_timestep, datesteps=1)
+
 variables:
   storage:
     foreach: [nodes, techs, timesteps]
```
brynpickering, line 70: Unnecessary duplication between the YAMLS

calliope/config/storage-inter-cluster.yaml
brynpickering: and then foreach and where are unnecessary duplications

brynpickering (changes requested):
Thanks for the PR @jnnr. Some updates needed, mostly to avoid unnecessary duplication between the YAMLs.

calliope/config/constraints.yaml
brynpickering, line 1: Should this file be called something like "base_optimisation_config.yaml"?

calliope/config/defaults.yaml
```diff
@@ -46,7 +46,8 @@ model:
     timeseries_data: null  # Dict of dataframes with time series data (when passing in dicts rather than YAML files to Model constructor)
     timeseries_dateformat: '%Y-%m-%d %H:%M:%S'  # Timestamp format of all time series data when read from file
     file_allowed: ['clustering_func', 'energy_eff', 'energy_ramping', 'export', 'om_con', 'om_prod', 'parasitic_eff', 'resource', 'resource_eff', 'storage_loss', 'carrier_ratios']  # List of configuration options allowed to specify "file=" to load timeseries data. This can be updated if you're adding a new custom constraint that requires a newly defined parameter to be a timeseries. If updating existing parameters, you can expect existing constraints to not change behaviour or to break on being constructed.
-
+    optimisation_config_overrides: []  # List of custom-constraints
```
brynpickering, line 49:
```suggestion
    optimisation_config_overrides: []  # List of references to optimisation problem overrides. References can be made to in-built overrides (without a `.yaml` suffix) and to user-defined override YAML files (at paths relative to the loaded `model.yaml` file)
```

brynpickering (changes requested):
I've left a few queries on the naming of internal variables / model attributes. I would say we move away from using "custom constraints" and focus more on custom optimisation configuration (since a user could also/instead provide custom variables/objectives/expressions). The exact name is up for discussion. Any preferences @jnnr @FraSanvit @sjpfenninger ?

calliope/config/defaults.yaml
sjpfenninger: Call this something other than `overrides` since overrides already exist elsewhere? Though in a sense this is the same because underneath it's also munging together dicts. Still - something like `custom_config`? `optimisation` seems somewhat superfluous too since we're in an optimisation modelling framework anyway?

calliope/core/model.py
```diff
@@ -148,7 +151,10 @@ def _init_from_model_run(self, model_run, debug_data, debug):
                 os.path.dirname(calliope.__file__), "config", "constraints.yaml"
             )
         )
-        self._add_observed_dict("component_config", component_config)
+        component_config_override = self._add_optimisation_config_override(
```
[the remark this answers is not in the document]
sjpfenninger: Neither seem particularly self-explanatory but I can't think of anything smarter either besides just going more minimal and calling it just `config` 🤷 (though isn't it really data rather than config ;)?

calliope/config/defaults.yaml
brynpickering: agreed offline: `custom_math`

calliope/core/model.py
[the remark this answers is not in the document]
brynpickering: agreed offline: `math`

calliope/config/storage_inter_cluster.yaml
```diff
@@ -0,0 +1,92 @@
+constraints:
+  storage_max:
+    active: False
```
brynpickering, line 3: should this be "active" or "enabled"/"exists"/...?

This could cause confusion where "active" means something different in the optimisation world (i.e., whether a constraint was ultimately bounding or not).

brynpickering (approved):

PR #417, opened 2023-03-17
Raise error when equations or expressions contain NaNs

jnnr:
Fixes issue(s) #

* Catches errors arising because of lhs or rhs containing NaN or possible expressions being NaN, but the imask defined by where is True. This is in improvement compared to hard-to-understand errors down the line when pyomo objects are created.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/calliope/test/test_backend_pyomo.py
+++ b/calliope/test/test_backend_pyomo.py
@@ -2127,8 +2127,8 @@ class TestNewBackend:
         # add constraint with nan
         constraint_dict = {
             "foreach": ["techs", "carriers"],
-            "equation": "sum(carrier_prod, over=[nodes, timesteps]) >= 100"
-            # "where": "carrier_prod"  # <- no error would be raised with this uncommented
+            "equation": "sum(carrier_prod, over=[nodes, timesteps]) >= 100",
+            # "where": "carrier AND allowed_carrier_prod=True AND [out, out_2, out_3] in carrier_tiers",  # <- no error would be raised with this uncommented
         }
         constraint_name = "constraint-with-nan"
 
@@ -2172,8 +2172,8 @@ class TestNewBackend:
 
         expression_dict = {
             "foreach": ["techs", "carriers"],
-            "equation": "sum(carrier_prod, over=[nodes, timesteps])"
-            # "where": "carrier_prod"  # <- no error would be raised with this uncommented
+            "equation": "sum(carrier_prod, over=[nodes, timesteps])",
+            # "where": "carrier AND allowed_carrier_prod=True AND [out, out_2, out_3] in carrier_tiers",  # <- no error would be raised with this uncommented
         }
         expression_name = "expression-with-nan"
 


calliope/backend/backends.py
```diff
@@ -534,11 +534,18 @@ def _constraint_setter(
             imask: xr.DataArray, expr: tuple[xr.DataArray, str, xr.DataArray]
         ) -> xr.DataArray:
             lhs, op, rhs = expr
+
+            incomplete_constraints = (rhs.isnull() | lhs.isnull()) & imask
+            if incomplete_constraints.any():
+                raise BackendError(
```
brynpickering, line 540: For consistency with the parsing errors being raised (see #436), it would be good to prefix this error with f"(constraints, {name}): " so that a user need not enter a debugger to find out the name.

calliope/backend/backends.py
```diff
@@ -561,6 +568,12 @@ def add_expression(
         expression_dict: parsing.UnparsedConstraintDict,
     ) -> None:
         def _expression_setter(imask: xr.DataArray, expr: xr.DataArray) -> xr.DataArray:
+            incomplete_expressions = expr.isnull() & imask
+            if incomplete_expressions.any():
+                raise BackendError(
```
brynpickering, line 573: As above: prefix with f"(expressions, {name}): "

brynpickering (changes requested):
Looks good. Would benefit from a slight change in the error message. 

Also: needs a test for both errors. The easiest would probably be to un-nest these private functions and then test them directly with mismatched dataarrays (otherwise you have to test it all the way from a constraint/expression dict of where/expression strings).

jnnr:
> Looks good. Would benefit from a slight change in the error message.

Done 

> Also: needs a test for both errors. The easiest would probably be to un-nest these private functions and then test them directly with mismatched dataarrays (otherwise you have to test it all the way from a constraint/expression dict of where/expression strings).

This would be easier, but very constructed and pass by design. I'd prefer to test with a small example similar to the one that made us aware of the need for this check because that is more illustrative. 

I imagine two small custom yamls that are passed to `m.backend.add_constraint`. If you agree, can you give me some hints where to put the test, the yamls and which minimal model I can use for that?

brynpickering:
You can put this test in test_backend_pyomo.py, there is a class at the end
for the new backend. I'm working on clearing out all the old stuff and I
will restructure the test suite after that. For now, new backend tests can
go in that class.

RE your plan, that's fine with me. My idea was to have more focussed unit
tests for the fix rather than an integration test as you suggest. But I
don't hold a strong view on the best approach!

You can find pytest fixtures of simple models in conftest.py. the first one
there, which brings in a "supply" technology is the one to use. I'd store
the test constraint definition as a dictionary defined in e.g. a fixture
within the test class rather than as a yaml. It's easier to then find it
when inspecting the tests at a later point.

jnnr:
Thanks! I added a test for constraints with nans. Can add an similar for expressions if you agree with the setup of the test.

calliope/test/test_backend_pyomo.py
```diff
@@ -2083,3 +2084,22 @@ def test_raise_error_on_preexistence_diff_type(self, simple_supply_new_build):
...
+            simple_supply_new_build.backend.add_constraint(
+                simple_supply_new_build.inputs,
+                constraint_name,
+                constraint_dict,
+            )
+
+        assert check_error_or_warning(
+            error,
+            "(constraints, constraint-with-nan): Missing rhs or lhs for some coordinates selected by 'where'. Adapting 'where' might help.",
+        )
```
brynpickering, line 2105:
```suggestion
    def test_add_constraint_with_nan(self, simple_supply_new_build):
        """ 
        A very simple constraint: For each tech, let the annual and regional sum of `carrier_prod` be larger than 100.
        However, not every tech has the variable `carrier_prod`.
        How to solve it? Let the constraint be active only where carrier_prod exists by setting 'where' accordingly.
        """

        constraint_dict = {
            "foreach": ["techs"]
            "equation": "sum(carrier_prod, over=[nodes, timesteps]) >= 100"
            # "where": "carrier_prod"  # <- no error would be raised with this uncommented
        }
        constraint_name = "constraint-with-nan"
        
        with pytest.raises(exceptions.BackendError) as error:
            simple_supply_new_build.backend.add_constraint(
                simple_supply_new_build.inputs,
                constraint_name,
                constraint_dict,
            )

        assert check_error_or_warning(
            error,
            f"(constraints, {constraint_name}): Missing rhs or lhs for some coordinates selected by 'where'. Adapting 'where' might help.",
        )
```

brynpickering (changes requested):
Test looks fine except that there is a lot of overhead to having it stored as YAML. See my suggestion which would do away with the need for the YAML and the fixture which loads it from file. 

I should point out that yes we do have lots of YAMLs in the tests and at some point we thought it would be a great idea to have little test models that we could then dump into the docs to show examples of different use cases for different constraints. That has since been shelved and the YAMLs in the tests are now mostly legacy (except the frequently used "simple" models) and due to be culled.

brynpickering (approved):

calliope/backend/backends.py
```diff
@@ -918,6 +927,35 @@ def _add_constraint_or_expression(
             name, component_da, component_type, component_dict, references
         )
 
+    @staticmethod
+    def _check_expr_imask_consistency(expression, imask, description):
```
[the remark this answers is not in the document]
jnnr: Not so sure, as it is a private method used in a private sub-method. Seems very dependent on the implementation to me - but maybe I am wrong?

jnnr:
From my side, this is ready. Shall we merge, @brynpickering?

brynpickering (approved):

Issue #433, opened 2023-03-17
Compare LPs to improve testing of constraints


brynpickering:
## Problem description

The new math formulation implementation (#411) has a time penalty for very small models which slows down the tests. 

To speed things up, we should reduce how much of a backend model we build on testing the validity of different constraints. 

Following the approach of Oemof, we can build incomplete models and then write an LP file that we do a string comparison against. Then we don't need to solve the model each time, nor do we need (or want) to run `m.build()` in the tests. Instead, we should use the backend API for adding constraints/variables/expressions one at a time.

## Calliope version

v0.7.0-dev

brynpickering:
Method to do so added in #453, now we just need to implement all tests in this format.

PR #435, opened 2023-03-20
Feature: Add coordinate reference to obj name strings

brynpickering:
Partially fixes issue #433 by providing the basis to have coordinates in the final LP.

This allows a user to decide to update the string representation of their backend objects to include references to the object's coordinates in the corresponding xarray dataset. To save memory, this is not done every time when building the model and should instead be considered a debug option.

From tests, the string representation update uses ~10% of the memory required to initially build the model. 

The result is, e.g.:
> parameters[energy_eff][0]*variables[carrier_con][0]

&#8595;

> parameters[energy_eff][a, test_demand_elec]*variables[carrier_con][electricity, test_demand_elec, a, 2005-01-01 00:00]

This will filter through to the LP file, e.g.:

> parameters(energy_eff)(0)*variables(carrier_con)(0)

&#8595;

> parameters(energy_eff)(a__test_demand_elec) * variables(carrier_con)(electricity__test_demand_elec__a__2005-01-01 00:00)

* Wrap pyomo parameter and variable classes with a set of methods to update result of getting the `name` property.
* Add `expand_object_string_representation` backend object method to update the string representation of all parameters and variables.

Specific questions for reviewers:

1. Ideas on a less verbose method name?
2. Does it make sense to make this a separate method that requires knowledge of the its existence, or to have an arg on "build" (e.g., "debug_mode") that will call the updater function? Since the method *can* be called after building the model - i.e., there's no need to re-build the backend model to update the strings - it does seem to make sense to have it as a separate step. Equally, it is perhaps a little hidden.
3. In a similar vein to (2), should this be automatically applied if requesting to "save_logs" in the call to "solve"? Saving the logs might suggest a wish to inspect the LP, which only really makes sense if you have the coords in the strings.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


sjpfenninger:
>     1. Ideas on a less verbose method name?

* `make_verbose`
* `verbose_strings`
* `expand_verbosity`

>     2. Does it make sense to make this a separate method that requires knowledge of the its existence, or to have an arg on "build" (e.g., "debug_mode") that will call the updater function? Since the method _can_ be called after building the model - i.e., there's no need to re-build the backend model to update the strings - it does seem to make sense to have it as a separate step. Equally, it is perhaps a little hidden.

I think a separate method is easiest to maintain and if documented, as easy to use as setting an arg on build.

>     3. In a similar vein to (2), should this be automatically applied if requesting to "save_logs" in the call to "solve"? Saving the logs might suggest a wish to inspect the LP, which only really makes sense if you have the coords in the strings.

I wouldn't do that since you might just want the logs. But we should make a note to think about what changes to make in the CLI, and have a flag there to enable this?

sjpfenninger (approved):

Issue #425, opened 2023-03-17
Inconsistent backend after adding constraint that applies nowhere


jnnr:
## Problem description

When adding constraints (m.backend.add_constraint) that do not apply anywhere because of the where, the backend is left in an inconsistent state:

The constraint does not appear in m.backend.constraints. But when you try to add a new version of the constraint with the same name, you will be notified that the constraint already exists. That's because it remains in the pyomo list created in _add_constraint_or_expression.

This list should be deleted if the constraint (/other optimisation problem element) does not apply anywhere.

## Steps to reproduce the problem

Add a constraint with a where string does not apply anywhere.

## Calliope version

v0.7dev

jnnr:
Closed by #437

PR #437, opened 2023-03-22
Fix: Add method to delete empty pyomo list objects

brynpickering:
Fixes issue(s) #425 

* Delete pyomo list object if the component being built evaluates to an array of NaNs

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


calliope/backend/backends.py
```diff
@@ -518,8 +518,10 @@ def add_parameter(
             default=default,
             use_inf_as_na=use_inf_as_na,
         )
-        if not parameter_values.shape and parameter_da.isnull().all():
-            parameter_da = parameter_da.astype(float)
+        if parameter_da.isnull().all():
+            self._delete_pyomo_list(parameter_name, "parameters")
+            if not parameter_values.shape:
```
jnnr, line 523: What does this check mean? Does it check that it is a np.array or xr.DataArray? Can this be checked explicitly?

jnnr (commented):
Thanks for the feature! Lost my example that let us notice the issue, unfortunately :( so I could not test that it is fixed.

calliope/backend/backends.py
brynpickering: It checks if there are any dimensions on the array. If there aren't and the value is NaN then the dtype of the "array" (which is actually not an array, since it isn't indexed over anything) is explicitly changed to match the NaN dtype (i.e., float). This ensures we don't get issues elsewhere.

Not sure how to make it more explicit. Maybe 
```python
is_array = parameter_values.shape
if not is_array:
    # NaN-filled dimensionless arrays must have a non-"object" dtype to 
    # avoid issues elsewhere.
    parameter_da = parameter_da.astype(float)
```

jnnr: In which case would `parameter_da.shape` return `None`? I tested with numpy and xarray - if the array is empty, I get an empty tuple. Other examples, like `np.nan.shape`, throw an error.

brynpickering: Also, `parameter_da` can only be an xarray DataArray, so it would never have reason to call e.g. `np.nan.shape`

jnnr: Thanks, now I got it!

Issue #427, opened 2023-03-17
Restructuring maths yaml files


FraSanvit:
## Problem description

Restructure the .yaml files such that there is a base.yaml file including the default variables, constraints, .. etc. Alongside this, specific run-mode .yaml files are provided (e.g. plan, operate, spores...). Plan mode can be the default so included in the base.yaml resulting in an empty plan.yaml file.

In this way, we can remove all the references to "run.mode=operate" in the `where` strings.

[Possible issue]: such files are specified when loading the configs, so it might be the case that once the model has been loaded, other custom math are not available anymore. E.g. Running a model in plan mode, saving it, loading it and then wanting to run it in operate mode.

## Calliope version

v0.7.0-dev

PR #439, opened 2023-03-22
Feature: custom run mode math files

brynpickering:
Fixes issue(s) #427 

* Math files moved to "math" subdir
* Custom math file per run mode added, "run.mode=operate" refs moved from base math. NOTE: there is no empty "plan.yaml" file as suggested in #427 since an empty file is not a valid YAML file and will raise an exception on attempting to load it.
* Method to override loaded math dict with run mode math dict on `build`. NOTE: this isn't applied if the user already referenced the relevant run mode custom math in their custom math list (`model_config.custom_math`), allowing the run mode math to be overridden by additional custom math if required.

Reviewer checklist:

- ~~Documentation updated~~
- ~~Changelog updated~~


calliope/core/model.py
```diff
@@ -246,15 +246,15 @@ def _add_math(self, custom_math: list) -> AttrDict:
         """
 
         base_math = AttrDict.from_yaml(
-            os.path.join(os.path.dirname(calliope.__file__), "config", "base_math.yaml")
```
sjpfenninger, line 249: This could be just `base.yaml` now that it's inside a `math` subdir

calliope/core/model.py
```diff
@@ -301,6 +301,23 @@ def _generate_default_dict(self) -> AttrDict:
...
         )
 
+    def _add_run_mode_custom_math(self) -> None:
+        """If not given in the custom_math list, override model math with run mode math"""
+        run_mode = self.run_config["mode"]
+        # FIXME: available modes should not be hardcoded here.
+        # They should come from a YAML schema.
+        not_run_mode = {"plan", "operate", "spores"}.difference([run_mode])
+        run_mode_mismatch = not_run_mode.intersection(self.model_config["custom_math"])
+        if run_mode_mismatch:
```
sjpfenninger, line 311: This extra check suggests to me that there should be two separate directories for math: one with a file for each run mode and one with custom math. Then you don't need to do this extra checking. Or you could check if a run mode was defined that doesn't have a corresponding file in the run mode math directory.

sjpfenninger (approved):
Looks good to me, though I also made one optional suggestion for a slight file system rearrangement.

calliope/core/model.py
brynpickering: The check here is actually for whether some run mode custom math has been applied by the user explicitly, which I think is a valid use case if you want to control in what order that custom math appears (otherwise it comes at the end and cannot be overridden). You may also want to apply the custom math of an in-built run mode so you can have it available for your custom run mode. 

I'm also not keen on creating more sub-directories if it can be helped.

Agreed though that this would be better if not hardcoded here. It would also be great if the run mode custom math wasn't linked to its custom math by filename alone. Ideally, this link would be in some config file instead and would allow a run mode in future to e.g., depend on multiple custom math files.

sjpfenninger: Ok, let's park it for now, and keep it hardcoded as-is with the FIXME marker.

PR #443, opened 2023-04-11
Update and clean docs

brynpickering:
First clean-up of the docs in quite a while!

* Ran `make linkcheck` and fixed all broken and redirected links. Some still say broken, but I've checked those manually and they do work fine.
* Removed most references to plotting and group constraints.
* Updated Sphinx and RTD config files to match latest recommendations.
* Updated Sphinx to the shiny new [v6](https://blog.readthedocs.com/sphinx6-upgrade/), which works so long as we stick to py3.10.
* Fixed things that cause warnings on building the docs locally.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/requirements_docs.yml
+++ b/requirements_docs.yml
@@ -5,8 +5,8 @@ channels:
   - nodefaults
 
 dependencies:
-  - python=3.11
-  - sphinx~=6.0.0
+  - python=3.10
+  - sphinx~=6.1.0
   - sphinx-autodoc-typehints < 2
   - ruamel.yaml < 18
   - nbconvert < 8 # To generate HTML files from example notebooks


sjpfenninger:
Generally looks good, but it seems like 0.6 didn't actually work given that 0.5 is pinned? What were the issues?

brynpickering:
There was an unresolvable package conflict when creating the conda env on RTD. I couldn't find any reason for it to be happening - I have no problems creating the env locally - nor any insight online. You can see the error if you check out the older failed RTD builds.

brynpickering:
To be precise:

> Encountered problems while solving:
> - package sphinx_rtd_theme-1.2.0-pyha770c72_0 requires docutils <0.19, but none of the providers can be installed

Docutils <0.19 is very clearly [available on conda forge](https://anaconda.org/conda-forge/docutils/files?version=0.18.1) so I can't really fathom it.

sjpfenninger:
Hm - where does `sphinx_rtd_theme` come from, I wonder? We shouldn't require it?

brynpickering:
Of note, xarray [documentation requirements file](https://github.com/pydata/xarray/blob/main/ci/requirements/doc.yml) sets no upper bound on sphinx, but ends up with [v5.3](https://readthedocs.org/api/v2/build/20092835.txt) in its latest builds, not v6. Granted, it has many more dependencies, but maybe it is `sphinx_rtd_theme` causing the issue again...

brynpickering:
OK, found the culprit, sphinx_rtd_theme doesn't work with py3.11 (see https://github.com/conda-forge/sphinx_rtd_theme-feedstock/issues/28) since docutils <19 doesn't support it. Let's see if the update I just pushed works...

sjpfenninger:
Looks like it worked!

Is the change to the license strings at the top of each file in order to automatically parse the file-level docstrings without a license text pollution?

brynpickering:
Success!

brynpickering:
> Is the change to the license strings at the top of each file in order to automatically parse the file-level docstrings without a license text pollution?

Precisely. It allows us to document entire module APIs without repetition of the license info _but_ keeping everything else in the module-level docstring.

sjpfenninger (approved):

Issue #422, opened 2023-03-17
Remove `equals` and `scale` bounds


sjpfenninger:
Right now, variables can define `equals` and `scale`:

```
variables:
  energy_cap:
    foreach: [nodes, techs]
    where: "NOT run.mode=operate"
    bounds:
      min: energy_cap_min
      max: energy_cap_max
      equals: energy_cap_equals
      scale: energy_cap_scale
```

We want to remove the functionality for both of these.

`equals` can be handled by setting `min` and `max` to the same value, `scale` is rarely needed and can be handled separately (e.g. with an additional parameter) where it is needed.

Pyomo recommends the approach of setting min and max to the same value to fix variables:

https://pyomo.readthedocs.io/en/stable/working_models.html#fixing-variables-and-re-solving

sjpfenninger:
Done in #466

PR #445, opened 2023-04-17
[work-in-progress] delete 'equals' bounds in variables' definitions

FraSanvit:
Partial Fixes issue(s) #422

* Remove 'equals' bounds from variables in subset.yaml
* Update defaults
* Update parsing

Work-in-progress:
* Check `operation` mode

Open questions
* shall we get rid of also those constraints such as `energy_capacity_per_storage_capacity_equals` that can be replaced by the joint action of max and min constraints? In this regard, there are other contraints which have not e.g. the min constraints as per the case of `energy_cap_equals_systemwide`.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/calliope/backend/parsing.py
+++ b/calliope/backend/parsing.py
@@ -46,7 +46,6 @@ class ParsedEquationDict(TypedDict):
 class UnparsedVariableBoundDict(TypedDict):
     min: str
     max: str
-    equals: str
     scale: NotRequired[str]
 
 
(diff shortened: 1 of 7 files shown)


brynpickering:
> shall we get rid of also those constraints such as energy_capacity_per_storage_capacity_equals that can be replaced by the joint action of max and min constraints? In this regard, there are other contraints which have not e.g. the min constraints as per the case of energy_cap_equals_systemwide.

Yes, and there's no reason that `energy_cap_min_systemwide` shouldn't already exist, so that can go back in when removing `energy_cap_equals_systemwide` (same from `unit_equals_systemwide`). Arguably, these systemwide parameters could be dropped in favour of a user defining them as custom constraints, but since they are already implemented maybe they can stay...

brynpickering:
Closed in favour of #466 due to the effort of merging latest changes into this version. Sorry @FraSanvit !

PR #463, opened 2023-04-25
Update: reformat yaml and imports

brynpickering:
General cleanup of the files so that we don't keep having these changes creating merge conflicts.

* Apply [isort](https://pycqa.github.io/isort/) to all python files, to correctly sort all the imports. We were mostly doing this correctly already, this just tended to sort things alphabetically.
* Add isort to pre-commit (I have added it to format-on-save in VSCode).
* Change default YAML indentation from 4 spaces to 2. This works well for lists of dicts (as we now use in the YAML math syntax), so makes sense to apply across the board.
* remove a `.vscode` directory that slipped in on a PR 4 years ago, and update `.gitignore` to catch `.vscode` in subdirectories in future.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


sjpfenninger (approved):
Fine with me! At some point maybe we should add https://editorconfig.org/ to specify indentation etc

Issue #429, opened 2023-03-17
Add public API to update a built optimisation model


brynpickering:
## Problem description

Since we have YAML math we also have a new set of methods to build optimisation models using Pyomo (and hopefully Gurobi python API). Certain methods need to be made available on the model object so a user can do the following:

- [x] Check the content of a specific variable/constraint/expression
- [x] Update a parameter's values (and update the corresponding expressions/constraints/objectives that rely on that parameter)
- [x] Fix a variable, so it acts as a parameter
- [ ] (De)activate a constraint so it is(n't) applied in the optimisation
- [x] Update decision variable bounds

This replaces #293 for the 0.7.x development (#293 will stay open for any future v0.6.x dev).

## Calliope version

v0.7.0-dev

brynpickering:
Ignoring constraint activation/deactivation since it can't be implemented with the gurobi python interface, so is a very specific Pyomo feature. If a user really wants to use the pyomo feature, they can do so directly with sufficient knowledge of pyomo and xarray.

brynpickering:
Completed in #471 based on not allowing contraint (de)activation

PR #471, opened 2023-05-19
Feature: public methods to interface with new backend model instance

brynpickering:
Fixes issue #429 

General restructuring:
* Move pyomo backend model class to its own file.
* Move all backend interface methods to be public on the backend object itself, rather than making them available as methods on the model object (e.g., "model.to_lp" -> "model.backend.to_lp"). This came after I realised just how many methods would need to be passed through to the model object (and each one would need wrapping in functions to check the backend model has been built and to copy across docstrings from the backend) - seemed like too much. We just have to get people used to the "backend" as a thing (or give it a different name).
* Store a copy of the model data in the backend so it doesn't need to be supplied every time a method is called.

New methods:
* Update parameter values
* Update variable bounds
* Fix/unfix a variable

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


sjpfenninger (approved):
Good to go - though we should track points to revisit later, e.g. the naming of `global_expressions` (one alternative suggestion is to call them `named_expressions`)

PR #487, opened 2023-10-23
Clean-up alongside dependency pinning updates

brynpickering:

* Update to latest versions of dependencies
* Fix issues created by these updates:
    * Can no longer use the pandas context `use_inf_as_na`.
    * Xarray behaves differently when trying to select on a stacked array, so I've changed a helper function to perform similarly to a solution provided in an xarray issue.
    * Cannot format strings to datetime if it's in ISO format and only a partial match of the string. Need to reference the ISO code directly.
    * Slight change in LP file structure with a Pyomo update
* Introduce Ruff as a linter (sorry Hound) and add pre-commit hooks for Ruff and for linting/formatting the tutorial notebooks.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


calliope/test/test_math.py
```diff
@@ -1,8 +1,8 @@
-import filecmp
 from pathlib import Path
 
 import numpy as np
 import pytest
+from pyomo.repn.tests import lp_diff
```
sjpfenninger, line 5: Is there a good reason to replace filecmp with this pyomo functionality?

requirements.txt
```diff
@@ -1,14 +1,14 @@
...
+jsonschema >= 4.17, < 4.19
+natsort >= 8, < 9
+netcdf4 >= 1.2, < 1.7
+numpy >= 1, < 2
+pandas >= 2, < 3
+pyomo >= 6.5, < 7
+pyparsing >= 3.0, < 3.1
+ruamel.yaml >= 0.17, < 0.18
+scikit-learn >= 1.2, < 1.3
+xarray >= 2023, < 2024
```
sjpfenninger, line 14: Not sure this is a good idea, why not keep pinning versions of key dependencies more explicitly?

.pre-commit-config.yaml
```diff
@@ -1,11 +1,21 @@
+default_language_version:
+  python: python3
+
 repos:
   - repo: https://github.com/psf/black
-    rev: stable
+    rev: 23.7.0
```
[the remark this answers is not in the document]
brynpickering: Yeah, we can probably have it as stable tbh. If pinning, the pre-commit CI also has a bot which will periodically prod to update to the latest version, so we could go with that. I don't have a strong opinion either way.

calliope/test/test_math.py
brynpickering: Only that pyomo ignores some immaterial differences between the files (floats and ints are considered equivalent if they are the same (`10` == `10.0`). When there is a diff, it is formatted a bit nicer, too.

requirements.txt
brynpickering: Some flexibility allows for calliope to be installed more easily alongside other packages. Between major releases of mature dependencies, there should be no breaking changes to be concerned with. Between younger packages, allowing only changes in patch releases seems fair. We could allow only changes between patch releases for everything if you prefer, but I would allow some of this flexibility.

.pre-commit-config.yaml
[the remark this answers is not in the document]
sjpfenninger: I also don't mind either way, so maybe going with `stable` is easiest as it doesn't require frequent changes?

sjpfenninger (dismissed):

sjpfenninger (approved):

PR #488, opened 2023-10-23
Restructure project

brynpickering:

* Moving to using `pyproject.toml` as is the recommended way to define python project metadata now.
* Move calliope into `src` to make sure tests are not getting confused about relative filepaths in a way we don't realise.
* Move from azure pipelines to github actions, which allow us to leverage the growing number of pre-built actions (like micromamba for speedy env install).
* Move tests out of the calliope package itself.
* Move and restructure requirements files so that dev install becomes a much easier one-liner for the mamba/conda part.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


sjpfenninger (approved):

brynpickering:
Hoping that codecov will recover and realise the error of its ways once this is merged into main and new PRs build on top of it (I am at 94% coverage locally, not sure why it is at 92% here.

Issue #461, opened 2023-04-22
Some backend component attributes need to be serialised before saving to NetCDF


brynpickering:
## Problem description

Each component in the backend dataset (e.g., the `energy_cap` variable dataarray) has its own attributes; a combination of useful metadata (the description of the variable) and internally useful stuff, like whether the variable has coordinate data attached to the string representation of objects in its array. One thing which *may* be considered as useful metadata, but was initially conceived for internal processing, is "references". This is a set of the other components in the backend in which the component in question is referenced. "cost" is referenced in the objective, "carrier_prod" is referenced in the "cost" expression, etc. 

These attributes make their way out of the backend dataset and into the model results dataset by way of variables and global expressions.

So, this "references" set needs to be serialised manually for every array before the netcdf can be saved to file, OR it is dropped since it is loses its relevance once it is outside the backend dataset.

I would be more in favour of dropping it, but I don't know if others can see value in keeping it in the results. When loading a model with results from file, it would allow you to see e.g., which constraints `carrier_prod` was used in, even though you would no longer have access to the constraint objects (until you run `model.build()`).

## Calliope version

0.7.0-dev.

PR #489, opened 2023-10-23
Add set serialisation & check for lists in attrs

brynpickering:
Fixes issue(s) #461 

* Extend the data types that can be included in xarray attributes to include sets, which are serialised as lists.
* Extend attribute serialisation to those attached to data variables, not just those at the top-level of the xarray dataset.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


sjpfenninger (dismissed):
Good to go
EDIT: though the tests actually fail, so maybe not

Issue #450, opened 2023-04-18
Allow variables and global expressions to be included in a YAML constraint "where" string


brynpickering:
## Problem description

Since variables and global expressions *must* be defined before constraints are added, they could also be used to "where" a constraint array, e.g.:

```python
constraint_dict = {
    "foreach": ["techs"]
    "equation": "sum(carrier_prod, over=[nodes, timesteps]) >= 100"
    "where": "carrier_prod" 
}
```

The implementation would require first checking "model_data" to see if the component is a parameter. If it isn't, then check the backend dataset variables/expressions to see if it is one of those. We shouldn't have to worry about name overlaps since a user cannot define the same name for any optimisation problem component (i.e., a parameter and decision variable cannot have the same name).

## Calliope version

v0.7.0-dev

sjpfenninger:
Addressed in #490

PR #490, opened 2023-10-23
Decision vars & global exprs in where strings

brynpickering:
Fixes issue(s) #450 

Since there is a strict order of things in the backend (parameters -> variables -> expressions -> constraints -> objective), we will have already constructed variables by the time we construct expressions and will have access to both variables and expressions by the time we reach constraints. Therefore, their existence can be checked in the constraint/objective where string. It makes some things less verbose.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


sjpfenninger (approved):
Good to go, with the proviso that the docs should highlight how vars and global_expressions are allowed in where strings and maybe some tips on how to differentiate between them when reading existing constraints

Issue #452, opened 2023-04-19
Separate top-level config and parameters 


brynpickering:
## Problem description

`run` and `model` configs mix up configuration and parameters. We want to be able to pipe these parameters to the Calliope model dataset and then onto the backend. We handle that at the moment with hardcoding which are parameters in the core code, which isn't ideal. An idea on how to change these YAML structures:

```yaml
config:
  init:
    name: null
    calliope_version: null
    random_seed: null
    subset_time: null
    time: {}
    timeseries_data_path: null 
    timeseries_data: null 
    timeseries_dateformat: "%Y-%m-%d %H:%M:%S"
  build:
    custom_math: []
    backend: pyomo
    cyclic_storage: true
    ensure_feasibility: false
    mode: plan
  solve: 
    operate_window: null
    operate_horizon: null
    operate_use_cap_results: false
    spores_number: 3
    spores_score_cost_class: spores_score
    spores_slack_cost_group: null
    spores_save_per_spore: false
    spores_save_per_spore_path: null
    spores_skip_cost_op: false
    save_logs: null
    solver_io: null
    solver_options: null
    solver: cbc
    zero_threshold: 1e-10

parameters:
  reserve_margin: {electricity: 0.1}
  bigM: 1e9
  objective_cost_class: {"costs": {"monetary": 1, "emissions": 0.1}}
  my_param: {("costs", "carriers"): {("monetary", "electricity"): 1, ("monetary", "heat"): 0.1}}
  spores_slack: 0.1
  spores_objective_cost_class: {"costs": {spores_score: 1, monetary: 0} }
```

Other things:
- `file_allowed` is deprecated.
- All config items could become arguments in the relevant methods (`init` in `calliope.Model`)/(`build` in `calliope.Model.build`)/(`solve` in `calliope.Model.solve`).
- ^ config items then have to be flattened so they can be used by args.
- parameters can be dimensionless or have dimensions in a dict form (for non-tech/node/tech+node dims). This is very simplified and wouldn't work well for large parameters. That's ok because we would solve it in the long (hopefully medium) term with #92. 
- Users can then add their own top-level parameters to 'parameters' for use in their custom constraints.

## Calliope version

v0.7.0-dev

PR #497, opened 2023-10-24
Update top level config params

brynpickering:
Fixes #496 with new parameter names and no `equals` or `scale`.

Fixes issue #452

EDIT: is automatically the case ~Possible TODO: allow init kwargs to be passed to the built-in example models on their initialisation (e.g., `calliope.examples.national_scale(override_dict={"config.init.subset_time": ...})` -> `calliope.examples.national_scale(subset_time=...)`)~

TODO: add top-level parameter tests.

* run and model config split into init/build/solve config keys with a flat dict layout in each.
* parameters moved out of config into a top-level parameters key with a syntax discussed in #452 and in a number of discussions.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


src/calliope/example_models/national_scale/model.yaml
```diff
@@ -4,27 +4,28 @@ import: # Import other files from paths relative to this file, or absolute paths
...
+  build:
+    ensure_feasibility: true # Switches on the "unmet demand" constraint
+    mode: plan # Choices: plan, operate
+
+  solve:
+    solver: cbc
+    zero_threshold: 1e-10 # Any value coming out of the backend that is smaller than this (due to floating point errors, probably) will be set to zero
+
+parameters:
+  objective_cost_class:
```
sjpfenninger, line 25: Thinking of #484 , this could also default to `data:1, index: [monetary], dims: costs` unless otherwise specified

src/calliope/backend/latex_backend_model.py
```diff
@@ -13,37 +13,37 @@
 from calliope.backend import backend_model, parsing
 from calliope.exceptions import ModelError
 
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
brynpickering: To avoid circular import errors for imports that are just used for type checking. I think it's handled better in py>3.10, but we'd need to keep this for py3.9 compatability.

src/calliope/example_models/national_scale/model.yaml
brynpickering: hmm. perhaps.

PR #506, opened 2023-10-26
Clean parsers

brynpickering:
Refactoring of the expression and where string parsers to be easier to maintain in future.

* Move to an abstract base class for our parsers, so they have the ability to output an xarray dataarray or a string (for latex math), and anything downstream can expect one of those types
* Move constraint expression setting to the expression parsing, so that we get one dataarray (of comparison expressions) instead of an lhs/rhs. Keeps the data structure in line with the other optimisation problem components.
* Rename `as_latex` to `as_math_string` which I think is more useful a description.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


sjpfenninger (dismissed):
Maybe also rename the `TestAsLatex` class

PR #517, opened 2023-11-10
Add reserves margin constraints

FraSanvit:

* add capacity values and operating reserve values (both time-varying)
* add planning and operating reserve group constraints
* allow time-series inputs in group constraints

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

Open questions:

- [ ] Naming of the target reserves
- [ ] Eligibility of technologies in providing the operating reserves


jmorrisnrel:
I finally got a chance to review this functionality. Looking at the code the math of the constraints seems to make sense although I've been getting a number of errors/issues with running tests. I've included the simple model I've been using to test the behavior consisting of a flat 20MW demand and one supply generation capex tech. Perhaps I’m misusing the constraints or didn’t set up my environment correctly but some of the issues are definitely in the PR code here.
 
Planning reserve share isn't behaving like I'd expect:
Example: Putting a target_reserve_share_equals of 0.5 in a system with flat 20MW demand and a single generator, I would expect the resulting capacity to be 30 MW representing (1+0.5)x the peak load of 20MW (assuming a default cap_value of 1.0). Instead I get 10MW of generation capacity.
When I specify the tech in the group definition it drops the resulting capacity of the generator to 0 and doesn't meet any demand.
 
Other constraints throw a Pyomo error including the operating constraints:
 
target_reserve_abs_min/max/equals and target_reserve_adder_min/max/equals all throw an AttributeError when I try and include all loc/techs by leaving those filters blank or include techs in the group definition

The operating constraints likely fail because there is no default value set for the operating_reserve parameter in defaults.yaml
 
target_reserve_share_operating_min:

If I set the operating_reserve to 0 (just representing a flat reserve) I get similar strange behavior with the resulting capacities. Setting reserve_share_operating_equals to 0.5 I would similarly expect 30MW but instead get 0.5 kW. And a similar story of 20MW if I use _min.

target_reserve_abs_operating_min/max/equals has the same AttributeError as the planning reserve constraints.
[inputs.zip](https://github.com/calliope-project/calliope/files/13709067/inputs.zip)

FraSanvit:
Hi @jmorrisnrel!
I've created a [repo](https://github.com/FraSanvit/reserve-margins/tree/main) with examples and a detailed guide on using the new functionalities. Also, I've tidied up and fixed some minor issues in the branch. Hope this can help.

calliope/backend/pyomo/constraints/group.py
```diff
@@ -1,892 +1,1416 @@
...
+        else:
+            loc_tech_carriers_transmission = []
+
+        loc_tech_carriers_con_all = [
+            loc_tech_carrier
+            for loc_tech_carrier in backend_model.loc_tech_carriers_con
+            if loc_tech_carrier not in loc_tech_carriers_transmission
+        ]
+
+        rhs = (
```
jmorrisnrel, line 762: I've been running into issues with the reserve margins when transmission is included in the model. It looks like (both directions) of transmission is included in the energy_cap calculations on the lhs of the reserve constraints, so adding a reserve doesn't result in generation capacity that would meet the reserve requirements. If I set `target_reserve_share_equals` in the sample model below (a single supply at one location and demand at another) the model either doesn't solve or only builds the supply to 1/3 of the reserve requirement with` ensure_feasibility`. Not sure if it would be possible to leave transmission out by default but allow you to specify transmission techs in the constraint and have them be added in if desired.

If I try and filter out the transmission techs, the group constraint doesn't seem to work at all. Setting `target_resrve_share_equals` again with a technology filter (and enabling `ensure_feasibility`) results in the model building no supply at all even with unmet demand.

Looking into the code it looks like the issue is because the rhs of the share and adder constraints uses the `locs` list derived from `lhs_loc_tech_carriers` which doesn't contain the demand location if there is nothing in the lhs there. With transmission included the demand loc is included because of the transmission link, but filtering out transmission to try and get the model to build enough generation for the reserve removes that link and results in the loc not being included.

[sample_model.zip](https://github.com/calliope-project/calliope/files/14544028/sample_model.zip)

jmorrisnrel: The `group_cosntraint_loc_tech_carriers_{}_rhs` attribute also seems to not include demand loc_techs for the appropriate reserve constraints as the code only looks for "share" and "demand"/"import" in the constraint name (`constraint_sets.py:524`). Grabbing the rhs loc_techs from `get_group_lhs_and_rhs_loc_tech_carriers` would need that code adjusted so that demand and consumption techs are included in the rhs for the non-absolute reserve constraints.

Issue #498, opened 2023-10-24
Regression in runtime


brynpickering:
## Problem description

Sometime since around #487, everything has slowed down considerably. Likely it is due to some combination of dependencies and some buggy implementation we've introduced. 

Pandas v2.1.1 has a known regression that is likely hitting us when we do multiindex operations (https://github.com/pandas-dev/pandas/issues/55256), so we should probably pin to `<=2.0.3`. However, it only explains part of the problem...

5 slowest tests pre #487:
```
11.82s call     calliope/test/test_core_time.py::TestClustering::test_hartigans_rule
9.58s call     calliope/test/test_example_models.py::TestNationalScaleClusteredExampleModelSenseChecks::test_nationalscale_clustered_example_closest_results_cbc
7.21s call     calliope/test/test_example_models.py::TestModelPreproccessing::test_preprocess_time_clustering
6.74s call     calliope/test/test_example_models.py::TestUrbanScaleExampleModelSenseChecks::test_urban_example_results_area[solve]
6.43s setup    calliope/test/test_backend_pyomo.py::TestMILPConstraints::test_loc_techs_storage_capacity_max_purchase_milp_constraint
```
5 slowest tests in #487 with pandas=2.0.3:

```
20.23s call     tests/test_example_models.py::TestNationalScaleClusteredExampleModelSenseChecks::test_nationalscale_clustered_example_closest_results_cbc[run]
15.85s call     tests/test_core_time.py::TestClustering::test_hartigans_rule
15.41s setup    tests/test_io.py::TestIO::test_save_netcdf
15.34s call     tests/test_example_models.py::TestModelPreproccessing::test_preprocess_time_clustering
14.74s call     tests/test_example_models.py::TestNationalScaleExampleModelSenseChecks::test_nationalscale_example_results_glpk
```

5 slowest tests post #487 with pandas=2.0.3:
```
==================================================================== slowest 5 durations =====================================================================
33.98s call     tests/test_cli.py::TestCLI::test_run_from_netcdf
31.62s call     tests/test_example_models.py::TestNationalScaleClusteredExampleModelSenseChecks::test_nationalscale_clustered_example_closest_results_cbc[run]
29.08s call     tests/test_core_time.py::TestClustering::test_hartigans_rule
22.66s call     tests/test_example_models.py::TestNationalScaleClusteredExampleModelSenseChecks::test_nationalscale_clustered_example_closest_results_cbc[solve]
[...]

PR #519, opened 2023-11-10
Bump to py3.12 support

brynpickering:
Move from py3.9-3.11 to py3.10-3.12.

This is made possible by the recent release of a new pandas patch (v2.1.2) that _should_ have dealt with the regressions introduced in v2.1.1 (see #498), according to their [changelog](https://pandas.pydata.org/pandas-docs/version/2.1.2/whatsnew/v2.1.2.html).

EDIT: I've run some tests (national scale model over one year) and it seems that the pandas patch hasn't solved the issues entirely. The run has gone from approx 3min50 to 6min. This is offset by a ~30% reduction in peak "resident" memory consumption (no change in "heap" consumption between the two).

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


brynpickering:
The primary benefit of killing py3.9 support is that cleaner type hints are possible without needing to retain backwards compatibility (`Union[A, B]` in <=py3.9 vs `A | B` in >=py3.10).

The only issue I see is that it keeps us quite constrained (currently) to one version of pandas. Once more v2.1 patches are released, we'll have a bit more breathing room.

PR #526, opened 2023-12-20
Add skeleton of mkdocs

sjpfenninger:
First skeleton of mkdocs

Open issues:

* `custom_math.py` could also run through `gen-files` but because it is quite slow, I think it's better to run it in a seprate pre_build step. Already including those generated markdown files let's build time increase from ~1 second to around 10 seconds, so there is some issue with them anyway. The idea for now would be to run `python ./pre_build/generate_math.py` and any other pre-build steps manually (and in the case of readthedocs, add this as pre-build steps in the readthedocs configuration).
* The rendering of the fenced math blocks needs fixing, e.g.: https://facelessuser.github.io/pymdown-extensions/extensions/arithmatex/#alternative-math-blocks
* `extra: version: 0.7.0` in the configuration should come from the installed calliope module, not be hardcoded in the config file
* Only three custom math files as proof-of-concept, with the rest to be moved once #518 is merged

The idea would be to sort out any issues with the basic setup here, merge this basic setup, then revise and move over the remaining documentation step-by-step, thus also moving from the old `doc` to the new `docs` directory.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


docs/index.md
```diff
@@ -0,0 +1,39 @@
+# Calliope: energy system modelling made simple
+
+!!! note
```
brynpickering, line 3: This doesn't seem necessary. RTD allows for version warnings: https://docs.readthedocs.io/en/stable/versions.html#version-warning

Then you don't need to track version in the mkdocs config and don't need to worry about making it dynamic.

docs/index.md
```diff
@@ -0,0 +1,39 @@
...
+
+    This is the documentation for version {{ version }} ([Version history](version_history.md)). See the [main project website at www.callio.pe](https://www.callio.pe/) for more general information, including a gallery of models built with Calliope, and other useful information.
+
+Calliope focuses on flexibility, high spatial and temporal resolution, the ability to execute many runs based on the same base model, and a clear separation of framework (code) and model (data). Its primary focus is on planning energy systems at scales ranging from urban districts to entire continents. In an optional operational mode it can also test a pre-defined system under different operational conditions. Calliope's built-in tools allow interactive exploration of results.
```
brynpickering, line 7: It would be good to follow the convention of one sentence per line so that future git diffs are simpler. [Here's a proponent](https://nick.groenen.me/notes/one-sentence-per-line/)

docs/pre_build/generate_math.py
brynpickering, line 1: This is showing up as brand new. Have any changes been made from the other version?

docs/version_history.md
```diff
@@ -0,0 +1,5 @@
+# Version history
+
+{%
```
brynpickering, line 3: `--8<-- CHANGELOG.md` if using [pymdownx.snippets](https://facelessuser.github.io/pymdown-extensions/extensions/snippets/)

brynpickering (changes requested):
For math to be rendered, do you just need to add the relevant config and javascript?

https://squidfunk.github.io/mkdocs-material/reference/math/

docs/index.md
sjpfenninger: 👍

docs/pre_build/generate_math.py
sjpfenninger: Minor changes. I'll move the old file to the new location in a next commit so that the changes are visible.

docs/version_history.md
sjpfenninger: I tried that, but the `on_page_markdown` hook in `changelog_highlight.py` doesn't work with snippets and I'm not sure why. Possibly because the snippets pre-processor runs after that. So I propose to stick to this as-is for now.

sjpfenninger:
> For math to be rendered, do you just need to add the relevant config and javascript?
>
> https://squidfunk.github.io/mkdocs-material/reference/math/

Not enough if you want the fenced blocks to work (```math ...). The easiest solution might just be to replace those with the `$$...$$` block syntax.

brynpickering (approved):

PR #529, opened 2023-12-28
Add distance unit config item & default to km

### LINKED ISSUE #209: Document and log automatically derived distances
## Problem description

Calliope can automatically derive distances between two locations if both locations have coordinates. This mechanism is very practical, but not properly documented. Issues are:

* Coordinates can be given in lat/long _or_ with cartesian coordinates on a 2d plane. The latter is not documented.
* When coordinates are given in lat/long, the resulting distance is given in meters. That is not documented.
* There is no way of checking whether the result is as expected. An easy solution would be to log the calculated total distance: "Calliope derived a total link length of 25000 km."
* Kilometers would be a better unit than meters. It is more intuitive to me, even in an urban model. But for national or continental scale models, there is also a numerical issue:

```Python
In [1]: 0.99999995**100000 # 100 km high voltage ac transmission
Out[1]: 0.995012479065401 # noise starts in e-5

In [2]: 0.99999995**10000000 # 1000 km high voltage ac transmission
Out[2]: 0.9512294232839076 # noise starts in e-3
```

## Calliope version

0.6.4.dev

brynpickering:
Fixes #209 

* `config.init.distance_unit` option available to define whether distances should be derived in metres or kilometres

Reviewer checklist:

- [ ] Documentation updated
- [x] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This enables un-indexed parameters to be defined, as well as those indexed over 
 |new| parameter dimensions at the `tech` or `node` level can be enhanced using the new `parameter` definition syntax.
 For instance, `flow_cap` can be defined per `carrier`.
 
+|changed| Automatically derived transmission link distances default to kilometres, with the configuration option (`config.init.distance_unit`) to switch to the old default of distances in metres.
+
 |changed| |backwards-incompatible| Costs must be defined using the new `parameter` definition syntax.
 
 |changed| `flow_cap` (formerly `energy_cap`) is indexed over `carriers` as well as `nodes` and `techs`.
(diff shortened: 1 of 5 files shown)


timtroendle (approved):
Looks good to me. I also have to say I am not very familiar with the Calliope source code and the issue (any more).

Now that only lat/lon are allowed I wonder if there is a check for that. An easy check would be to make sure numbers are within ±90 and ±180. Maybe something smarter is possible. And maybe this is already implemented?

brynpickering:
Thanks. We now have a YAML schema! It include exactly what you suggest: https://github.com/calliope-project/calliope/blob/d288ff8e59b8bbfdf3c1b6ccc5dff842c1eb0000/src/calliope/config/model_def_schema.yaml#L888-L897

sjpfenninger (approved):
Looks good!

Issue #127, opened 2018-06-25
Increasingly masked rolling horizon


brynpickering:
## Problem description

Our current rolling horizon method (`operate` run mode) doesn't really allow planning for e.g. seasonality. We might want to think of the horizon as high resolution over the next 48 hours (hourly), then daily resolution (e.g. mean/max of the hours) for the few weeks after that, then weekly/monthly/annually after that, depending on total time horizon. 

It would allow future seasonality to be captured, even though at low resolution (indicative of our understanding of the climate, if perhaps not the weather). As the window rolls forward, information becomes clearer. This is particularly handy when considering long term storage.

This wouldn't really affect the `window` aspect of the `operate` run mode.

## Steps to introduce functionality

- [ ] Work out how to add the horizon points at which resolution changes into the YAML format
- [ ] Work out how to efficiently update a model with a newly masked time horizon on each run iteration.
    At the moment, we can just update some Pyomo params with the new horizon data, without having to rebuild the model, because the number of timesteps stays the same. This would involve a different number of timesteps on each iteration...

brynpickering:
An implementation was started in https://github.com/calliope-project/calliope/tree/windowsteps_matrix

brynpickering:
Some thoughts on this:

We could have a list of horizon periods for each window, e.g. in YAML:

```yaml
config.init:
  operate_window: 24h
  operate_horizon:
    - offset_start: 0
      offset_end: 47
    - offset_start: 48
      offset_end: 215 # 1 week later
      resolution: 24h
      agg_method: mean
    - offset_start: 216
      offset_end: 945 # 1 month later
      resolution: 1w 
      agg_method: mean  
    - offset_start: 946
      resolution: 1m
      agg_method: mean  
```

Then, every 24h window there would be a masked horizon of 2 days at full res, one week at 24hr res, one month at weekly res, and the rest at weekly res. For a year @ hourly resolution, this would lead to the maximum number of timesteps in a window being 48+7+4+11 = 70.  This would reduce as we go along as e.g., there will be fewer months in the future as we progress through the months.

brynpickering:
If we loaded this directly from file using #532 then it would require a table of the form:

|  |  2005-01-01 00:00 | ... | 2005-01-02 01:00  | ... | 
|--------|--------|--------|--------|--------|
| 2005-01-01 | 10 | ... | 10 | ... |
| 2005-01-02 |  | ... | 10 | ... |
| ... | ... | ... | ... | ... |

[...]

PR #539, opened 2024-01-18
Add operate mode

### LINKED ISSUE #432: Reincorporate operate mode in v0.7.0
## Problem description

The new functionalities of the under-development v0.7.0 require a rethinking of the way run modes other than `plan` are incorporated in the code. The  `operate` modes need to be re-implemented in a compatible way and properly tested

## Calliope version

v0.7dev

brynpickering:
Fixes #432 

This is mostly to confirm that we're happy with how we approach adding new run modes into the new calliope internals.
I would love to have operate mode work with #127 but suspect it needs more planning.

* Add a slim-line operate mode back into the model
* Switch from defining operate window/horizon as timesteps to being pandas time frequencies

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


brynpickering:
Worth noting that in 4991efefb5459f61dffa023d8e10ed2a002d68be I implemented it in the backend, then I switched it to being implemented in `model.py`. I think it works better as a "controller" outside the backend, but maybe not polluting `model.py`? Imagine adding `spores` mode to the same class and then you start to see `calliope.Model` becoming quite large. Maybe that's ok though

sjpfenninger (changes requested):

PR #551, opened 2024-01-26
Feature: Gurobi backend interface

### LINKED ISSUE #349: Gurobi solver interfaces
We want to interface with Gurobi more directly, to leverage performance improvements in the solver backend.

One method is to use the [Gurobi Python interface](https://www.gurobi.com/documentation/9.1/quickstart_mac/cs_grbpy_the_gurobi_python.html)

Another method is the Pyomo Gurobi [persisent solver](https://pyomo.readthedocs.io/en/stable/library_reference/solvers/gurobi_persistent.html#) ([usage documentation](https://pyomo.readthedocs.io/en/stable/advanced_topics/persistent_solvers.html))

The Python interface would probably lead to the greatest improvements, but needs more work to make our mathematical formulation more general (I.e. not written for Pyomo). However, the persistent solver could be a 'quick' fix for models that end up making small changing and then re-solving (cf. SPORES).

This issue can be used to track any updates in Gurobi solver interfaces.

### LINKED ISSUE #597: Pyomo v6.7.2 regression: Gurobi Python interface not found
### What happened?

Seen in #551 CI tests, some of the [changes in Pyomo v6.7.2](https://github.com/Pyomo/pyomo/blob/main/CHANGELOG.md#pyomo-672---9-may-2024) have caused an issue with using `solver_io=python` with gurobi. Easiest fix for now is to pin Pyomo to <v6.7.2.

- [X] macOS
- [X] Windows
- [X] Linux

### Version

v0.7.0-dev4

### Relevant log output

```Shell
[gw1] linux -- Python 3.10.14 /home/runner/micromamba/envs/calliope-ubuntu-latest-310-25fc8eb0047f7544a957ed75a82f2329ea5bc9f33ae1c087115cbb13d6ef0f33/bin/python3.10

self = <tests.test_backend_pyomo.TestLogging object at 0x7fc2dd24cbb0>
caplog = <_pytest.logging.LogCaptureFixture object at 0x7fc2dae78910>
gurobi_model = <calliope.model.Model object at 0x7fc2dc0c1990>

    def test_no_duplicate_log_message(self, caplog, gurobi_model):
        caplog.set_level(logging.DEBUG)
>       gurobi_model.solve()

tests/test_backend_pyomo.py:1689: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/calliope/model.py:470: in solve
    results = self.backend._solve(warmstart=warmstart, **solver_config)
src/calliope/backend/pyomo_backend_model.py:322: in _solve
    results = opt.solve(self._instance, tee=True, **solve_kwargs)
../../../micromamba/envs/calliope-ubuntu-latest-310-25fc8eb0047f7544a957ed75a82f2329ea5bc9f33ae1c087115cbb13d6ef0f33/lib/python3.10/site-packages/pyomo/solvers/plugins/solvers/direct_solver.py:75: in solve
    self.available(exception_flag=True)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pyomo.solvers.plugins.solvers.gurobi_direct.GurobiDirect object at 0x7fc2da95a9e0>
exception_flag = True

    def available(self, exception_flag=True):
        """Returns True if the solver is available.
    
        :param exception_flag: If True, raise an exception instead of returning
            False if the solver is unavailable (defaults to False)
        :type exception_flag: bool
    
        In general, ``available()`` does not need to be called by the user, as
        the check is run automatically when solving a model. However it is useful
        for a simple retry loop when using a shared Gurobi license::
    
            with SolverFactory('gurobi', solver_io='python') as opt:
                while not available(exception_flag=False):
                    time.sleep(1)
                opt.solve(model)
    
        """
        # First check gurobipy is imported
        if not gurobipy_available:
            if exception_flag:
                gurobipy.log_import_warning(logger=__name__)
>               raise ApplicationError(
                    "No Python bindings available for %s solver plugin" % (type(self),)
                )
E               pyomo.common.errors.ApplicationError: No Python bindings available for <class 'pyomo.solvers.plugins.solvers.gurobi_direct.GurobiDirect'> solver plugin

../../../micromamba/envs/calliope-ubuntu-latest-310-25fc8eb0047f7544a957ed75a82f2329ea5bc9f33ae1c087115cbb13d6ef0f33/lib/python3.10/site-packages/pyomo/solvers/plugins/solvers/gurobi_direct.py:215: ApplicationError
```

brynpickering:
Fixes #349 
Fixes #597 (by pinning pyomo)

* Duplicated pyomo backend interface as a pure `gurobipy` interface
* Added basic tests of the interface methods

How to use:

1. Install `gurobipy` into your calliope env: `mamba install gurobi::gurobi`
2. Build optimisation problem with gurobi backend: 
```python
import calliope

model = calliope.Model(...)
model.build(backend="gurobi")
model.solve()
```

NOTE: without a license, one can only work with _very_ small models. Our example models are already too big (maybe if run over ~2 timesteps it would work?).

Reviewer checklist:

- [x] Documentation updated
- [x] Changelog updated


brynpickering:
I _have_ added tests, but we don't have gurobipy installed in the CI builds so those tests are being skipped. Hence the massive coverage drop.

brynpickering:
I have run the urban scale model for a full year using the pyomo (with gurobi solver) and gurobi interfaces and get the attached results.

Headline: Gurobi interface peaks at ~30% less memory than pyomo.
I know you got different results when running locally @sjpfenninger. This could be because I'm using a cloud license for Gurobi which I _think_ means it offloads the actual solving onto a remote machine.

[memray_flamegraphs.zip](https://github.com/calliope-project/calliope/files/14310120/memray_flamegraphs.zip)

---
gurobi run:

```python
import calliope

calliope.set_log_verbosity("info")
m = calliope.examples.urban_scale(time_subset=None)
m.build(backend="gurobi")
m.solve(solver_options={"Threads": 1})
```

pyomo run:

```python
import calliope

calliope.set_log_verbosity("info")
m = calliope.examples.urban_scale(time_subset=None)
m.build(backend="pyomo")
m.solve(
solver="gurobi", solver_options={"Threads": 1})
```

Profiling using [memray](https://bloomberg.github.io/memray/) and setting [out] and [runfile] separately for pyomo and gurobi runs:

```shell
memray run --trace-python-allocators -o [out].bin [runfile].py    
memray flamegraph [out].bin -o [out].html    
```

brynpickering:
Last week's new pyomo release has caused an issue in one of our tests (see #597). I've pinned Pyomo upper bound to avoid this issue for now.

src/calliope/model.py
```diff
@@ -16,6 +16,7 @@
 from calliope._version import __version__
 from calliope.attrdict import AttrDict
 from calliope.backend import parsing
+from calliope.backend.gurobi_backend_model import GurobiBackendModel
```
irm-codebase, line 19: Will probably fail due to the new backend approach

tests/test_backend_general.py
```diff
@@ -0,0 +1,630 @@
...
+import numpy as np
+import pandas as pd
+import pytest  # noqa: F401
+import xarray as xr
+
+from .common.util import build_test_model as build_model
+from .common.util import check_error_or_warning
+
+
+@pytest.fixture(scope="class", params=["pyomo", "gurobi"])
```
irm-codebase, line 13: shouldn't the scope of these be "module"? Is there an advantage to "class" in this case?

tests/test_backend_general.py
```diff
@@ -0,0 +1,630 @@
...
+        "component_type", ["variable", "global_expression", "parameter", "constraint"]
+    )
+    def test_get_missing_component(self, solved_model_cls, component_type):
+        """Try and fail to get a component that doesn't exist in the backend dataset."""
+        with pytest.raises(KeyError):
+            getattr(solved_model_cls.backend, f"get_{component_type}")("foo")
+
+    def test_get_variable(self, solved_model_cls):
+        """Check a decision variable has all expected attributes."""
+        var = solved_model_cls.backend.get_variable("flow_cap")
```
irm-codebase, line 122: Nice-to-have: I'd make this a fixture and test for each requirement separately. Shorter tests are usually easier to maintain.

Ditto for other tests of this type in this file.

tests/test_backend_general.py
```diff
@@ -0,0 +1,630 @@
...
+            assert not (expression_string.str.contains("test_demand_elec").any())
+
+    @pytest.mark.usefixtures(
+        "built_model_func_longnames", "built_model_func_updated_cost_flow_cap"
+    )
+    @pytest.mark.parametrize("model_suffix", ["_longnames", "_updated_cost_flow_cap"])
+    def test_update_parameter_refs_in_obj_func(
+        self, request: pytest.FixtureRequest, model_suffix: str
+    ):
+        """Check that parameter re-definition propagates from global expressions to objective function."""
```
irm-codebase, line 528: Nice-to-have: I think only checking the updated version is sufficient here. The other seems a bit like "testing your own test". LMKWYT.

tests/test_backend_gurobi.py
```diff
@@ -0,0 +1,312 @@
+import importlib
+
+import calliope.exceptions as exceptions
+import pytest  # noqa: F401
+import xarray as xr
+
+from .common.util import build_test_model as build_model
+from .common.util import check_error_or_warning
+
+if importlib.util.find_spec("gurobipy") is not None:
```
irm-codebase, line 10: Shouldn't we just add the solver to our dev requirements to avoid this one?

Avoiding it for the base requirements makes sense. For dev, not so much, imo.

tests/test_backend_gurobi.py
```diff
@@ -0,0 +1,312 @@
...
+        fixed = simple_supply_gurobi.backend._apply_func(
+            self._is_fixed,
+            simple_supply_gurobi.backend.variables.flow_cap.notnull(),
+            1,
+            simple_supply_gurobi.backend.variables.flow_cap,
+        )
+        assert fixed.where(fixed.notnull()).all()
+
+        # reset
+        simple_supply_gurobi.build(backend="gurobi", force=True)
```
irm-codebase, line 174: Just to confirm:  is this because the fixture is shared across "class" tests?

irm-codebase: Also, maybe consider a function for these so that the lines are not repeated as much?

tests/test_core_model.py
```diff
@@ -318,3 +318,32 @@ def test_operate_timeseries(self, operate_model):
...
+
+    def test_solve_operate_not_allowed(self, simple_supply):
+        simple_supply.backend.inputs.attrs["config"]["build"]["mode"] = "operate"
+        simple_supply._model_data.attrs["allow_operate_mode"] = False
+
+        try:
+            with pytest.raises(calliope.exceptions.ModelError) as excinfo:
+                simple_supply.solve(force=True)
+            assert check_error_or_warning(excinfo, "Unable to run this model in op")
+        except AssertionError as e:
```
irm-codebase, line 343: The purpose of the `except` and `else` a bit hard to understand. Should these be their own tests?

irm-codebase (changes requested):
Fairly big review. I tried to be thorough since this is an important change.
"Nice to haves" are just that, consider ignoring them if they result in too much work.

Also, I expect that the recent changes to the backend handling and the linter will trigger errors once you try to merge this into main.

src/calliope/backend/expression_parser.py
```diff
@@ -379,28 +384,32 @@ def _eval(
...
 
     def as_array(self) -> xr.DataArray:  # noqa: D102, override
         lhs, rhs = self._eval("array")
-        return self.eval_attrs["backend_interface"]._apply_func(
-            self._compare_bitwise, self.eval_attrs["where_array"], lhs, rhs
-        )
+        where = self.eval_attrs["where_array"]
+        for side, arr in {"left": lhs, "right": rhs}.items():
+            extra_dims = set(arr.dims).difference(set(where.dims))
+            if extra_dims:
```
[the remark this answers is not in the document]
irm-codebase: This should be its own issue ( #633). Feel free to ignore!

requirements/base.txt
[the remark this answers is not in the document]
brynpickering: See #597

src/calliope/backend/backend_model.py
```diff
@@ -452,39 +451,64 @@ def _add_to_dataset(
...
+        where: Optional[xr.DataArray],
+        n_out: Literal[2, 3, 4, 5],
+        *args,
+        **kwargs,
+    ) -> tuple[xr.DataArray, ...]: ...
+
     def _apply_func(
-        self, func: Callable, *args, output_core_dims: tuple = ((),), **kwargs
-    ) -> xr.DataArray:
+        self, func: Callable, where: Optional[xr.DataArray], n_out: int, *args, **kwargs
```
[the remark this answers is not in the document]
brynpickering: Yeah, fair enough, it can always be reintroduced if we want to be able to apply a function without a where but right now all applications have a boolean where array associated with them

src/calliope/backend/expression_parser.py
```diff
@@ -347,14 +346,20 @@ class EvalComparisonOp(EvalToArrayStr):
 
     OP_TRANSLATOR = {"<=": r" \leq ", ">=": r" \geq ", "==": " = "}
 
-    def __init__(self, tokens: pp.ParseResults) -> None:
+    def __init__(self, instring: str, loc: int, tokens: pp.ParseResults) -> None:
```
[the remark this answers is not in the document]
brynpickering: It's a pyparsing logic anomoly. You can init a parsing class with one or three args and it will interpret them as `tokens` or [instring, loc, tokens] accordingly. `loc` is an int and we have ~100 comparisons opt classes created in a given model, so I expect the memory hit to negligble.

src/calliope/backend/gurobi_backend_model.py
```diff
@@ -0,0 +1,575 @@
...
+
+    def __init__(self, inputs: xr.Dataset, **kwargs) -> None:
+        """Gurobi solver interface class.
+
+        Args:
+            inputs (xr.Dataset): Calliope model data.
+            **kwargs: passed directly to the solver.
+        """
+        if importlib.util.find_spec("gurobipy") is None:
+            raise ImportError(
```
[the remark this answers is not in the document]
brynpickering: Hmm, backenderror seems to make sense for errors that we can't assign to standard builtins. This is clearly an importerror so I'm happy to use a builtin!

tests/test_backend_general.py
brynpickering: Some classes end up changing the backend model and I'd prefer that not to bleed between classes

tests/test_backend_general.py
brynpickering: hmm, I guess I want to catch false negatives in future here (if we give `test_demand_elec` a value for some reason, the `updated` check will pass but not necessarily because the functionality we're testing is passing.

tests/test_core_model.py
brynpickering: It was to handle updating a session-scoped fixture, but it isn't pretty. I'll update

brynpickering:
Thanks for the review @irm-codebase - updated!

tests/test_backend_gurobi.py
brynpickering: I tried, but it won't work. Our dependencies are currently pip and conda cross-compatible. On PyPI, this dependency is gurobipy, on Anaconda it is gurobi... So, this breaks our compatibility. I've still removed the check in the tests for gurobipy. I would prefer to alert the developer to the need to install gurobipy!

requirements/base.txt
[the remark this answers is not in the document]
irm-codebase: 🤦

tests/test_backend_gurobi.py
irm-codebase: Hmm... in the future we might want to switch to `pixi` to avoid this kind of stuff. It prefers `conda-forge` by default, but you can redirect dependencies in a case-by-case basis if needed...
I'm ok with leaving this as is for now though!

docs/creating/config.md
```diff
@@ -39,6 +39,16 @@ To test your model pipeline, `config.init.time_subset` is a good way to limit yo
 
 ## Deep-dive into some key configuration options
 
+### `config.build.backend`
+
+By default, the optimisation problem is built using the [pyomo][] library.
```
irm-codebase, line 44: Broken `pyomo` link?

src/calliope/backend/backend_model.py
```diff
@@ -923,8 +911,8 @@ def _rebuild_references(self, references: set[str]) -> None:
                 self.delete_component(ref, component)
                 getattr(self, "add_" + component.removesuffix("s"))(name=ref)
 
-    def _get_capacity_bound(
-        self, bound: Any, name: str, references: set, fill_na: Optional[float] = None
+    def _get_variable_bound(
```
irm-codebase, line 914: Just for clarity: this is _either_ a lower or upper bound, not both at the same time?
If that's the case, could you change the the docstring to reflect it (`...the upper or lower...`)?

Otherwise, `_bounds` might be better.

irm-codebase: I see a the latex model uses `_get_variable_bounds`, which is a tad confusing.

irm-codebase (commented):
Very happy with the changes!
I've added a couple of things to clarify. After that, I can approve.

irm-codebase (approved):
Great, thank you for the changes @brynpickering!

irm-codebase (dismissed):
I was going to send this PR back because pytest failed for some tests (due to some duplicated warning messages)... but re-making my `mamba` environment fixed this.

I think we should re-think our approach to dependencies, though, because following your instructions also leads to failures due to the lack of `cbc` solver, but that's been there for a while so it does not relate to this PR.

brynpickering:
We can't provide generalised instructions for cbc in the installation guide due to Windows users not being able to install it directly (and `pixi` wouldn't solve that). I would rather move to using `HiGHs` as our default open source solver, but I haven't got it to work with the pyomo kernel library...

docs/advanced/backend_choice.md
```diff
@@ -0,0 +1,19 @@
...
+On loading a model, there is no solver backend, only the input dataset.
+The backend is generated when a user calls `build()` on their model.
+By default this will call back to [Pyomo](https://www.pyomo.org/) to build the model and send it off to the solver given by the user in the run configuration `#!yaml config.solve.solver`.
+
+Pyomo has the benefit of _mutable_ input parameters, which allows you to [update parameter values in your optimisation problem](backend_interface.md) without having to re-build any Pyomo objects.
+However, it is otherwise a relatively memory and time-consuming library for building optimisation problems.
+
+Since larger models tend to require a commercial solver to successfully complete in a reasonable amount of time (see our [solver comparison for justification](solver.md)), we have also introduced a direct interface to the Gurobi solver Python API.
+Our tests show that this reduces peak memory consumption and time to solution compared to using the Pyomo backend with Gurobi as the solver in both cases.
+If you have access to a Gurobi License, this does not require any extra effort on your part.
```
sjpfenninger, line 13:
```suggestion
If you have access to a Gurobi license, this does not require any extra effort on your part, besides having to:
```

docs/advanced/backend_choice.md
```diff
@@ -0,0 +1,19 @@
...
+You just need to:
```
sjpfenninger, line 14:
```suggestion
```

docs/advanced/backend_choice.md
```diff
@@ -0,0 +1,19 @@
...
+
+1. Install the gurobi python library into your Calliope environment: `mamba install gurobi::gurobi`.
```
sjpfenninger, line 16:
```suggestion
1. Install the Gurobi Python library into your Calliope environment: `mamba install gurobi::gurobi`.
```

sjpfenninger (approved):
Since I had looked at this previously and Ivan did a detailed review I'm only making some minor suggestions for the documentation. Good to go!

PR #560, opened 2024-02-08
Update data source dtype coercion

brynpickering:
Fixes issue where data sources contain numeric data with separators.

E.g., `25,000` would be loaded by stored in CSV as "25,000" and pandas can't natively coerce that to a numeric dtype. 

* Data that should be numeric but is stored in the array as a string due to a separator will be coerced to a numeric value according to the "locale" of the user's machine. I believe this defaults to `,` being the thousand separator and `.` the decimal point. It would probably be feasible for a user to change the locale manually (
`locale.setlocale(..)`) if they had data using other separators.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/preprocess/data_sources.py
+++ b/src/calliope/preprocess/data_sources.py
@@ -29,6 +29,8 @@ LOGGER = logging.getLogger(__name__)
 
 locale.setlocale(locale.LC_ALL, "")
 
+exceptions.warn(f"Locale: {locale.getlocale()}")
+
 
 class DataSourceDict(TypedDict):
     rows: NotRequired[str | list[str]]
--- a/src/calliope/preprocess/data_sources.py
+++ b/src/calliope/preprocess/data_sources.py
@@ -27,7 +27,7 @@ from calliope.util.tools import listify, relative_path
 
 LOGGER = logging.getLogger(__name__)
 
-locale.setlocale(locale.LC_ALL, "en_GB.utf-8")
+locale.setlocale(locale.LC_ALL, "")
 
 
 class DataSourceDict(TypedDict):
--- a/src/calliope/preprocess/data_sources.py
+++ b/src/calliope/preprocess/data_sources.py
@@ -27,7 +27,7 @@ from calliope.util.tools import listify, relative_path
 
 LOGGER = logging.getLogger(__name__)
 
-locale.setlocale(locale.LC_ALL, "en_GB")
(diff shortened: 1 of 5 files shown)


src/calliope/preprocess/data_sources.py
```diff
@@ -23,7 +27,7 @@
 
 LOGGER = logging.getLogger(__name__)
 
-DTYPE_OPTIONS = {"str": str, "float": float}
+locale.setlocale(locale.LC_ALL, "en_GB.utf-8")
```
sjpfenninger, line 30: Contrary to the PR description, the locale is hardcoded. I think that's a bad idea and I think it might break on some platforms, e.g. Windows, where locale strings have a different format. Can we use a `locale` module independent approach to converting strings to numbers?

src/calliope/preprocess/time.py
```diff
@@ -157,7 +157,7 @@ def cluster(data: xr.Dataset, clustering_file: str | Path, time_format: str):
     """
     clustering_timeseries = pd.read_csv(clustering_file, index_col=0).squeeze()
     clustering_timeseries.index = _datetime_index(
-        clustering_timeseries.index, time_format
+        clustering_timeseries.index + " 00:00:00", time_format
```
sjpfenninger, line 160: How is this connected to the dtype coercion fix?

sjpfenninger (changes requested):

src/calliope/preprocess/time.py
brynpickering: Aha, I had wondered where these changes had got to... I'll split them out and into another PR

src/calliope/preprocess/data_sources.py
brynpickering: Reverted to using system locale

sjpfenninger (changes requested):
Still unsure about this. Either:

1. make the behavior hardcoded and 100% locale-independent, and document that, or
2. make it locale-dependent and warn whenever locale-dependent parsing happens (and again, document that this is locale-dependent so might lead to machine-specific effects).

I think 1 would lead to less potential confusion?

brynpickering:
Closing as, after some testing, `locale` setting turns out to not be that powerful. E.g., swiss locale can't handle `1'000.0` and german locale can't handle `1.000,0`. Users will just need to sanitise their Excel data before loading it into Calliope.

PR #559, opened 2024-02-06
Documentation updates

brynpickering:
Fixes #555 

## Changes made in this Pull Request

- rename "custom math" to "user-defined math" across the board.
- rename "inbuilt math" to "pre-defined math" across the board.
- 4-space -> 2-space indentation in all YAML snippets.
- Added graphic to help visualise how the math syntax gets converted to a matrix operation.
- Cleaned up some confusing text that wasn't correctly updated from old docs.
- Update latex math representation to improve readability (as per @irm-codebase's suggestion).

Reviewer checklist:

- [ ] Test(s) added to cover contribution
- [ ] Documentation updated
- [ ] Changelog updated
- [ ] Coverage maintained or improved

PR #563, opened 2024-02-09
Update math docs and handling of missing data

brynpickering:
Fixes issue mentioned in #559 w.r.t. information provided in the pre-defined math docs.

## Summary of changes in this pull request:

* Parameters are now included in the math docs directly, with cross-refs to the other math components in which they're used.
* Cross-refs have been extended to use of a constraint/variable in the `where` string of other math components, not only in the math expression.
* Variables/global expressions have `defaults` which are used to initialise empty array elements. This helps handle situations where a sparsely defined model would lead to NaN values that shouldn't make their way into the LP file. The values are set such that the variable/expression does not impact the optimisation problem if taking its "default" value.
* Updating a Pyomo parameter now looks at the use of the parameter in `where` strings to decide whether some variables/expressions/constraints need to be re-built. This should be expected functionality (i.e., if updating a parameter turns a NaN input value into a non-NaN value, then the `where` string should be re-evaluated).
* Parameter units moved to a user-defined schema attribute so we can load them into the math docs. This _does_ mean they are lost from the schema reference page (user-defined attributes aren't picked up there).
* Added extra info to variable/expression/parameter descriptions in the math docs. Namely, default values, unit, and cross-refs.
 
## TODO:

- [x] Either _none_ or only a _subset_ of parameters should show up in the "inter-cluster storage" math docs. Currently, all parameters relavant to base _and_ inter-cluster storage math are shown.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


irm-codebase:
Fantastic addition! Adding a list of the constraints where a parameter is used is a really good idea.
However, it seems like some math is broken in the new parameters section. Here's an example:
image

brynpickering:
OK great, that was a missing `$` in the string. If you find any others, do flag them!

docs/advanced/constraints.md
irm-codebase, line 1: Just me, but I think the added "which includes a description of our pre-defined parameters" is not really needed.
Most people would expect it to be there, I think, and it does not relate to the stuff discussed in this section of the docs.

Not the most important thing, though, so it can be kept if you wish.

docs/pre_defined_math/index.md
irm-codebase, line 1: Another nit-pick here is that this really seems like it should go at the top of the next section.
It is not clear what the `if` is referring to until you go to the next page.

irm-codebase (approved):

sjpfenninger (approved):

PR #574, opened 2024-02-22
Update dependencies

brynpickering:
Update to latest pandas, xarray, ruamel.yaml, and pytest.

Also gone through and updated our globally filtered warnings in pytest.

*
*
*

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated


requirements/base.txt
```diff
@@ -8,9 +8,9 @@ jsonschema >= 4, < 5
 natsort >= 8, < 9
 netcdf4 >= 1.2, < 1.7
 numpy >= 1, < 2
-pandas >= 2.1.3, < 2.2
+pandas >= 2.1.3, < 2.3
```
sjpfenninger, line 11: why change the maximum but not the minimum version? should this be `>= 2.2.1`?

sjpfenninger (changes requested):

requirements/base.txt
brynpickering: Generally, we pin projects at >= v1.0.0 to their current major version. I've set the minimum bound above this for pandas because they introduced a regression in v2.1.0 (or 2.1.1) that vastly inflated time/memory consumption on doing groupby operations with MultiIndex. So, now it's a matter of cautiously expanding pandas pinning not keeping it as restricted as it was.

sjpfenninger:
```suggestion
pandas >= 2.1.3, < 2.3  # Minimum bound is 2.1.3 because of a regression in v2.1.0/2.1.1 inflating time/memory consumption on groupby operations with MultiIndex
```

PR #575, opened 2024-02-27
Coerce model dimension dtypes to numeric

### LINKED ISSUE #567: Cannot define numeric index items when loading tabular data
### What happened?

If you have a data dimension that you want to be numeric (e.g. years) then you can't load it from file as pandas will load it from CSV as strings. If you load the same data from YAML the index will be numeric. If you have a bit of data in YAML and a bit in tabular data then you get duplicated index items, one being the string format and the other the numeric format.

## Why might we care?

If you have numeric dimensions, you can use them in the math. E.g., you can subtract technology lifetime from the investment year in pathway analysis. This isn't currently possible because the investment year is a timestamp and you can't subtract an integer (lifetime) from a timestamp.

## A few possible workarounds

1. If you want a numeric index, just use YAML
2. Use the YAML index dtype to coerce the tabular data index dtype to numeric. This works if you define a bit of data in YAML.
3. Always _try_ to coerce all tabular data dimensions to numeric dtype (int then float if that fails?). This works if you don't define any data in YAML but still expect numeric data in CSV to lead to numeric dimensions.

(3) would seem "safest" as it doesn't rely on any YAML having been defined. However, it could be a problem if a user wanted their numeric-like data to actually be strings.

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Version

v0.7.0.dev2

### Relevant log output

_No response_

brynpickering:
Fixes #567 

* Always try to convert data dimensions to numeric and keep any successful attempts. 
* Add some tests.
* Add mention of this in the docs.

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/preprocess/model_data.py
+++ b/src/calliope/preprocess/model_data.py
@@ -695,18 +695,23 @@ class ModelDataFactory:
 
         return link_tech_dict
 
-    def _add_to_dataset(self, to_add: xr.Dataset, id: str):
+    def _add_to_dataset(self, to_add: xr.Dataset, id_: str):
         """Add new data to the central class dataset.
 
         Before being added, any dimensions with the `steps` suffix will be cast to datetime dtype.
 
         Args:
             to_add (xr.Dataset): Dataset to merge into the central dataset.
-            id (str): ID of dataset being added, to use in log messages
+            id_ (str): ID of dataset being added, to use in log messages
         """
-        to_add = time.timeseries_to_datetime(to_add, self.config["time_format"], id)
+        to_add_numeric_dims = self._update_numeric_dims(to_add, id_)
+        to_add_numeric_ts_dims = time.timeseries_to_datetime(
+            to_add_numeric_dims, self.config["time_format"], id_
+        )
         self.dataset = xr.merge(
-            [to_add, self.dataset], combine_attrs="no_conflicts", compat="override"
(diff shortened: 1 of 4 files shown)


src/calliope/preprocess/model_data.py
```diff
@@ -747,6 +752,30 @@ def _update_one_way_links(node_from_data: dict, node_to_data: dict):
         node_from_data.pop("carrier_out")  # cannot import carriers at the `from` node
         node_to_data.pop("carrier_in")  # cannot export carrier at the `to` node
 
+    @staticmethod
+    def _update_numeric_dims(ds: xr.Dataset, id_: str) -> xr.Dataset:
+        """Try coercing all dimensions to numeric.
+        Any that don't raise an error will remain as numeric.
```
sjpfenninger, line 758:
```suggestion
        Any that don't raise an error will be made numeric.
        Any that do raise an error will not be touched.
```

sjpfenninger: Docstring could be a bit clearer I think

sjpfenninger (changes requested):

src/calliope/preprocess/model_data.py
sjpfenninger: Other than the minor comment above, good to merge

sjpfenninger (approved):

PR #578, opened 2024-02-28
Remove sorting on attrdict key nesting

### LINKED ISSUE #576: AttrDict sorts keys on union
### What happened?

This is intended behaviour, but it is silent and probably not what we actually want it to do:

Say we have:

```python
foo = AttrDict({"c": 1, "b": 2})
bar = AttrDict({"d": 10, "b": 2, "a": 20})
foo.union(bar, allow_override=True)
```

We get:
```python
{'c': 1, 'b': 2, 'a': 20, 'd': 10}
```

I would expect the result to actually be:

```python
{'c': 1, 'b': 2, 'd': 20, 'a': 10}
```

The reason the keys in `bar` are being sorted alphabetically is that in `AttrDict.union` `keys_nested` is being called, which sorts keys before returning them nested. `bar` then has its (sorted) keys iterated over to merge into `foo`, hence why the keys from `foo` are _not_ sorted.

To achieve what I think would be more intuitive (and more useful when adding user-defined YAML math) behaviour, we should remove key sorting when calling `keys_nested`.

- [X] macOS
- [ ] Windows
- [ ] Linux

### Version

v0.7.0.dev3

### Relevant log output

_No response_

brynpickering:
Fixes #576 

* Remove sorting on key nesting
* Add test

Reviewer checklist:

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/attrdict.py
+++ b/src/calliope/attrdict.py
@@ -355,7 +355,7 @@ class AttrDict(dict):
 
     def keys_nested(self, subkeys_as="list"):
         """
-        Returns all keys in the AttrDict, sorted, including the keys of
+        Returns all keys in the AttrDict, including the keys of
         nested subdicts (which may be either regular dicts or AttrDicts).
 
         If ``subkeys_as='list'`` (default), then a list of
@@ -366,7 +366,7 @@ class AttrDict(dict):
 
         """
         keys = []
-        for k, v in sorted(self.items()):
+        for k, v in self.items():
             # Check if dict instance (which AttrDict is too),
             # and for non-emptyness of the dict
             if isinstance(v, dict) and v:
--- a/tests/test_core_attrdict.py
+++ b/tests/test_core_attrdict.py
@@ -16,8 +16,8 @@ class TestAttrDict:
         d = {
             "a": 1,
             "b": 2,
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,16 @@ Both the index and the values of the timeseries (both being date strings) should
 
 |changed| `inbuilt` math -> `pre-defined` math and `custom` math -> `pre-defined` math in the documentation.
 
+|changed| Calliope attribute dictionaries (AttrDicts) no longer sort dictionary keys on `union`. Key order is now: original dictionary key order + any new keys being added in the order they appear in the new dictionary.
+
+### Internal changes
+
+|new| `py.typed` file so that mypy recognises Calliope as a typed library when it is imported as a dependency.
+
+|changed| GitHub action versions updated: `codecov/codecov-action` -> v4, `actions/checkout` -> v4, `dawidd6/action-download-artifact` -> v3.
```
sjpfenninger, line 31: Do we want CI changes in the changelog? We didn't in the past, but I realise some CI-related changes have been there since 0.7.0.dev2

sjpfenninger (approved):

CHANGELOG.md
brynpickering: Maybe we don't, you're right

PR #596, opened 2024-05-10
Force minimum of one header row; Add to docs

### LINKED ISSUE #573: Multiple rows and columns introduce false index items
### What happened?

Loading a file with multiple rows and multiple columns adds fake indexes sometimes.

See the file:
```csv
vintagesteps,2020,2030,2040,2050,2030,2040,2050,2040,2050,2050
investsteps,2020,2020,2020,2020,2030,2030,2030,2040,2040,2050
techs,,,,,,,,,,
geothermal,1,1,1,1,1,1,1,1,1,1
hydropower,1,1,1,1,1,1,1,1,1,1
waste,1,1,0.0,0,1,1,0.0,1,1,1
bioenergy,1,1,0.8,0,1,1,0.8,1,1,1
oil,1,1,1,1,1,1,1,1,1,1
coal,1,1,1,1,1,1,1,1,1,1
ccgt,1,1,0.0,0,1,1,0.0,1,1,1
wind,1,1,0.0,0,1,1,0.0,1,1,1
pv,1,1,1,0.2,1,1,1,1,1,1
battery_li,1,1,0.5,0,1,1,0.5,1,1,1
battery_phs,1,1,1,1,1,1,1,1,1,1
```

Loaded via:
```yaml
  vintage_availability_techs:
    source: data_sources/investstep_series/available_vintages_techs.csv
    rows: techs
    columns: [vintagesteps, investsteps]
    add_dimensions:
      parameters: available_vintages
```

In this case, a fake index called `techs` will be added.

- [ ] macOS
- [ ] Windows
- [X] Linux

### Version

v0.7

### Relevant log output

_No response_

brynpickering:
Fixes #573 

Based on the decision in #573, I have forced the need for a header row by setting `header=[0]` if a user does not provide `columns` data in their data source config.

It can lead to a difficult to understand failure. When a user forgets a header row, data ends up as the column / index level names. This gets picked up later when Calliope tries to match index/column levels to what has been defined in the config. That seems to be the only place we can catch it, but it leads to an error like this:

`calliope.exceptions.ModelError: (data_sources, ds_name) | Trying to set names for index but names in the file do no match names provided | in file: ['bar1', 'bar2'] | defined: ['test_row1', 'test_row2'].`

Maybe just adding to that error with something like "Make sure you have a header row in your CSV file"?

* Force header=[0] as a minimum
* Added to docs with a bit about how to structure the CSV
* Updated tests

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


docs/creating/data_sources.md
```diff
@@ -23,6 +23,90 @@ In addition, when loading from file, there is the _required_ dimension `paramete
 This is a placeholder to point Calliope to the parameter name(s) that your tabular data is referring to.
 The values assigned as parameters will become the array names in your loaded model dataset ([`model.inputs`][calliope.Model.inputs]).
 
+## Structuring your CSV files
+
+### Include a header row
+
+CSV files in which tabular data is stored _must_ have at least one header row.
+If you do not have this header row, then the first row of data will still be taken to be dimension names, which Calliope will complain about.
```
irm-codebase, line 31: Maybe specify _how_ calliope will complain about this. Error, Warning, unexpected behavior?

src/calliope/preprocess/data_sources.py
```diff
@@ -245,7 +245,8 @@ def _read_csv(self) -> pd.DataFrame:
             pd.DataFrame: Loaded data without any processing.
         """
         filename = self.input["source"]
-        csv_reader_kwargs = {"header": self.columns, "index_col": self.index}
+        header = [0] if self.columns is None else self.columns
```
irm-codebase, line 248: I am a bit confused by your comment on this line (in your  Would this generate a weird error for users in certain situations, or avoid it?

Since Calliope making an assumption here, maybe just throw a warning?

irm-codebase (changes requested):

src/calliope/preprocess/data_sources.py
irm-codebase: Honestly, I think the most robust approach is to just ask users to fully fill it by default, and to ensure it has no numbers in it on our end if we set it to 0. It reduces the number of test cases on our end, leads to better documentation, and may avoid headaches in the future. 

Unless I am missing something?

brynpickering: Sadly, there are legitimate reasons to have your dimensions or their values be numeric (e.g. years of data), so we can't set that strict rule.

brynpickering: > I am a bit confused by your comment on this line (in your Would this generate a weird error for users in certain situations, or avoid it?

The issue with pandas Series/DataFrames and their CSV representation is that column names and MultiIndex index levels are stored on the same CSV row. It gets even weirder with MultiIndex columns. If you don't define any dimension linked to your non-index data, you still have a header row in the CSV in which your `row` dimensions are stored (MultiIndex levels). Since we're forcing that a user explicitly defines in their CSV header any dimensions that are in `rows`/`columns` in their data source definition, we always take it that there is at least one header row. If there are multiple `columns` dimensions then those will be inferred from the list the user provides.

irm-codebase: Go it. This one is hard for sure. I am satisfied with the improved error handling.
We can revisit this if it's causing too many issues, although I am not sure how to solve it.

irm-codebase (approved):

PR #603, opened 2024-06-11
Add `title` to math var and expr components

### LINKED ISSUE #582: Add `title` or `long-name` parameter to math syntax
### What can be improved?

This would add a longer name to math components that can be extracted and used in downstream visualisation. E.g.,

```yaml
variables:
  flow_cap:
    title: Flow capacity
    description: >-
      A technology's flow capacity, also known as its nominal or nameplate capacity.
    default: 0
    unit: power
    foreach: [nodes, techs, carriers]
    bounds:
      min: flow_cap_min
      max: flow_cap_max

```

### Version

v0.7.0.dev

brynpickering:
Fixes #582 

Chosen "title" as it fits with the YAML/JSON schema syntax.

* Variables and global expressions get a "title" key that will be stored in the attributes of the result dataarrays
* Parameters already have a "title" key in the schema, this is also now passed to the attributes of the input dataarrays

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/tests/test_backend_pyomo.py
+++ b/tests/test_backend_pyomo.py
@@ -1800,17 +1800,16 @@ class TestNewBackend:
         assert (
             var.to_series().dropna().apply(lambda x: isinstance(x, pmo.variable)).all()
         )
-        expected_keys = set(
-            [
-                "obj_type",
-                "references",
-                "description",
-                "unit",
-                "default",
-                "yaml_snippet",
-                "coords_in_name",
-            ]
-        )
+        expected_keys = {
+            "obj_type",
+            "references",
+            "title",
+            "description",
+            "unit",
+            "default",
+            "yaml_snippet",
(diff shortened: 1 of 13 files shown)


tests/test_backend_pyomo.py
irm-codebase, line 1: Consider adding a short description to the tests describing their goal.

tests/test_core_model.py
```diff
@@ -10,6 +10,8 @@
 
 LOGGER = "calliope.model"
 
+DUMMY_INT = 0xDEADBEEF
```
irm-codebase, line 13: Maybe we should consider moving this to test utilities?

irm-codebase (approved):
Great update! The added features and the code look good overall.
I just added a couple of suggestions to the test side.

PR #609, opened 2024-06-19
Allowing complete math overrides

### LINKED ISSUE #606: Feature: allow for complete math overrides
### What can be improved?

As of now, builds custom math "on top" of the pre-defined math.

While this is good to simplify the way users interact with Calliope, it introduces problems in cases were the math needs to be heavily modified since users have to essentially look in two different places and are limited by what we put in the base math.

I'd like to propose a simple override were users introduce the "base" math file themselves (with no guarantees from our side).

This would make the library much more flexible, and may even allow others to discover improvements to our base math.

### Version

v0.7

irm-codebase:
Fixes #606 

* Introduces a new parameter (`config.init.base_math`) that will tell the model to skip base math if set to `False`.
* Run mode math updates have been moved from the backend to the main model file.
* Math is now always validated when modified (either in `init` or `build`). This is to prevent unexpected surprises if users save an initialized model without building it.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,7 @@ exclude = '''
 
 [tool.ruff]
 line-length = 88
-select = ["E", "F", "I", "Q", "W"]
+select = ["E", "F", "I", "Q", "W", "PT"]
 # line too long; Black will handle these
 ignore = ["E501"]
 
--- a/src/calliope/backend/backend_model.py
+++ b/src/calliope/backend/backend_model.py
@@ -7,7 +7,6 @@ Methods to interface with the optimisation problem
 
 from __future__ import annotations
 
-import importlib
 import logging
 import time
 import typing
@@ -36,11 +35,9 @@ from calliope.backend import helper_functions, parsing
 from calliope.exceptions import warn as model_warn
 from calliope.io import load_config
 from calliope.util.schema import (
-    MATH_SCHEMA,
     MODEL_SCHEMA,
     extract_from_schema,
     update_then_validate_config,
-    validate_dict,
 )
 
 if TYPE_CHECKING:
@@ -211,7 +208,6 @@ class BackendModelGenerator(ABC):
         )
 
     def _build(self) -> None:
-        self._add_run_mode_math()
         # The order of adding components matters!
         # 1. Variables, 2. Global Expressions, 3. Constraints, 4. Objectives
         for components in [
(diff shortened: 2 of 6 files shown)


src/calliope/backend/backend_model.py
irm-codebase, line 1: Moved this functionality to `model.py` because it does not really interact with the backend in any way.
Keeps the code a bit leaner.

pyproject.toml
irm-codebase, line 1: Enabled some extra linting help for `pytest`.
It already helped in removing some small issues in some files for this PR.

src/calliope/config/config_schema.yaml
irm-codebase, line 1: New boolean to specify if the base math is enabled in the model. The core of this PR.

src/calliope/model.py
```diff
@@ -188,9 +188,6 @@ def _init_from_model_def_dict(
         model_config.union(model_definition.pop("config"), allow_override=True)
 
         init_config = update_then_validate_config("init", model_config)
-        # We won't store `init` in `self.config`, so we pop it out now.
```
irm-codebase, line 191: I am now allowing `config.init` to remain in `model._model_data.attrs`.
I think it makes the model creation more transparent, particularly because the added math files specified are kept.

src/calliope/model.py
```diff
@@ -223,21 +220,25 @@ def _init_from_model_def_dict(
...
-        self._model_data.attrs["math"] = self._add_math(init_config["add_math"])
 
         log_time(
             LOGGER,
             self._timings,
             "model_data_creation",
             comment="Model: preprocessing stage 2 (model_data)",
         )
+
+        math, applied_math = self._math_init_from_yaml(init_config)
```
irm-codebase, line 232: The new `_math`  functions follow a functional programming approach for two reasons:
- easier to write tests for this style
- is explicit

I can change this to an implicit OOP approach if need-be.

src/calliope/model.py
```diff
@@ -302,26 +302,62 @@ def _add_math(self, add_math: list) -> AttrDict:
...
-            base_math.union(override_dict, allow_override=True)
         if file_errors:
             raise exceptions.ModelError(
                 f"Attempted to load additional math that does not exist: {file_errors}"
             )
-        self._model_data.attrs["applied_additional_math"] = add_math
-        return base_math
+        return (math, applied_math)
+
+    def _math_update_with_mode(
```
irm-codebase, line 313: This is the function I removed from the backend. Works in the same way, except that it returns the updated math instead of directly modifying it.

Also, I made sure to avoid the use of `.copy()` unless absolutely needed. Should be more efficient in cases were no modification is necessary.

tests/test_backend_pyomo.py
irm-codebase, line 1: Most of this file is linter updates. Run mode tests were moved to a different test file for consistency.

tests/test_core_model.py
```diff
@@ -30,48 +42,43 @@ def test_info_simple_model(self, simple_supply):
         simple_supply.info()
 
 
-class TestAddMath:
-    @pytest.fixture
-    def storage_inter_cluster(self):
-        return build_model(
-            {"config.init.add_math": ["storage_inter_cluster"]},
-            "simple_supply,two_hours,investment_costs",
+class TestInitMath:
```
irm-codebase, line 45: Math tests have been subdivided into 3 groups for clarity:
- Testing the math init function directly.
- Testing the math update function directly.
- Testing overrides from Model() to the math.

irm-codebase (commented):
@brynpickering a few comments to aid with the review.

Let me know what you think!

irm-codebase:
Did some extra digging around. The core of the issue is that the netCDF standard does not really consider attributes as anything more than single values per key (https://cfconventions.org/cf-conventions/cf-conventions.html#_attributes).

`xarray` does not support saving data in the attributes of datasets beyond single values either, which is why we have our own serialization.

I'm playing around with using pre-defined encoders for this, but so far the produced array is too big for `xarray`...
The encoding works perfectly fine, but the saving step does not.

Here's an example:

```python
import codecs
import pickle

import calliope

model = calliope.examples.national_scale()
attrs_orignial = model._model_data.attrs.copy()

pickled_attrs = codecs.encode(pickle.dumps(attrs_orignial), "base64").decode()
model._model_data.attrs.clear()
model._model_data.attrs["encoded"] = pickled_attrs
unpickled_attrs = pickle.loads(codecs.decode(pickled_attrs.encode(), "base64"))

assert attrs_orignial == unpickled_attrs
```

irm-codebase:
Converting to draft because #614 is a blocker to this PR.

PR #610, opened 2024-06-19
Fix sync issues in model math / config / defaults attributes

### LINKED ISSUE #608: Potential desync in model configuration at the model.py level
### What happened?

Calliope v0.7 keeps most of its attributes within `model._model_data.attrs[]` in the form of dictionaries.

However, model math and config are replicated in `model.math` and `model.config`. This introduces potential sync issues, necessitating some helper functions to keep these up-to-date. This is done via the `_add_observed_dict()` and `_add_model_data_methods()`.

But this requires careful use of these functions in several parts of the code, and is a bit inefficient (we essentially duplicate the same model property).

- [ ] macOS
- [ ] Windows
- [x] Linux

### Version

v0.7

### Relevant log output

_No response_

irm-codebase:
Fixes #608 (and is a prerequisite for #606).

This PR is meant to fix potential issues due to duplication of model data at the model.attribute level and at the model._model_data.attrs level.

Instead of copying stuff, class @properties are used to call and modify specific things (math, config, defaults).
model.attributes should be used for data we want to lose between runs (instance-specific timestamps, temp flags, etc).

By drawing this distinction, the code should become easier to maintain down the line.

* Turned `model.math`, `model.config` and `model.defaults` into properties that refer to `model._model_data.attrs` directly to avoid double instancing and potential desyncs
* Removed depreciated methods to sync `model.math` and `model.config` with values in `model._model_data`.
* Updated some tests.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/model.py
+++ b/src/calliope/model.py
@@ -93,9 +93,6 @@ class Model(object):
                 Defaults to None.
         """
         self._timings: dict = {}
-        self.config: AttrDict
-        self.defaults: AttrDict
-        self.math: AttrDict
         self._model_def_path: Optional[Path]
         self.backend: BackendModel
         self.math_documentation = MathDocumentation()
@@ -143,6 +140,18 @@ class Model(object):
         return self._model_data.filter_by_attrs(is_result=1)
 
     @property
+    def config(self):
+        return self._model_data.attrs["config"]
+
+    @property
+    def defaults(self):
+        return self._model_data.attrs["defaults"]
+
+    @property
+    def math(self):
+        return self._model_data.attrs["math"]
+
+    @property
     def is_built(self):
         return self._is_built
 
@@ -172,8 +181,8 @@ class Model(object):
         log_time(
             LOGGER,
             self._timings,
-            "model_run_creation",
-            comment="Model: preprocessing stage 1 (model_run)",
+            "model_definition_creation",
+            comment="Model: preprocessing stage 1 (model_definition)",
         )
         model_config = AttrDict(extract_from_schema(CONFIG_SCHEMA, "default"))
         model_config.union(model_definition.pop("config"), allow_override=True)
@@ -194,8 +203,10 @@ class Model(object):
             "applied_overrides": applied_overrides,
             "scenario": scenario,
             "defaults": param_metadata["default"],
+            "name": init_config["name"],
+            "config": model_config,
+            "math": None,
         }
-
         data_sources = [
             DataSource(
                 init_config,
@@ -208,13 +219,13 @@ class Model(object):
                 "data_sources", {}
             ).items()
         ]
-
         model_data_factory = ModelDataFactory(
             init_config, model_definition, data_sources, attributes, param_metadata
         )
         model_data_factory.build()
 
         self._model_data = model_data_factory.dataset
+        self._model_data.attrs["math"] = self._add_math(init_config["add_math"])
 
         log_time(
             LOGGER,
@@ -222,13 +233,6 @@ class Model(object):
             "model_data_creation",
             comment="Model: preprocessing stage 2 (model_data)",
         )
-
-        self._add_observed_dict("config", model_config)
-
-        math = self._add_math(init_config["add_math"])
-        self._add_observed_dict("math", math)
-
-        self._model_data.attrs["name"] = init_config["name"]
         log_time(
             LOGGER,
             self._timings,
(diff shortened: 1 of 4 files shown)


irm-codebase:
@sjpfenninger @brynpickering 
Here is a proposed improvement to avoid sync issues in the `model` object and its underlying `xarray` data.

I tried to update relevant tests, let me know if I missed something.

src/calliope/model.py
```diff
@@ -93,9 +93,6 @@ def __init__(
                 Defaults to None.
         """
         self._timings: dict = {}
-        self.config: AttrDict
-        self.defaults: AttrDict
```
irm-codebase, line 97: These are no longer necessary since they are now `@properties`

irm-codebase:
Changed this PR to merge directly into main instead of the schema override feature.
I figured that getting this in will benefit other things beyond that.

irm-codebase:
My initial setup for testing the new linked properties was ugly and hard to maintain. I improved it following better testing practices.

brynpickering:
I'd prefer just two ways to define the config/defaults/math for a model: 1. in the YAML files, 2. at the point of method calls (`Model(**kwargs)`, `model.build(**kwargs)`,  `model.solve(**kwargs)`). `Model.config`, `Model.defaults`, `Model.math` are then just handy references for the current state of the configuration.

irm-codebase:
@brynpickering :+1: 
Alright, I'll try to summarize the points. For what I understand, there are two issues
- Issue 1 is whether or not we want `model._model_data` to be frozen. Particularly in these situations
    - Should `kwargs` be saved to it, or only affect the object temporarily?
    - Should we offer interfaces to modify this beyond YAML files and `kwargs`?
- Issue 2 (this PR) is that `model.attrs` have inconsistent access to `model._model_data.attrs`. Python works with pointers, so in most cases modifying `model.config` will also modify `model._model_data.attrs["config"]`... but in our current approach this is not guaranteed!

My fix was not meant to be user-facing. It was meant to streamline the way we access data within our code. But I see why it could be seen that way...

Here are my proposed fixes to this PR in order to address Issue 2:
- remove `model.defaults`: I included it because it was declared in `model.__init__`, but unless we ensure updates propagate it should definitely not be there.
- alter `model.math` to `model._math`, which will have a setter and a getter.
- ditto for `model.config`
- modify our code accordingly

brynpickering:
I don't think they should be private properties (hence why they are public right now), but they should be _immutable_. Providing a setter goes against that idea as you essentially say "you can replace this entire dictionary if you like". It probably just makes sense to make the change you've implemented except for the setters. I.e., you don't provide a way for a user to completely replace those properties, although we still have no way of stopping them replacing the equivalent dictionary in `model._model_data.attrs`, which would have the same effect.

brynpickering:
Options for making these properties read-only:

1. Use something like [frozendict](https://github.com/Marco-Sulla/python-frozendict) to make the dictonaries linked to the properties un-settable.
2. pretty-print the dictionary instead of providing as a dictionary object (e.g. `print(yaml.dump(m.config.init.as_dict()))`).
3. Add some switch in AttrDict to allow freezing (i.e. just suppressing the dictionary setter in there somehow).
4. Turning the AttrDict into a set of nested `ModelConfig` objects that you can _get_ (`model.config.init.name`), but you can't _set_

irm-codebase:
Python handles this automatically by creating a getter, but no setter. We already do this for `model.is_solved`.
I plan to follow a similar approach in the re-try.

irm-codebase:
Retry!
This time I kept things simple: I've added getter `@property` that will let users see model math, defaults and configuration, **but not modify them**.

In theory, we could define "protected" `@property` version of these to make our code leaner (i.e., not always having to use `self._model_data.attrs["yaddayadda"]` for everything. For example: `model._math` could have a setter and a getter, while `model.math` only has a getter.

I've decided against this to keep the PR simple, and because it's a nice-to-have. Let me know if you want this included.

irm-codebase:
@brynpickering converting this back to draft because #617 was identified and relates to this fix directly.

irm-codebase:
This PR is related to #617 too. Unfortunately, fixing that is currently breaking operate mode. It's likely that fixing that issue involves too many changes, so this PR will be kept as-is to make the review easier.

irm-codebase:
I've just realized that we are putting the model configuration in ANOTHER attribute: `model._model_def_dict`

This PR is going back to the drawing board, unfortunately.

irm-codebase:
This PR is replaced by #625.

PR #611, opened 2024-06-20
Purge spurious debug parameter

irm-codebase:
Fixes #607

## Summary of changes in this pull request

Just removed a parameter that is no longer used from the Calliope `__init__` function.
Also updated the docstring from v0.6 to 0.7

## Reviewer checklist

- [ ] Test(s) added to cover contribution
- [ ] Documentation updated
- [ ] Changelog updated
- [ ] Coverage maintained or improved

Issue #607, opened 2024-06-18
debug functionality missing in Model.__init__


irm-codebase:
### What happened?

Calliope's `Model`  class initialization has a debug parameter with the following description:
```
            debug (bool, optional):
                If True, additional debug data will be included in the built model.
                Defaults to False.
```
At the moment, this parameter is completely unused by the code.

We check for `model_data.attrs["_debug_data"]` in other parts of the code, and for now it seems like this functionality won't work at all.

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [X] Linux

### Version

v0.7

brynpickering:
Yeah, we need to purge that. Debug info is all available via the logger (and you can store that in a file following our tutorial: https://calliope.readthedocs.io/en/latest/examples/calliope_logging/)

PR #620, opened 2024-06-25
Feature better linting

irm-codebase:
#607 / #611 mostly happened because inputs and docstrings were not being evaluated by the linter.

I propose to activate additional linting. Docstring linting tends to be annoying, but in our case I think it'll help.
It's very likely that this will trigger some hassle when modifying code at the start, but in the long run it will make our lives easier.

Additional linting is not always well received, though. Please let me know what you think!

*  I am activating two additional tests: "D" ([docstyle](https://docs.astral.sh/ruff/rules/#pydocstyle-d)) and "PT" ([pytest-style](https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt)).
* Added editor config for additional consistency in .YAML files

## Reviewer checklist

- [ ] Documentation updated
- [x] Changelog updated

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,25 @@
+# Documentation: https://EditorConfig.org
+# Inspired by Django .editorconfig file
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+
+[*.{yaml,yml}]
+indent_size = 2
+
+[*.md]
+trim_trailing_whitespace = true
+indent_size = 4
+
+[LICENSE]
+insert_final_newline = false
+
+[*.{diff,patch}]
+trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -20,6 +20,3 @@ indent_size = 4
 
 [LICENSE]
 insert_final_newline = false
-
-[*.{diff,patch}]
-trim_trailing_whitespace = false
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,16 +2,12 @@ default_language_version:
   python: python3
 
 repos:
-  - repo: https://github.com/psf/black
-    rev: 24.3.0
-    hooks:
-      - id: black
-
   - repo: https://github.com/astral-sh/ruff-pre-commit # https://beta.ruff.rs/docs/usage/#github-action
-    rev: v0.3.5
+    rev: v0.4.10
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
+      - id: ruff-format
 
   - repo: https://github.com/codespell-project/codespell
     rev: v2.2.6
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,7 @@ exclude = '''
 
 [tool.ruff]
 line-length = 88
-select = ["E", "F", "I", "Q", "W"]
+select = ["E", "F", "I", "Q", "W", "D", "PT"]
 # line too long; Black will handle these
 ignore = ["E501"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,31 +55,32 @@ exclude = '''
 
 [tool.ruff]
 line-length = 88
+[tool.ruff.lint]
 select = ["E", "F", "I", "Q", "W", "D", "PT"]
 # line too long; Black will handle these
 ignore = ["E501"]
 
-[tool.ruff.mccabe]
+[tool.ruff.lint.mccabe]
 # Unlike Flake8, default to a complexity level of 10.
 max-complexity = 10
 
 # Ignore `E402` (import violations) and `F401` (unused imports) in all `__init__.py` files
-[tool.ruff.per-file-ignores]
+[tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["E402", "F401"]
 "*.ipynb" = ["E402"]
 
-[tool.ruff.flake8-quotes]
+[tool.ruff.lint.flake8-quotes]
 docstring-quotes = "double"
 
-[tool.ruff.pydocstyle]
+[tool.ruff.lint.pydocstyle]
 convention = "google"
 
-[tool.ruff.pycodestyle]
+[tool.ruff.lint.pycodestyle]
 max-doc-length = 200
 ignore-overlong-task-comments = true
 
 [tool.codespell]
-skip = 'tests/**/*,**/*.ipynb,doc/**/*'
+skip = 'tests/**/*,**/*.ipynb,doc/**/*,AUTHORS'
 count = ''
 quiet-level = 3
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ exclude = '''
 [tool.ruff]
 line-length = 88
 [tool.ruff.lint]
-select = ["E", "F", "I", "Q", "W", "D", "PT"]
+select = ["E", "F", "I", "Q", "W", "PT"]
 # line too long; Black will handle these
 ignore = ["E501"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ exclude = '''
 [tool.ruff]
 line-length = 88
 [tool.ruff.lint]
-select = ["E", "F", "I", "Q", "W", "PT"]
+select = ["E", "F", "I", "Q", "W", "D", "PT"]
 # line too long; Black will handle these
 ignore = ["E501"]
 
@@ -68,6 +68,8 @@ max-complexity = 10
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["E402", "F401"]
 "*.ipynb" = ["E402"]
+"tests/*" = ["D"]
+"docs/examples/*" = ["D"]
 
 [tool.ruff.lint.flake8-quotes]
 docstring-quotes = "double"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,31 +30,12 @@ directory = "reports/coverage"
 [tool.coverage.xml]
 output = "reports/coverage/coverage.xml"
 
-[tool.black]
-line-length = 88
-skip-magic-trailing-comma = true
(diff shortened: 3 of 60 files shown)


irm-codebase:
`pre-commit` breaking is expected. I will fix these issues before merging this PR, but only after confirming these new lints are desired.

.editorconfig
```diff
@@ -0,0 +1,25 @@
...
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+
+[*.{yaml,yml}]
+indent_size = 2
+
+[*.md]
+trim_trailing_whitespace = true
+indent_size = 4
```
brynpickering, line 19: how does this work when doing two-line indenting inside yaml code blocks in the docs?

.editorconfig
```diff
@@ -0,0 +1,25 @@
...
+
+[LICENSE]
+insert_final_newline = false
+
+[*.{diff,patch}]
```
brynpickering, line 24: are these files even tracked? I assume they'd all be in the `.gitignore`

brynpickering (commented):
If you're happy to fix the issues raised by `precommit.ci` then I'm strongly in favour of this!

.editorconfig
irm-codebase: It will still apply 4. I'll update `.md` files to use 2 by default. Honestly not that big of a problem since most markdown previews do not need the 4 spaces.

.editorconfig
irm-codebase: They are not in the `gitingore`... but we do not even use them. I'll just remove this.

irm-codebase:
> If you're happy to fix the issues raised by `precommit.ci` then I'm strongly in favour of this!
Happy is a _strong_ word for linting fixes... but I will do it hahaha.

.editorconfig
brynpickering: no, that won't work either. 4 lines is required for some mkdocs features to work

irm-codebase: .editorconfig is not automatic in that sense. It only alters the default behavior of pressing "tab" in the editor. It is not going to modify files.

There are 2 options:
- Setting 4 then modifying to 2.
- Setting 2 and being aware of the 4 cases.

Let's go with whichever is more generic.

brynpickering: Ok, go with 4

brynpickering:
@irm-codebase don't forget to add yourself to the list in `AUTHORS` during this PR

irm-codebase:
@brynpickering good chuck of files done. Most of these are minor changes.
I am not completely re-working the documentation (I think it's better/safer to do that as we go on).

Question though: is it possible to configure pre-commit to only run "D" (pydocstrings) for merges?
I think it makes sense to allow people to develop their code with docstring flexiblity, and only enforce it before bringing it into main.

I've read a bit of pre-commit's configuration, and it seems possible, but I am not experienced enough with setting hooks to know how to achieve it.

irm-codebase:
This additional setting in `pr-ci.yml` could do the trick, but it depends on whether or not we want people to be pestered by docstrings during regular commits, or have them be surprised by extra changes in the PR.

```yaml
  ruff-docstring:
    runs-on: ubuntu-latest
    if: startsWith(github.event.pull_request.title, 'Release v')
    steps:
      - uses: chartboost/ruff-action@v1
        with:
          args: check --select D
```

brynpickering:
I think I prefer it being there by default. Hopefully shouldn't be too much effort to keep pre-commit / ruff happy with additions once you have the fixes done in this PR. It instills good documenting practice from our development team, which is no bad thing ;)

.pre-commit-config.yaml
irm-codebase, line 1: As requested, no more black. I also updated the version of Ruff in pre-commit.

src/calliope/backend/helper_functions.py
irm-codebase, line 1: Please go over this file. I am a tad worried about my lack of familiarity with the parser, and the naming was a bit hard to follow.

pyproject.toml
irm-codebase, line 1: Removed black and replaced with ruff format.
I only included checks for files not in the Ruff defaults, to keep the file lean.

I also _removed_ the magic comma skipping because it conflicted with our current isort (I) configuration!

pyproject.toml
```diff
@@ -30,56 +30,40 @@ directory = "reports/coverage"
...
 # Unlike Flake8, default to a complexity level of 10.
 max-complexity = 10
 
 # Ignore `E402` (import violations) and `F401` (unused imports) in all `__init__.py` files
-[tool.ruff.per-file-ignores]
+[tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["E402", "F401"]
 "*.ipynb" = ["E402"]
+"tests/*" = ["D"]
+"docs/examples/*" = ["D"]
```
irm-codebase, line 53: I am skipping the docs because the added markdown makes docstrings redundant.

In the case of tests, because he have more than a thousand and the tester should be the one adding it!
I suggest that, from now on, every new/modified test should have a short one liner describing its goal with an imperative.

e.g.`"""Sorting must be lower to upper by default"""`

irm-codebase (commented):
Ready for review @brynpickering !

Again, this is just a start to good documentation. I tried to be light on what I added because I was not the main developer. If you have the time, put special attention to my docstrings in the parser functionality. That's the only one I am worried about.

src/calliope/backend/helper_functions.py
```diff
@@ -24,6 +23,8 @@
 
 
 class ParsingHelperFunction(ABC):
+    """Helper class for parsing."""
```
brynpickering, line 26:
```suggestion
    """Abstract base class for helper function parsing."""
```

src/calliope/backend/latex_backend_model.py
```diff
@@ -362,7 +374,7 @@ def add_parameter(
         default: Any = np.nan,
         use_inf_as_na: bool = False,
     ) -> None:
-
+        """Add parameter to docs."""
```
brynpickering, line 377: These (and below) shouldn't be needed. They inherit their docs from the abstract base class. By adding this line in, I expect the inherited docsstrings to be overwritten, which is not ideal as those are quite verbose. If ruff can't handle this then we need some way to deal with it that doesn't involve overwriting what's in the ABC

src/calliope/backend/parsing.py
```diff
@@ -762,21 +789,26 @@ def generate_top_level_where_array(
         *,
         align_to_foreach_sets: bool = True,
         break_early: bool = True,
-        references: Optional[set] = None,
+        references: set | None = None,
```
brynpickering, line 792: Is it now recommended to _not_ use `Optional`?

src/calliope/backend/where_parser.py
```diff
@@ -282,42 +304,49 @@ def __init__(self, instring: str, loc: int, tokens: pp.ParseResults) -> None:
...
         return f"SUBSET:{self.set_name}{self.val}"
 
     def _eval(self) -> list[str | float]:
-        "Evaluate each element of the subset list"
+        """Evaluate each element of the subset list."""
         values = [val.eval("array", **self.eval_attrs) for val in self.val]
         return [val.item() if isinstance(val, xr.DataArray) else val for val in values]
 
     def as_math_string(self) -> str:
+        """Process and return as LaTeX."""
```
brynpickering, line 316: As with the backend models, these docstrings are inherited from the ABC so they shouldn't need to be repeated ad nauseam. We need a solution for this on the linting side

src/calliope/preprocess/load.py
```diff
@@ -42,12 +34,13 @@ def load_model_definition(
...
             These will be applied _after_ `scenario` overrides.
             Defaults to None.
+        **kwargs: initialisation overrides.
 
     Returns:
         tuple[AttrDict, Optional[Path], str]:
-            1. Model definition with overrides applied.
-            1. Path to model definition YAML if input `model_definiton` was pathlike, otherwise None.
-            1. Expansion of scenarios (which are references to model overrides) into a list of named override(s) that have been applied.
+            - Model definition with overrides applied.
```
brynpickering, line 41: I previously had it numbered so that it would be obvious the order of items in the tuple, is bullet points recommended instead?

brynpickering (commented):
Thanks for the effort @irm-codebase !

RE helper functions, we have a bit more detail on each [in the docs](https://calliope.readthedocs.io/en/latest/user_defined_math/syntax/#helper-functions). Could you take snippets from there to use for the introduction docstring for each class?

Then the only major thing is the requirement for docstrings in overrides to abstract methods. We really don't want to have to repeat a docstring that is already given in full by the ABC.

src/calliope/preprocess/load.py
irm-codebase: Oh, I'll re introduce them. They all had 1 so I thought it was fine.

src/calliope/backend/where_parser.py
irm-codebase: Gah, you are right. I wish we could just use `@override` (https://docs.python.org/3/library/typing.html#typing.override), then the linter would catch these cases.

Unfortunately it would break compatibility with 3.11 and lower :(

irm-codebase: I'll just add   # noqa: D102, override. Already did it for other functions, it'll be enough for now.

irm-codebase:
@brynpickering corrections implemented!
Sadly the only way to avoid the `@abstractmethod` issue is to deactivate linting for those functions.
We could implement the `@override` decorator to avoid this, but it's python 3.12 exclusive, and will break our tests for 3.10 / 3.11.

irm-codebase:
No idea why codeconv does not like this update, there is nothing added to the code :shrug:

brynpickering:
> edit: it's because of the added ... in some @abstractmethod cases. But without it we get linter issues.

Yeah, I think that's why I had docsstrings in them to begin with, to keep codecov happy. I'm not sure there's much to be done about it (or, at least, I can't find an obvious answer online).

src/calliope/backend/helper_functions.py
brynpickering: Updated

irm-codebase:
Thanks for the help @brynpickering! I went through your changes and I agree to them.
Is there anything else left to do for this PR?

brynpickering (approved):

PR #623, opened 2024-06-27
Sort ref ordering within components

### LINKED ISSUE #616: model.backend.update_parameter does not propagate in all equations
### What happened?

The functionality to update backend parameters does not seem to work as expected. If updating a given parameter, the parameter itself does end up being updated in the backend. However, the update does not seem to propagate to the equations using the parameter, which all remain fixated on the old value.

As an example, after building and solving the national-scale example model, one can try:
` model.backend.update_parameter('cost_flow_cap', 0)`.
This will set all investment costs in capacity to a null value across all techs. If one inspects the parameters in the backend, one can see that the operation works smoothly and as expected.

Then, if re-running the model via `model.solve(force=True)`, as suggested in the docs after updating a parameter, the solution does not reflect the expectations. In fact, `model.results.cost_investment.to_series().dropna()` (or similar ways to inspect the outputs, including from the backend) will show that investment costs are still there for all technologies, including technologies that had no other investment cost than the one for `flow_cap`. In other words, the updating of the parameter does not seem to propagate across the equations of the Pyomo model.

This was not the case in v0.6, so something must have gone wrong in the reformulation of our way to interface with Pyomo

- [ ] macOS
- [ ] Windows
- [X] Linux

### Version

v0.7.0dev3

### Relevant log output

_No response_

brynpickering:
#update

Fixes #616 

* Rebuild references in the order they appear in the backend model dataset, which is the same order they are added to it (so expressions that depend on previous expressions will be built in the correct order).
* Remove investment cost from demands in the dummy test model - this leads to a change in shadow prices which I have updated, but there's no way of knowing if the new shadow prices (or, indeed, the old ones) are correct.
* Force verbose strings on rebuilt backend model components, if verbose strings have already been triggered.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/tests/test_backend_pyomo.py
+++ b/tests/test_backend_pyomo.py
@@ -3,6 +3,7 @@ import logging
 from copy import deepcopy
 from itertools import product
 
+import calliope
 import calliope.exceptions as exceptions
 import numpy as np
 import pandas as pd
@@ -14,6 +15,8 @@ from calliope.backend.pyomo_backend_model import PyomoBackendModel
 
 from .common.util import build_test_model as build_model
 from .common.util import check_error_or_warning, check_variable_exists
+
+DUMMY_INT = 0xDEADBEEF
 
 
 @pytest.mark.xfail(reason="Not expecting operate mode to work at the moment")
@@ -1724,6 +1727,15 @@ class TestNewBackend:
         assert m.backend._has_verbose_strings
         return m
 
+    @pytest.fixture(scope="class")
+    def simple_supply_updated_cost_flow_cap(
+        self, simple_supply: calliope.Model
+    ) -> calliope.Model:
+
+        simple_supply.backend.verbose_strings()
+        simple_supply.backend.update_parameter("cost_flow_cap", DUMMY_INT)
+        return simple_supply
+
     @pytest.fixture
     def temp_path(self, tmpdir_factory):
         return tmpdir_factory.mktemp("custom_math")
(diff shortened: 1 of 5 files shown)


tests/test_backend_pyomo.py
```diff
@@ -2365,6 +2366,51 @@ def test_update_parameter_replace_undefined(self, caplog, simple_supply):
...
+        """
+        `cost_flow_cap` for test_demand_elec is not defined initially.
+        We update `cost_flow_cap` to zero for all index values
+        and then check that it has propagated through to `cost_investment_flow_cap`, `cost`, and the objective function.
+        """
+
+        caplog.set_level(logging.DEBUG)
+        simple_supply.backend.verbose_strings()
+
+        def _check_components(true_or_false: bool):
```
irm-codebase, line 2379: A tad verbose... I'd say the stages before `update_parameter` are not needed (or a separate test?).

Suggestions:
- I'd keep the docstring to a short sentence, like "Check that parameter re-definition propagates to constraints and obj. function"
- re-arrange so that asserts are done once, not inside `for` loops.
- avoid defining functions inside the test.
- Change 0 (which could be a default) to something rare. I like `0xDEADBEEF`, a common debug integer.

irm-codebase (changes requested):
Few comments on the tests.

brynpickering:
Thanks for the comments @irm-codebase - updated!

irm-codebase (approved):
Great changes! Approved.

PR #625, opened 2024-06-28
Rework calliope model attributes

### LINKED ISSUE #619: Ensuring data and model configuration remain in sync at all times (model, backend, backendmath)
### What happened?

# What is happening?

Our current approach of handling data and configuration updates is inconsistent across the code:
- Sometimes configuration updates are passed to `model._model_data` (e.g., during initialization using a preexisting file).
- Sometimes the configuration update is lost, **but the model data is modified!** (e.g., requesting operation mode for a solved model).

Similarly, the `model._model_data` object is handled inconsistently across modules. See this example:

```python
id(model._model_data)
140507415076176
id(model.math_documentation.inputs)
140507415076176
id(model.backend.inputs)  # shallow copy
140507414858960
```

Lastly, there is another variable with similar behavior, but defined at a different level: `model._model_def_dict`. This means that we are handling similar parameters at different levels.

# Why is this a problem?

As it stands, you cannot be 100% sure that, for a saved file, the solved data and the stored metadata / configuration correspond. **This is a severe issue for replicability!**.
For example: you can never be sure if your solved file is the original "planning" solution, or a test you ran on operate mode.

If you solve and save different models with different configurations, modes and math files, you should be able to reconstruct how that data was created via the configuration. 

# How to solve this?

I propose to introduce the following design requirements:
1. `model._model_data` shall not be duplicated or shallow copied by underlying functionality.
2. All modules that are not `model.py` should strive to only realize operations with `model._model_data`, and not modify it unless absolutely necessary (like saving results!).
3. Configuration, math, and model definition must be saved and handled at the `model` level (i.e., no more saving stuff at `model._model_data.attrs`. This is to encourage better practices and avoid desync issues / losing sight of were things are. When saving files, we can serialize and _only then_ add things to `model._model_data.attrs`.
5. Configuration and model definition updates must always be saved to `model`. Is up to users to decide if they save the results or not, but the calliope model file should always tell them what it will do.
6. "Warm start" updates to built models must reflect updates to parameters both in the backend and in `model._model_data`.

This helps us keep the code clean, ensures we better follow netCDF specifications (since it cannot handle lists or dictionaries), and should reduce bugs in the future.

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Version

v0.7

### Relevant log output

_No response_

irm-codebase:
Fixes #619

* Introduces better handling of Calliope configuration (by differentiating between configuration from files, overrides and runtime).
* Limits backend access to only things it _needs_ (math, defaults, build config, data)
* Better data handling to avoid duplicating information between the backend and the model object (unless necessary).
* Removes unnecessary attributes

See this table: https://github.com/calliope-project/calliope/issues/619#issuecomment-2197236102

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/.github/workflows/commit-ci.yml
+++ b/.github/workflows/commit-ci.yml
@@ -54,5 +54,8 @@ jobs:
     - name: upload coverage report to Codecov
       if: github.ref == 'refs/heads/main'
       uses: codecov/codecov-action@v4
-      env:
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+        fail_ci_if_error: true
         directory: "./reports/coverage/"
+        verbose: true


irm-codebase:
@brynpickering first step towards the new attribute setup. This one fixes how math dictionaries are handled.
Some tests related to operate mode seem to be failing... but I thought that feature was not working???

I'll try to debug that before continuing.

brynpickering:
`operate` mode _is_ working (#539)!

src/calliope/model.py
```diff
@@ -81,13 +81,15 @@ def __init__(
...
         self._model_def_path: Optional[Path]
-        self.backend: BackendModel
-        self.math_documentation = MathDocumentation()
+        self.backend: BackendModel | None = None
         self._is_built: bool = False
         self._is_solved: bool = False
 
+        # new
+        self.math: AttrDict = AttrDict()
+        self._math_dir: Path = Path(calliope.__file__).parent / "math"
```
brynpickering, line 91: I prefer `importlib.resources.files("calliope")`

src/calliope/model.py
```diff
@@ -504,7 +509,9 @@ def run(self, force_rerun=False, **kwargs):
 
     def to_netcdf(self, path):
         """Save complete model data (inputs and, if available, results) to a NetCDF file at the given `path`."""
-        io.save_netcdf(self._model_data, path, model=self)
+        self._model_data.attrs["math"] = self.math
```
brynpickering, line 512: I would prefer this step to be in IO, i.e. pass math to IO

irm-codebase:
@brynpickering this fix is causing some VERY weird behavior on the backend (actually reintroducing #623) and around operate mode.
I will try to re-approach with your suggestions and by first _only_ cleaning `model.py`. I will only introduce the backend changes after making sure that everything on the front is behaving adequately.

irm-codebase:
> `operate` mode _is_ working (#539)!

In that case, shouldn't we re-activate the tests that check it / remove the unnecessary ones?

irm-codebase:
My approach was incorrect. Retrying!

PR #627, opened 2024-06-30
Add record of single element attribute lists in IO

### LINKED ISSUE #614: Incorrect net_cdf serialisation for top level attributes
### What happened?

When saving models into net_cdf files, our serialization algorithm will convert single list elements at the top level into strings.

**Replication steps**:
```python
import calliope

model = calliope.examples.national_scale()

model_file = "outputs/model.nc"
out_file = "outputs/output.nc"

model._model_data.attrs["applied_math"] = ["a single value"]
model.to_netcdf(model_file)

test = calliope.read_netcdf(model_file)

print(model._model_data.attrs["applied_math"])
```

Here, `"applied_math"` will be a string, not a list.

- [ ] macOS
- [ ] Windows
- [X] Linux

### Version

v0.7

### Relevant log output

_No response_

brynpickering:
Fixes #614

* single element lists / sets are logged and returned to being lists / sets on loading from NetCDF

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/io.py
+++ b/src/calliope/io.py
@@ -15,6 +15,7 @@ import xarray as xr
 
 from calliope import exceptions
 from calliope.attrdict import AttrDict
+from calliope.util.tools import listify
 
 CONFIG_DIR = importlib.resources.files("calliope") / "config"
 
@@ -39,13 +40,11 @@ def read_netcdf(path):
 
 
 def _pop_serialised_list(
-    attrs: dict, serialised_items: Union[str, list, np.ndarray]
+    attrs: dict, serialised_items: Union[str, list]
 ) -> Union[list, np.ndarray]:
+    """Pop a list of serialised attributes from the attribute dictionary."""
     serialised_ = attrs.pop(serialised_items, [])
-    if not isinstance(serialised_, (list, np.ndarray)):
-        return [serialised_]
-    else:
-        return serialised_
+    return listify(serialised_)
 
 
 def _serialise(attrs: dict) -> None:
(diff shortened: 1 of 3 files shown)


src/calliope/io.py
```diff
@@ -39,26 +40,33 @@ def read_netcdf(path):
...
+    We also make note of any single element lists as they will be stored as simple strings in netcdf format,
+    but we want them to be returned as lists on loading the data from file.
+
     Args:
         attrs (dict):
             Attribute dictionary from an xarray Dataset/DataArray.
             Changes will be made in-place, so be sure to supply a copy of your dictionary if you want access to its original state.
     """
-    # Convert dicts attrs to yaml strings
+    # Keep track of single element lists, to listify them again when loading from file.
```
irm-codebase, line 63: Shouldn't we instead make sure to serialize all lists regardless of their length?
NetCDF does not support lists in attributes anyways. The single case was just what triggered this error.

irm-codebase: Maybe just highlighting that a list was serialized should be enough?
This code would work, but I wonder if just noting serialization for all lists would be a more generic, leaner fix.

irm-codebase (changes requested):
Looks good overall. I just have one question about whether or not the fix can be made more generic.
After that I can approve.

src/calliope/io.py
brynpickering: Hmm, so xarray is already doing some magic to serialise lists before saving to netcdf? Because we didn't have problems with saving lists in the past, only sets/tuples. Lists is something it can seemingly handle

brynpickering: So, xarray has handled lists of strings via its own encoding method as of [v0.10.4](https://docs.xarray.dev/en/stable/whats-new.html#v0-10-4-16-may-2018). So I'd continue with the current approach of doing the minimum necessary to make xarray happy.

irm-codebase: Understood. Changes approved!

irm-codebase (approved):

PR #628, opened 2024-07-01
Update operate mode and associated tests

brynpickering:
Fixes confusion from the tests as to whether operate mode works in v0.7 or not (it does).

I've also made some fixes to the code to deal with edge cases that appeared while I was cleaning up the tests.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/tests/test_core_model.py
+++ b/tests/test_core_model.py
@@ -1,4 +1,5 @@
 import logging
+from contextlib import contextmanager
 
 import calliope
 import numpy as np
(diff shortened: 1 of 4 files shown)


tests/test_core_model.py
```diff
@@ -281,39 +282,107 @@ def test_add_math_fails_marker_correct_position(self, simple_supply, eq_string):
...
         model.build(mode="plan")
         model.solve()
         return model
 
-    @pytest.fixture(scope="class")
-    def operate_model(self):
+    @pytest.fixture(
+        scope="class", params=[("6h", "12h"), ("12h", "12h"), ("16h", "20h")]
+    )
+    def operate_model_and_log(self, request):
```
irm-codebase, line 307: Just to understand what this is doing: this will create 3 versions of this fixture, meaning tests using it will be repeated 3 times?

irm-codebase (approved):
Looks good, I think.

Just one question related to fixture parametrization (because I did not know that was possible)

tests/test_core_model.py
brynpickering: Exactly! It's a nice way to parameterise multiple tests using the same input params

PR #629, opened 2024-07-01
Fix: Update commit-ci codecov step

brynpickering:
Fixes codecov not uploading successfully on `main` and its failure being silent. This is why we have failing coverage reports in all our PRs.

## Summary of changes in this pull request

* Copy over codecov config from pr-ci to commit-ci.

## Reviewer checklist

- [ ] Test(s) added to cover contribution
- [ ] Documentation updated
- [ ] Changelog updated
- [ ] Coverage maintained or improved

PR #630, opened 2024-07-01
Improved backend maintainability 

### LINKED ISSUE #624: Improving the handling of invalid backends
### What can be improved?

Our current implementation of different backends is a bit fickle.

```python
_BACKENDS: dict[str, Callable] = {"pyomo": PyomoBackendModel}
...
backend = self._BACKENDS[backend_name](input, self.math, **updated_build_config)
```

- It breaks the linkage between backend versions and language support tools in IDEs (e.g., refactoring won't work properly)
- If an invalid key is given, it'll return a key error. We probably want to give a more specific message to users.
- Additional backed-specific parameters are not possible?

I propose changing it to a python `match` statement with a specific error at the end.
https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching

### Version

v0.7

irm-codebase:
Fixes #624

* The backend is now a module that shares only what is necessary to `model.py`.
* Backend fetching is now a `match-case` function. We can add new backends as needed.
* Invalid backend requests now return a clear error.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/backend/__init__.py
+++ b/src/calliope/backend/__init__.py
@@ -1,6 +1,6 @@
-"""Callipe's backend module."""
+"""Calliope's optimisation backend module."""
 
-from xarray import Dataset
+import xarray as xr
 
 from calliope.backend.latex_backend_model import MathDocumentation
 from calliope.backend.parsing import ParsedBackendComponent
@@ -10,7 +10,7 @@ from calliope.exceptions import BackendError
 MODEL_BACKENDS = ("pyomo",)
 
 
-def get_model_backend(name: str, data: Dataset, **kwargs):
+def get_model_backend(name: str, data: xr.Dataset, **kwargs):
     """Assign a backend using the given configuration.
 
     Args:
@@ -29,4 +29,5 @@ def get_model_backend(name: str, data: Dataset, **kwargs):
             backend = PyomoBackendModel(data, **kwargs)
         case _:
             raise BackendError(f"Incorrect backend '{name}' requested.")
+
     return backend
--- a/src/calliope/backend/__init__.py
+++ b/src/calliope/backend/__init__.py
@@ -1,1 +1,31 @@
 """Callipe's backend module."""
+
+from xarray import Dataset
+
+from calliope import exceptions
+from calliope.attrdict import AttrDict
+from calliope.backend.latex_backend_model import MathDocumentation
+from calliope.backend.parsing import ParsedBackendComponent
+from calliope.backend.pyomo_backend_model import PyomoBackendModel
+
+
+def get_model_backend(name: str, data: Dataset, **kwargs):
+    """Assign a backend using the given configuration.
+
+    Args:
+        name (str): name of the backend to use.
+        data (Dataset): model data for the backend.
+        **kwargs: backend keyword arguments corresponding to model.config.build.
+
+    Raises:
+        exceptions.BackendError: If invalid backend was requested.
+
+    Returns:
+        BackendModel: Initialized backend object.
+    """
+    match name:
+        case "pyomo":
+            backend = PyomoBackendModel(data, **kwargs)
+        case _:
+            raise exceptions.BackendError(f"Incorrect backend '{name}' requested.")
+    return backend
--- a/src/calliope/backend/__init__.py
+++ b/src/calliope/backend/__init__.py
@@ -2,11 +2,12 @@
 
 from xarray import Dataset
 
-from calliope import exceptions
-from calliope.attrdict import AttrDict
 from calliope.backend.latex_backend_model import MathDocumentation
 from calliope.backend.parsing import ParsedBackendComponent
 from calliope.backend.pyomo_backend_model import PyomoBackendModel
+from calliope.exceptions import BackendError
+
+MODEL_BACKENDS = ("pyomo",)
 
 
 def get_model_backend(name: str, data: Dataset, **kwargs):
@@ -27,5 +28,5 @@ def get_model_backend(name: str, data: Dataset, **kwargs):
         case "pyomo":
             backend = PyomoBackendModel(data, **kwargs)
         case _:
-            raise exceptions.BackendError(f"Incorrect backend '{name}' requested.")
+            raise BackendError(f"Incorrect backend '{name}' requested.")
     return backend
--- a/src/calliope/backend/__init__.py
+++ b/src/calliope/backend/__init__.py
@@ -1,5 +1,7 @@
 """Calliope's optimisation backend module."""
 
+from typing import TYPE_CHECKING
+
 import xarray as xr
 
 from calliope.backend.latex_backend_model import MathDocumentation
@@ -9,8 +11,11 @@ from calliope.exceptions import BackendError
 
 MODEL_BACKENDS = ("pyomo",)
 
+if TYPE_CHECKING:
+    from calliope.backend.backend_model import BackendModel
 
-def get_model_backend(name: str, data: xr.Dataset, **kwargs):
+
+def get_model_backend(name: str, data: xr.Dataset, **kwargs) -> "BackendModel":
     """Assign a backend using the given configuration.
 
     Args:
--- a/src/calliope/backend/backend_model.py
+++ b/src/calliope/backend/backend_model.py
@@ -9,19 +9,16 @@ import logging
 import time
 import typing
 from abc import ABC, abstractmethod
+from collections.abc import Callable, Iterable
 from copy import deepcopy
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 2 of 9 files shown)


src/calliope/backend/latex_backend_model.py
```diff
@@ -45,7 +45,7 @@ def build(self, include: Literal["all", "valid"] = "all", **kwargs) -> None:
             **kwargs: kwargs for the LaTeX backend.
         """
         backend = LatexBackendModel(self._inputs, include=include, **kwargs)
-        backend._build()
+        backend.build()
```
irm-codebase, line 48: Changed this to a public since we do access it outside of this file. Can revert if needed.

src/calliope/backend/helper_functions.py
```diff
@@ -31,7 +30,7 @@ def __init__(
         *,
         equation_name: str,
         input_data: xr.Dataset,
-        backend_interface: Optional[type[BackendModel]] = None,
+        backend_interface=None,
```
irm-codebase, line 33: For some reason this ignores `TYPE_CHECKING`. So I removed it.

src/calliope/model.py
irm-codebase, line 1: Most modifications are style fixes to avoid unnecessary imports.

src/calliope/model.py
```diff
@@ -357,23 +354,24 @@ def build(self, force: bool = False, **kwargs) -> None:
...
             )
         else:
-            input = self._model_data
-        backend_name = updated_build_config["backend"]
-        backend = self._BACKENDS[backend_name](input, **updated_build_config)
-        backend._build()
-        self.backend = backend
+            backend_input = self._model_data
+        backend_name = backend_config.pop("backend")
+        self.backend = backend.get_model_backend(
```
irm-codebase, line 371: This is the actual important fix in this file.

tests/common/util.py
```diff
@@ -90,9 +91,9 @@ def build_lp(
         model (calliope.Model): Calliope model.
         outfile (Union[str, Path]): Path to LP file.
         math (Optional[dict], optional): All constraint/global expression/objective math to apply. Defaults to None.
-        backend (Literal["pyomo"], optional): Backend to use to create the LP file. Defaults to "pyomo".
+        backend_name (Literal["pyomo"], optional): Backend to use to create the LP file. Defaults to "pyomo".
     """
-    backend_instance = model._BACKENDS[backend](model._model_data)
+    backend_instance = backend.get_model_backend(backend_name, model._model_data)
```
irm-codebase, line 96: Also converted to the new approach.

tests/test_backend_module.py
irm-codebase, line 1: New test file for the backend module. Just focuses on assessing if backend requests behave as intended.

irm-codebase (commented):
@brynpickering some comments to help with the review.
Let me know what you think!

irm-codebase:
Today I learned that python's use of `TYPE_CHECKING` is inconsistent for some cases: https://vickiboykis.com/2023/12/11/why-if-type_checking/

src/calliope/backend/__init__.py
```diff
@@ -1 +1,32 @@
 """Callipe's backend module."""
+
+from xarray import Dataset
```
brynpickering, line 3: I prefer keeping the source library where possible, so `import xarray as xr` and then `xr.Dataset`. Makes it easier to track what is an calliope method/class and what isn't.

src/calliope/backend/helper_functions.py
brynpickering: Could you try `Optional[type["BackendModel"]]`?

src/calliope/backend/latex_backend_model.py
brynpickering: It may make sense. The reason I had it hidden was because we don't want `model.backend.build()` to be available to the user

irm-codebase:
The codeconv issue is likely due to #629. The only added file that has extra code is the backend's `__init__.py`, and it has 100% coverage.

src/calliope/backend/__init__.py
```diff
@@ -1 +1,32 @@
 """Callipe's backend module."""
```
brynpickering, line 1:
```suggestion
"""Calliope's optimisation backend module."""
```

src/calliope/backend/pyomo_backend_model.py
```diff
@@ -8,12 +8,11 @@
 import os
 import re
 from abc import ABC
+from collections.abc import Iterable, Iterator
```
brynpickering, line 11: I assume this is now the recommended source for these?

brynpickering (commented):
Looks good! Just some minor comments to resolve

src/calliope/backend/latex_backend_model.py
irm-codebase: Hmmm... I tend to prefer viewing each module by its own "scope" (i.e., `model` is aware of `backend`, but not the other way around). This makes it clear what is meant to be used at the file level, and it's a bit less tangly in the long-run.

If we want to avoid users messing with the backend, then it's `model._backend` what should change (with us providing interfaces / wrappers). But honestly I think that even in that case it is not needed.

irm-codebase: In theory, an advanced user could create an additional "step" that injects stuff and re-builds the backend... and it may work!

However, if we really want to "hide" this function, I could move it to the `__init__.py` get function, and just return the fully built backend (no difference).
That way, all the `_build()` usage stays within the backend module, somewhat.

brynpickering: We had `_backend` and an interface to that in the previous version and it was a real pain to maintain. We could simply rename `_build` to something more exlicit like `add_all_math`

Or, since we never have any intermediate steps between backend model initialisation and calling `_build`, we could move the `_build` step to within `BackendModel.__init__`. I think that's my prefered option.

irm-codebase: I'll do the following:
- Rename `backend.build` -> `backend.build_all_math` (in theory it's added during `__init__`, so I like this name more.
- Move `backend.build` to the `__init__` function so that upper functionality has no access to it.
- Add a `build_math` boolean to the `__init__` function (defaults to `True`) to deactivate math building in cases were we just want the backend.

irm-codebase: I am doing this because, for some test cases, we only want the backend, and we inject math into it.

src/calliope/backend/helper_functions.py
```diff
@@ -30,7 +34,7 @@ def __init__(
         *,
         equation_name: str,
         input_data: xr.Dataset,
-        backend_interface=None,
+        backend_interface: type["BackendModel"] | None = None,
```
irm-codebase, line 37: the suggestion worked!

src/calliope/backend/__init__.py
```diff
@@ -1 +1,33 @@
...
+
+from calliope.backend.latex_backend_model import MathDocumentation
+from calliope.backend.parsing import ParsedBackendComponent
+from calliope.backend.pyomo_backend_model import PyomoBackendModel
+from calliope.exceptions import BackendError
+
+MODEL_BACKENDS = ("pyomo",)
+
+
+def get_model_backend(name: str, data: xr.Dataset, **kwargs):
```
brynpickering, line 13: This should have a return type hint

src/calliope/backend/backend_model.py
```diff
@@ -203,8 +200,8 @@ def _check_inputs(self):
             check_results["warn"], check_results["fail"]
         )
 
-    def build(self):
-        """Build the backend using defined math and input data."""
+    def build_all_math(self):
```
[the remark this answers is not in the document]
brynpickering: Can we use `add_all_math`?

brynpickering (approved):
One request to change the name of the (now public) backend method, otherwise all good by me.

src/calliope/backend/__init__.py
irm-codebase: Unfortunately this is where the nasty `TYPE_CHECKING` issue occurs.
I have not found a good way of avoiding it...

I did add it to the docstring, but the only way I've found to fix it would involve changing the circular typing issue between helper functions and the generic backend file.

brynpickering: Hmm, and `IF TYPE_CHECKING` and then stringifying the type hint won't do it?

irm-codebase: Oh, that is a neat trick! I thought it would only work for `type["XXXXX"]` cases, for some reason.

irm-codebase:
@brynpickering all fixes are in!

PR #631, opened 2024-07-02
Add pyupgrade to ruff config

brynpickering:
Fixes all our outdated Python syntax.

Thanks to @irm-codebase for pointing to [pyupgrade](https://docs.astral.sh/ruff/rules/#pyupgrade-up). I think it's worth having it in our core config as it will keep our code cleaner and easier to review/maintain. 

* Added `UP` to ruff config.
* Applied `pre-commit`, cleaning up any unresolved fixes.
* Reintroduce `skip-trailing-comma` and add the relevant `isort` config to make it work. I much prefer the result, even if it leads to more line changes per commit, on average.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,9 +35,13 @@ line-length = 88
 
 [tool.ruff.format]
 exclude = [".*.egg-info", "requirements/**"]
+skip-magic-trailing-comma = true
+
+[tool.ruff.isort]
+split-on-trailing-comma = false
 
 [tool.ruff.lint]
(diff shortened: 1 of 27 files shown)


irm-codebase (changes requested):
Super in favor if this... but the docstrings also need to be updated!
I use a plug-in for this, so I'll submit a PR into this PR to get these in with your approval.

irm-codebase:
Please update the CHANGELOG (adding it to the other linting update should be enough, imo).

PR #639, opened 2024-07-16
Rework math

### LINKED ISSUE #608: Potential desync in model configuration at the model.py level

irm-codebase:
Partially fixes #608 #619

* `MathDocumentation` is now in postprocessing. `Model` no longer includes it (instead, it's an input to it).
* `model._model_def_path` is clearly processed to avoid unnecessary code. As a result, `load.py` is now `scenarios.py`, which is what it was really doing all along.
* `ModelMath` object introduced to handle math file additions better.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


irm-codebase:
@brynpickering requesting a preeliminary review of this, with some open questions:

After this cleanup, it's clear that `Model` does not even use math at all. I am still initializing it there, but in reality only the backend uses it.
- Should we just create the math when the backend is called? Or do we want to keep the `init` copy for some reason?

docs/hooks/generate_math_docs.py
irm-codebase, line 1: Changes: 
- Refactoring for better naming.
- base.yaml -> plan.yaml

src/calliope/backend/__init__.py
irm-codebase, line 1: Changes:
- Math input to backend (may be unnecessary?)
- Added gurobi to backend list.

src/calliope/backend/backend_model.py
irm-codebase, line 1: Changes:
- inclusion of `math` object
- improved logic of `mode` check.

src/calliope/backend/backend_model.py
```diff
@@ -223,21 +224,16 @@ def _add_run_mode_math(self) -> None:
...
+        run_mode_mismatch = not_run_mode.intersection(self.math.history)
         if run_mode_mismatch:
             exceptions.warn(
                 f"Running in {mode} mode, but run mode(s) {run_mode_mismatch} "
                 "math being loaded from file via the model configuration"
             )
-
-        if mode != "plan" and mode not in add_math:
+        if mode not in self.math.history:
             LOGGER.debug(f"Updating math formulation with {mode} mode math.")
```
irm-codebase, line 235: Should I remove this? math file addition already adds log messages.

src/calliope/backend/latex_backend_model.py
irm-codebase, line 1: Changes:
- Extraction of `MathDocumentation`

src/calliope/io.py
irm-codebase, line 1: Changes:
- Improved `save_netcdf` logic with added flexibility (caller now determines how attributes will be saved).

src/calliope/math/plan.yaml
irm-codebase, line 1: New name to match its `mode` declaration. Helps manage math history and improves clarity.

src/calliope/model.py
```diff
@@ -497,7 +454,15 @@ def run(self, force_rerun=False, **kwargs):
 
     def to_netcdf(self, path):
         """Save complete model data (inputs and, if available, results) to a NetCDF file at the given `path`."""
-        io.save_netcdf(self._model_data, path, model=self)
+        saved_attrs = {}
+        for attr in set(self.ATTRS_SAVED) & set(self.__dict__.keys()):
+            if not isinstance(getattr(self, attr), str | list | None):
+                # TODO: remove `dict`` once AttrDict init issue is fixed
```
irm-codebase, line 460: this comment is invalid, I should remove

src/calliope/model.py
```diff
@@ -561,8 +526,8 @@ def validate_math_strings(self, math_dict: dict) -> None:
         """
```
irm-codebase, line 526: I believe this function belongs here and not in `ModelMath` since it is aware of the existence of `_model_data` and `math`.

src/calliope/model.py
irm-codebase, line 1: Changes:
- `_add_math` logic removed
- more explicit `init` logic to remove complexity in some preprocess libraries.
- modified read / save to netCDF logic to handle new math object (also more flexible)

src/calliope/postprocess/math_documentation.py
irm-codebase, line 1: Taken from  the latex backend. Basically the same, except that it receives  `Model` now.

src/calliope/preprocess/model_math.py
irm-codebase, line 1: This the most important file in this PR. Functions aim to do a single thing, to keep them lean.

Disallows addition of the same file twice, but this can be removed.

src/calliope/preprocess/scenarios.py
irm-codebase, line 1: Removed chunks of code thanks to the `init` improvements in `Model`.

tests/conftest.py
irm-codebase, line 1: If `backend` is the only one containing `math`, most of these changes would not be needed.

tests/test_postprocess_math_documentation.py
irm-codebase, line 1: Taken from the latex test file, for better book keeping.

tests/test_preprocess_model_math.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def test_user_math_add(
+        self, model_math_w_mode_user, predefined_mode_data, user_math
+    ):
+        """Added user math should be in data."""
+        expected_math = deepcopy(predefined_mode_data)
+        expected_math.union(user_math, allow_override=True)
+        flat = expected_math.as_dict_flat()
+        assert all(
+            model_math_w_mode_user.data.get_key(i) == flat[i] for i in flat.keys()
```
irm-codebase, line 133: Checks every single math combination!

irm-codebase (commented):
Few comments to help with the review. Let me know what you think!

brynpickering (commented):
Nice cleanup. Just a few comments:


1. math documentation building isn't really a postprocessing step, it's a parallel process to building/solving the optimisation problem. Not sure where it should go, tbh.
2. This is getting us one step closer to removing math from `calliope.Model` entirely, and I think I'm in favour of that. It just requires moving the `validating_math_strings` method to somewhere more suitable. 
3. I like being able to add a math dictionary and not just a reference to a math YAML. It's effectively equivalent to adding `scenarios` vs `override_dict` at `calliope.Model` instatiation. Ideally we'd add this to `calliope.Model.build` and allow that dict to completely replace the math dict or to be added as another override. If the approach matches `scenarios`/`override_dict` then `add_math` list would be applied first, followed by the `math_dict`. It just then needs a flag for ignoring `math/base.yaml`(/`math/plan.yaml`).

src/calliope/backend/__init__.py
```diff
@@ -5,23 +5,30 @@
...
+    ALLOWED_MATH_FILE_FORMATS,
+    LatexBackendModel,
+)
 from calliope.backend.parsing import ParsedBackendComponent
 from calliope.backend.pyomo_backend_model import PyomoBackendModel
 from calliope.exceptions import BackendError
+from calliope.preprocess import ModelMath
 
-MODEL_BACKENDS = ("pyomo",)
+MODEL_BACKENDS = ("pyomo", "gurobi")
```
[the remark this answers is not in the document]
irm-codebase: In the tests for this `__init__` I am referring to this directly. Kind of a catch 22: either you forget to add it here, or you forget to add it to the test. I like this one because it's more visible.

The real backend tests do not use this tuple, so it's not a big issue.

src/calliope/backend/backend_model.py
```diff
@@ -201,6 +201,7 @@ def _check_inputs(self):
     def add_all_math(self):
         """Parse and all the math stored in the input data."""
         self._add_run_mode_math()
```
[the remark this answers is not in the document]
irm-codebase: No, that would mix purposes: the backend handles math components/delivery to the optimiser, `model_math` is just for dict definitions, file handling and similar.

How about renaming this to `add_all_components`?

src/calliope/model.py
```diff
@@ -93,14 +89,22 @@ def __init__(
...
+                model_def_dict = AttrDict.from_yaml(model_definition)
+
+            (model_def, applied_overrides) = preprocess.load_scenario_overrides(
+                model_def_dict, scenario, override_dict, **kwargs
             )
+
             self._init_from_model_def_dict(
                 model_def, applied_overrides, scenario, data_source_dfs
             )
+            self.math = preprocess.ModelMath(
```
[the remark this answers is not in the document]
irm-codebase: Honestly? No, I do not think so. I like the idea of just moving stuff to `backend.__init__` above:
- Create `ModelMath` with `add_math`.
- Add mode (which is kind of redundant now...)

`Model` will have no math until you build, but maybe that's good?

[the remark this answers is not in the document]
brynpickering: I think it's ok to not have any math until you build, so long as we have a parallell method to build and write the math docs. Then `calliope.Model` could just have `applied_math`, like `applied_overrides` in some way. It tells you what was applied but doesn't give you the sense that you can edit the math dict and it will change the built model in some way.

src/calliope/model.py
irm-codebase: Hmmm...
I think it's fine here. The math and parser classes are specialized, and this function is aware of the overall structure of calliope. I will change the validation to use the math object's method, though. I forgot to switch it...

src/calliope/backend/__init__.py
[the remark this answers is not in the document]
irm-codebase: Also, this global is just so users can see the available backend options. Gurobi would not fail even if you remove it.

We can remove, if you wish

src/calliope/backend/backend_model.py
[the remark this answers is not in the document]
irm-codebase: Settled for `add_optimisation_components`. Feels more descriptive to me.

src/calliope/preprocess/model_math.py
```diff
@@ -0,0 +1,136 @@
...
+
+        if isinstance(math_to_add, list):
+            self._init_from_list(["plan"] + math_to_add, model_def_path)
+        else:
+            self._init_from_dict(math_to_add)
+
+    def __eq__(self, other):
+        """Compare between two model math instantiations."""
+        if not isinstance(other, ModelMath):
+            return NotImplemented
```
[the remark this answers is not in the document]
irm-codebase: Nope, you return `NotImplemented` so that comparisons between `1 == model_math` returns `False`.
boolean comparison is not expected to fail

src/calliope/preprocess/model_math.py
```diff
@@ -0,0 +1,136 @@
...
+        Raises:
+            ModelError: user-math requested without providing `model_def_path`.
+        """
+        for math_name in math_to_add:
+            if not math_name.endswith((".yaml", ".yml")):
+                self.add_pre_defined_file(math_name)
+            elif model_def_path is not None:
+                self.add_user_defined_file(math_name, model_def_path)
+            else:
+                raise exceptions.ModelError(
```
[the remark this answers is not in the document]
irm-codebase: Ahhh, good idea. Then the caller is in charge of sorting this out. I like it.

src/calliope/preprocess/model_math.py
```diff
@@ -0,0 +1,136 @@
...
+                    "Must declare `model_def_path` when requesting user math."
+                )
+
+    def _init_from_dict(self, math_dict: dict) -> None:
```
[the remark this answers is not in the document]
irm-codebase: Pretty much! Check __iter__ in this class.
It's possible to omit specific attributes too.

src/calliope/preprocess/model_math.py
```diff
@@ -0,0 +1,136 @@
...
+            )
+        self._add_math(math)
+        self.history.append(name)
+        LOGGER.info(f"ModelMath: added file '{name}'.")
+
+    def add_pre_defined_file(self, filename: str) -> None:
+        """Add pre-defined Calliope math (no suffix)."""
+        if self.check_in_history(filename):
+            raise exceptions.ModelError(
+                f"Attempted to override math with pre-defined math file '{filename}'."
```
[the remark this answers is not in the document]
irm-codebase: Improved the message... but I do not like the idea of calling functions within exceptions unless absolutely necessary. Makes code less explicit.

src/calliope/preprocess/model_math.py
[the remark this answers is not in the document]
irm-codebase: I went back and forth in this, but I prefer to force the external caller to always "know" the `def_path`.
This is because the model can be saved, files moved, and then reinitialised (meaning def_path must be modified).

Saving def_path in this class would require additional logic (is it a string? a None?), while forcing the caller to provide it makes it very clear that its expected to be known.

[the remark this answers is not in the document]
irm-codebase: Adding it would also complicate the saving to dicts (we also save it in `Model`). Should we also pass it to the saved math dict? Omit? I prefer an additional parameter in `__init__` over all those vulnerabilities.

Finally, forcing the caller to pass relative paths would increase overhead, since `add_math` is meant to be relative.

src/calliope/preprocess/model_math.py
```diff
@@ -0,0 +1,136 @@
...
+
+        Args:
+            math_to_add (list[str]): Calliope math files to load. Suffix implies user-math.
+            model_def_path (str | Path | None, optional): Model definition path. Defaults to None.
+
+        Raises:
+            ModelError: user-math requested without providing `model_def_path`.
+        """
+        for math_name in math_to_add:
+            if not math_name.endswith((".yaml", ".yml")):
```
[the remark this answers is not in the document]
irm-codebase: This comes from the yaml, which is always in `str`. Since these are saved to `history`, it's better to handle it as a string to make saving more stable.

src/calliope/model.py
```diff
@@ -324,11 +322,20 @@ def build(self, force: bool = False, **kwargs) -> None:
             )
         else:
             backend_input = self._model_data
+
+        init_math_list = [] if backend_config.get("ignore_mode_math") else [mode]
+        end_math_list = [] if add_math_dict is None else [add_math_dict]
+        full_math_list = init_math_list + backend_config["add_math"] + end_math_list
+        LOGGER.debug(f"Math preprocessing | Loading math: {full_math_list}")
+        model_math = preprocess.CalliopeMath(full_math_list, self._def_path)
```
irm-codebase, line 330: This "math mode salad" is quite hard to follow because of its "mixed value" nature.
- `init_math_list` will have either `[]` or `[str]`
- the middle (`backend_config["add_math"]`) will have `[]` `[str]`
- `end_math_list` (`add_math_dict`) will have either `None` or `[dict]` 

How about:
```python
math_list = [backend_config.get("mode"), backend_config.get("add_math"), add_math_dict]
math_list = list(filter(bool, math_files))  # Removes False equivalents None, "", [], {}
model_math = preprocess.CalliopeMath(math_list, self._def_path)
```

src/calliope/preprocess/model_math.py
```diff
@@ -1,55 +1,56 @@
...
-        elif isinstance(math_to_add, str):
-            math_to_add = [math_to_add]
-
-        if isinstance(math_to_add, list):
-            self._init_from_list(["plan"] + math_to_add, model_def_path)
-        else:
-            self._init_from_dict(math_to_add)
+        self.data: AttrDict = AttrDict(
+            {
+                "variables": {},
```
irm-codebase, line 37: Should be `TypedDict` above?
The backend also refers to through `ORDERED_COMPONENTS_T`, so maybe the order should be defined here so that we have it in one tidy place.

src/calliope/preprocess/model_math.py
```diff
@@ -1,55 +1,56 @@
...
-MATH_DIR = files("calliope") / "math"
 
 
-class ModelMath:
-    """Calliope math preprocessing."""
+class CalliopeMath:
+    """Calliope math handling."""
 
     ATTRS_TO_SAVE = ("history", "data")
+    ATTRS_TO_LOAD = ("history",)
```
irm-codebase, line 20: Why are we not loading back the math data?
I was doing it because the file structure might have changed, meaning the "rebuild" would fail since the reference directories would be different...

src/calliope/preprocess/model_math.py
```diff
@@ -58,79 +59,116 @@ def __iter__(self):
...
-        """Load math definition from a list of files.
+    def __repr__(self) -> str:
+        """Custom string representation of class."""
+        return f"""Calliope math definition dictionary with:
+    {len(self.data["variables"])} decision variable(s)
+    {len(self.data["global_expressions"])} global expression(s)
+    {len(self.data["constraints"])} constraint(s)
+    {len(self.data["piecewise_constraints"])} piecewise constraint(s)
+    {len(self.data["objectives"])} objective(s)
+        """
```
irm-codebase, line 70: Incorrect tabs trigger a `mypy` error.
```suggestion
    def __repr__(self) -> str:
        """Custom string representation of class."""
        return f"""
        Calliope math definition dictionary with:
        {len(self.data["variables"])} decision variable(s)
        {len(self.data["global_expressions"])} global expression(s)
        {len(self.data["constraints"])} constraint(s)
        {len(self.data["piecewise_constraints"])} piecewise constraint(s)
        {len(self.data["objectives"])} objective(s)
        """
```

src/calliope/preprocess/model_math.py
```diff
@@ -58,79 +59,116 @@ def __iter__(self):
...
         """
-        math_to_validate = deepcopy(self.data)
-        if extra_math is not None:
-            math_to_validate.union(AttrDict(extra_math), allow_override=True)
-        validate_dict(math_to_validate, MATH_SCHEMA, "math")
-        LOGGER.info("ModelMath: validated math against schema.")
+        if not math_to_add.endswith((".yaml", ".yml")):
+            self._add_pre_defined_file(math_to_add)
+        else:
+            self._add_user_defined_file(math_to_add, model_def_path)
```
irm-codebase, line 163: Bug: does not catch invalid call when requesting user defined math without a def_path.
```suggestion
        if not math_to_add.endswith((".yaml", ".yml")):
            self._add_pre_defined_file(math_to_add)
        elif model_def_path:
            self._add_user_defined_file(math_to_add, model_def_path)
        else:
            raise ModelError("Must define model definition path when requesting user math.")
```

irm-codebase (commented):
Went through the changes. I'm quite happy with them!

I've added some suggestions to avoid bugs catched by `mypy`, and to improve logic in some parts.

src/calliope/backend/backend_model.py
```diff
@@ -280,7 +279,7 @@ def _add_component(
                 this name must be available in the input math provided on initialising the class.
             component_dict (Tp): unparsed YAML dictionary configuration.
             component_setter (Callable): function to combine evaluated xarray DataArrays into backend component objects.
-            component_type (Literal["variables", "global_expressions", "constraints", "objectives"]):
+            component_type (Literal["variables", "global_expressions", "constraints", "piecewise_constraints", "objectives"]):
```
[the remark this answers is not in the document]
irm-codebase: See proposal to move this to a `TypedDict` in `CalliopeMath` below.

src/calliope/backend/expression_parser.py
```diff
@@ -788,7 +789,7 @@ def as_array(self) -> xr.DataArray:  # noqa: D102, override
                 evaluated = backend_interface._dataset[self.name]
             except KeyError:
                 evaluated = xr.DataArray(self.name, attrs={"obj_type": "string"})
-        if "default" in evaluated.attrs:
+        if "default" in evaluated.attrs and pd.notnull(evaluated.attrs["default"]):
```
irm-codebase, line 792: `pd.notnull` here should change.
- Generally, `pd.isna` and `pd.notna` is preferred over `null` (https://docs.astral.sh/ruff/rules/pandas-use-of-dot-not-null/)
- This will generate odd behavior if there is a `list` in `"default"`, since it will return an object and the check will pass if there is _anything_ in it.

src/calliope/backend/backend_model.py
```diff
@@ -605,7 +602,7 @@ def _filter(val):
         in_math = set(
             name
             for component in ["variables", "global_expressions"]
-            for name in self.math.data[component].keys()
+            for name in self.math.data[component]
```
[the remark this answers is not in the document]
brynpickering: A linter suggestion. `keys()` is implied when looping over a dictionary.

src/calliope/backend/parsing.py
```diff
@@ -855,8 +855,7 @@ def raise_caught_errors(self):
             exceptions.print_warnings_and_raise_errors(
                 errors={f"{self.name}": self._errors},
                 during=(
-                    "math string parsing (marker indicates where parsing stopped, "
-                    "which might not be the root cause of the issue; sorry...)"
+                    "math string parsing (marker indicates where parsing stopped, but may not point to the root cause of the issue)"
```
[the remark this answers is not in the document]
brynpickering: who doesn't wrap lines in their IDEs these days?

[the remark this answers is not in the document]
irm-codebase: Huh, apparently me! Thanks for the tip.

Although, I'd still recommend to do this, because code isn't always evaluated through IDEs (GitHub reviews, for example).

I'll mark this as resolved since it does not trigger a linter issue, and is just style.

irm-codebase:
@brynpickering what's next for this feature?
Will we wait until the approach to `parameters` is set, or should we merge this beforehand?

src/calliope/backend/expression_parser.py
brynpickering: Fair enough, although "default" _shouldn't_ ever have a list in it... `notnull` was always used to capture NaN and None values, but now that `notna` catches None values we may as well move to using it

src/calliope/backend/latex_backend_model.py
```diff
@@ -491,7 +483,7 @@ def generate_math_doc(
             ]
             if getattr(self, objtype).data_vars
         }
-        if not components["parameters"]:
+        if "parameters" in components and not components["parameters"]:
```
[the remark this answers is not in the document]
brynpickering: `components` is a dictionary of lists that we explicitly generate in the preceding statement. `not components["parameters"]` is simply checking whether the `parameter` list is empty.

irm-codebase:
@brynpickering I've updated everything, and tests pass.
However, there seems to be a nasty pre-commit issue that prevents fixes to go through on my side...
For some reason it's re-arranging the imports every time...

`ruff check --fix` will sort them one way....
and pre-commit will cancel it, leading to a failure.

Seems related to breaking changes in 0.6 (https://astral.sh/blog/ruff-v0.6.0#migrating-to-v06)

Have you faced this issue before?

brynpickering:
Yes, I'm having the same issues. It is just a mismatch in the version of ruff you're using on autosave vs the one pre-commit has updated to. Best is to use `fromEnvironment` import strategy and then make sure that the version in your dev conda env is the same as in pre-commit.

irm-codebase:
That's odd... that's the configuration I am using. I even tried updating the pre-commit hook and ruff to 0.6.8 (the newest), and it still fails. Let me know if you find an approach that works.

brynpickering:
Weird, works fine for me having updated ruff to v0.6 in my calliope dev env

brynpickering (approved):
Let's get this merged in and hope we can clean up around the edges as necessary!

Issue #643, opened 2024-07-26
Proposed fixes / consistency changes for documentation


sstroemer:
### Description

Follow up after a discussion with @sjpfenninger on 18.07., regarding some findings after combing through the current documentation, in no particular order (just numbered to make references in discussion easier):

1. `lookup_cluster_first_time` occurs in the base math; should that be part of the _Inter-cluster storage math_? [[link](https://calliope.readthedocs.io/en/latest/math/base/#balance_supply_with_storage)]
2. Missing links between entries:
	- `flow_cap` [[link](https://calliope.readthedocs.io/en/latest/math/base/#flow_cap)] makes use of, e.g., `flow_cap_min`, but the latter does not show that in it's _Used in_ section [[link](https://calliope.readthedocs.io/en/latest/math/base/#flow_cap_min)]
    - other parameters may be missing too, I assume these relations are built automatically, so probably just a single fix necessary
3. Adding "backlinks" between entries: [`area_use_max`](https://calliope.readthedocs.io/en/latest/math/base/#area_use_max) is used in [`area_use`](https://calliope.readthedocs.io/en/latest/math/base/#area_use), which is easy to follow. The latter makes use of [`source_unit`](https://calliope.readthedocs.io/en/latest/math/base/#source_unit), which may be hard to see on a first look, and further is then "hard" to follow up on. I assume the _"used in"_ section is built automatically, could the same approach also be used to build a _"makes use of"_  section? That would
    - show all used parameters, etc. immediately in a single list, and
    - make it easy to traverse the documentation by just hopping along the "dependency graph".
4. `balance_storage` checks for `where: (include_storage=true or base_tech=storage) ...`, whereas others check for `where: storage ...` instead; it seems to me those should be equivalent, because `storage` only exists with exactly the first condition [[link](https://calliope.readthedocs.io/en/latest/math/base/#storage)] - if that is true this could be unified for consistency, if not, the difference in behavior could be documented
5. `flow_cap_min` [[link](https://calliope.readthedocs.io/en/latest/math/base/#flow_cap_min)] contains a note stating _"this will force `flow_cap` to a minimum value unless `cap_method` is set to `integer`"_, which seems to not relate to the definition of `flow_cap` in a setting where `cap_method == LP` (and maybe more to `flow_cap_per_unit`, etc.).
[...]

Issue #642, opened 2024-07-19
Letting users define parameters and additional dimensions in YAML files


irm-codebase:
### What can be improved?

Opening this issue after discussions with @sjpfenninger  and @brynpickering 

Currently, the math data and schemas are a bit 'mixed'. In particular, `model_def_schema.yaml` contains  parameter defaults, which makes looking for them difficult.

Similarly, the current parsing of yaml files is a bit difficult in where/how new dimensions and parameters are declared.

For parameters:
- Declare current parameters in `base.yaml` instead.
- Make users explicitly declare new parameters in yaml files, with at least the default value. We can just add a new `params` section for it.
- Parameters must at least have "default" declared (with some additional logic to convert 0 to null, or warning users of it).

For dimensions: 
- Make users declare new dimensions in yaml files. Possible important settings are "ordered" / "numeric" for stuff like years and timesteps were the distance between them might matter.
- Ditto for groups (see #604 for discussion).

The idea is to make model definition files less ambiguous. We probably should also think on how this affects schema evaluation, and some of the parsing.

### Version

v0.7

brynpickering:
I've had a look at this and think it could work pretty well _most_ of the time. However, there is the issue that config for parameters is then only analysed on calling `calliope.Model.build`. This is problematic for time resampling since one of the options is about how to resample a given parameter (taking the average or the sum).

One thing we could do is resample the data on-the-fly when calling `calliope.Model.build`  (something I considered in the past), so the timeseries isn't resampled at instatiation. This would probably work fine but would come with the disadvantage that _either_ results have to be in their own xarray dataset (with a different `timesteps` dimension) or the results are returned with a sparse `timesteps` dimension (e.g. every other timestep has a value with a 2h resolution). 

If we continue with resampling at instatiation, there needs to be a way of defining parameter config that is separate from the math - does this get messy?

irm-codebase:
I think it's better to go with the option that is (long term) easier to manage, which would be resampling during build.

Functionality-wise, it makes sense: you are telling the model to build/prepare itself. If need be, we can split backend steps and just wrap around them:
1. Model init
2. Model build
   - `build.math`
   - `build.timeseries`
   - `build.whatever`
3. Model solve...

For now, let us assume that we control the build process (i.e., `model.build` will run everything).
If we want users to be able to build specifics, we'll need an attribute that lets us know the [state machine's status](https://en.wikipedia.org/wiki/Finite-state_machine).

[...]

PR #644, opened 2024-07-26
Fixes in docs and new data checks

sstroemer:
Related to #643 

* Small fixes (typos, ...) to docstrings
* Checks for valid values of ~~`flow_cap_*` and~~ `storage_initial` (d99bf9451e4d0a1a7e26cf98bef28ea19d6132fb); ~~note that I'm unsure if the check for `flow_cap_min <= flow_cap_max` is done correctly?~~

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/config/model_data_checks.yaml
+++ b/src/calliope/config/model_data_checks.yaml
@@ -29,12 +29,6 @@ fail:
   - where: storage_initial<0 OR storage_initial>1
     message: "storage_initial is a fraction, requiring values within the interval [0, 1]."
 
-  - where: flow_cap_min<0
-    message: "flow_cap_min has to be non-negative."
-
-  - where: (flow_cap_min AND flow_cap_max) AND flow_cap_max<flow_cap_min
-    message: "flow_cap_min should be less than (or equal to) flow_cap_max."
-
   - where: integer_dispatch=True AND NOT cap_method=integer
     message: Cannot use the integer `integer_dispatch` unless the technology is using an integer unit capacities (`cap_method=integer`).
 
--- a/src/calliope/config/model_data_checks.yaml
+++ b/src/calliope/config/model_data_checks.yaml
@@ -26,11 +26,8 @@ fail:
   - where: carrier_export and not any(carrier_out, over=nodes)
     message: "Export carriers must be one of the technology outflow carriers."
 
-  - where: storage_initial>1
-    message: "storage_initial is a fraction; values larger than 1 are not allowed."
-
-  - where: storage_initial<0
-    message: "storage_initial is a fraction; values smaller than 0 are not allowed."
+  - where: storage_initial<0 OR storage_initial>1
+    message: "storage_initial is a fraction, requiring values within the interval [0, 1]."
 
   - where: flow_cap_min<0
     message: "flow_cap_min has to be non-negative."
--- a/src/calliope/config/model_data_checks.yaml
+++ b/src/calliope/config/model_data_checks.yaml
@@ -29,6 +29,9 @@ fail:
   - where: storage_initial>1
     message: "storage_initial is a fraction; values larger than 1 are not allowed."
 
+  - where: storage_initial<0
+    message: "storage_initial is a fraction; values smaller than 0 are not allowed."
+
   - where: integer_dispatch=True AND NOT cap_method=integer
     message: Cannot use the integer `integer_dispatch` unless the technology is using an integer unit capacities (`cap_method=integer`).
 
--- a/src/calliope/config/model_data_checks.yaml
+++ b/src/calliope/config/model_data_checks.yaml
@@ -32,6 +32,12 @@ fail:
   - where: storage_initial<0
     message: "storage_initial is a fraction; values smaller than 0 are not allowed."
 
+  - where: flow_cap_min<0
+    message: "flow_cap_min has to be non-negative."
+
+  - where: (flow_cap_min AND flow_cap_max) AND flow_cap_max<flow_cap_min
+    message: "flow_cap_min should be less than (or equal to) flow_cap_max."
+
   - where: integer_dispatch=True AND NOT cap_method=integer
     message: Cannot use the integer `integer_dispatch` unless the technology is using an integer unit capacities (`cap_method=integer`).
 
--- a/src/calliope/config/model_def_schema.yaml
+++ b/src/calliope/config/model_def_schema.yaml
@@ -707,7 +707,7 @@ properties:
             title: Required sink use.
             description: >-
               Required amount of carrier removal from the system (e.g., electricity demand, transport distance).
-              Unit dictated by `source_unit`.
+              Unit dictated by `sink_unit`.
 
           source_unit:
             type: string
--- a/src/calliope/config/model_def_schema.yaml
+++ b/src/calliope/config/model_def_schema.yaml
@@ -697,7 +697,7 @@ properties:
             title: Maximum bound on sink.
             description: >-
               Maximum sink use to remove a carrier from the system (e.g., electricity demand, transport distance).
-              Unit dictated by `source_unit`.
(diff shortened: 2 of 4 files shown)


src/calliope/config/model_data_checks.yaml
```diff
@@ -29,6 +29,15 @@ fail:
   - where: storage_initial>1
     message: "storage_initial is a fraction; values larger than 1 are not allowed."
 
+  - where: storage_initial<0
```
brynpickering, line 32: merge with above: `where: storage_initial<0 or storage_initial>1`

src/calliope/config/model_data_checks.yaml
```diff
@@ -29,6 +29,15 @@ fail:
...
+    message: "storage_initial is a fraction; values smaller than 0 are not allowed."
+
+  - where: flow_cap_min<0
```
brynpickering, line 35: tbh, the choice of what to catch here is increasingly difficult to justify. You could set this kind of requirement on pretty much every parameter. A better approach would be to extract the min/max values from the schema and to use those once the data is loaded in.

src/calliope/config/model_data_checks.yaml
```diff
@@ -29,6 +29,15 @@ fail:
...
+    message: "flow_cap_min has to be non-negative."
+
+  - where: (flow_cap_min AND flow_cap_max) AND flow_cap_max<flow_cap_min
```
brynpickering, line 38: should this be extended to every min/max param pair?

src/calliope/config/model_def_schema.yaml
```diff
@@ -675,7 +675,7 @@ properties:
             x-type: str
             title: Sink unit
             description: >-
-              Sets the unit of `Sink` to either `absolute` x-unit: energy), `per_area` x-unit: energy/area), or `per_cap` x-unit: energy/power).
+              Sets the unit of `Sink` to either `absolute` (e.g. kWh), `per_area` (e.g. kWh/m2), or `per_cap` (e.g. kWh/kW).
```
brynpickering, line 678: I'd stick with `energy/power/area`. We have found in the past that by defining specific scales of units - even in examples - users can confuse that with _the_ scale that must be used.

brynpickering (changes requested):

src/calliope/config/model_def_schema.yaml
sstroemer: That's what it currently looks like
image

and that's what (as an example) `source_unit` looks like
image

I assumed the second one was what it should look like. Just to clarify, do you mean both should read `... (energy) ...` (or `... (x-unit: energy) ...`?) instead?

src/calliope/config/model_data_checks.yaml
sstroemer: I'm not sure if this is the most efficient way to approach the problem, but the underlying idea would be: Parameter checks should catch all errors that may otherwise lead to something that a user possibly cannot easily understand. That is either something that (1) leads to an obscure internal error (e.g., a non-existing key in some lookup that is not directly related to the initial mistake), or (2) leads to an outcome that is hard to track back to the initial mistake.

Related to (2), an example, setting:

```yaml
storage_cap_max: 3000
storage_cap_min: 3001
```

for a tech, results in a `BackendWarning: Model solution was non-optimal.` (due to the infeasibility). Of course it's evident here, but for larger models, that may be generated from something, or based on parameters/config from somewhere this could be really hard to find (if one does not have access to an IIS calculation).

src/calliope/config/model_data_checks.yaml
sstroemer: Fixed in 576c13e1a585702d08c1332bb192a0e62b4378e0

src/calliope/config/model_data_checks.yaml
sstroemer: See the remark above. Maybe a general copy-pasted:

```yaml
where: (param_min AND param_min<LB) OR (param_max AND param_max>UB) OR (param_min AND param_max AND param_min>param_max)
message: Minimum or maximum value of parameter `param` are misconfigured.
```

That could be quickly replicated (only necessary replacements: `param`, `LB`, and `UB`) for each parameter that this would apply to, and serve as a dumb, but maybe sufficient, default check that could later on be made more explicit?

---

Alternatively, if the parser would support this, a lower- and upper-bound could be set for each parameter, which could then be automatically added to the documentation, and serve as check:

```yaml
flow_cap_min:
  default: 0
  min: 0
  max: min(.inf, flow_cap_max)
  x-type: float
  # ... other stuff
```

That would require the `min(.inf, flow_cap_max)` to gracefully handle the case that `flow_cap_max` does not exist (I suppose?).

brynpickering: Sadly, the schema wouldn't allow it

src/calliope/config/model_data_checks.yaml
brynpickering: This race to the bottom with data checking caused us quite some maintenance burden in v0.6, hence why we have removed it in favour of the user taking on more responsibility. We can't hope to capture _everything_ that they might do to create an infeasible model, especially once they add their own parameters and math. Even without their own math, there are complex combinations of pre-defined parameters that could lead to infeasibility (setting an resource_area_min, linking area and flow_cap, then setting a flow_cap_max that violates that, for instance).

src/calliope/config/model_def_schema.yaml
brynpickering: It should probably be `... (unit: energy) ...`. Looks like the `x-unit` bit has probably been caused by a replace-all.

sstroemer: Thanks! Updated that (also accordingly for `source_unit`).

src/calliope/config/model_data_checks.yaml
sstroemer: Perfectly fine, makes sense - just wanted to point out these immediate relations. I'll remove these from the PR, since they would otherwise just be inconsistent with other parameters, and the `< 0` does not matter anyways ...

src/calliope/config/model_data_checks.yaml
sstroemer: Too bad :/

sstroemer (commented):
Adjusted all files according to discussion.

src/calliope/config/model_def_schema.yaml
```diff
@@ -675,7 +675,7 @@ properties:
             x-type: str
             title: Sink unit
             description: >-
-              Sets the unit of `Sink` to either `absolute` x-unit: energy), `per_area` x-unit: energy/area), or `per_cap` x-unit: energy/power).
+              Sets the unit of `Sink` to either `absolute` (unit: `energy`), `per_area` (e.g. unit: `energy/area`), or `per_cap` (unit: `energy/power`).
               `per_area` uses the `area_use` decision variable to scale the sink while `per_cap` uses the `flow_cap` decision variable.
```
brynpickering, line 679:
```suggestion
              Sets the unit of `Sink` to either `absolute` (unit: `energy`), `per_area` (unit: `energy/area`), or `per_cap` (unit: `energy/power`).
              `per_area` uses the `area_use` decision variable to scale the sink while `per_cap` uses the `flow_cap` decision variable.
```

brynpickering (approved):
Just a minor suggestion, otherwise good to go.

I'm thinking we could do something about parameter-level checks as part of solving #642 (defining parameters as part of the YAML math, then having a "data-check" key to add checks at a parameter-level).

brynpickering:
@sstroemer please fix the test that is now failing! Plus, add your name to the `AUTHORS` file.

brynpickering (approved):

PR #655, opened 2024-08-07
tech/node groups to templates

### LINKED ISSUE #600: Change the name of 'tech_groups' to 'tech_class'
### What can be improved?

`tech_groups`, as well as `node_groups`, can be confusing names. According to the documentation, they are actually 'parent' `techs` or `nodes` that can be inherited by other techs rather than a group of different `techs` or `nodes`. 
We do not specifically propose for the name `tech_class`, but something similar and more explanatory to what it actually is.

### Version

v0.7

brynpickering:
Fixes #600 

I've merged `tech_` and `node_` into `templates` as there's no real reason to split them. You can access the same template from techs or nodes but the chance that it causes issues is minimal. The benefit is fewer top-level keys.

I've kept the reference to "inheriting" templates in the docs and the helper function. If there is a perference to purge all concept of inheritance, we could change it, but I don't have a good idea what to replace it with. 

* `tech_groups`/`node_groups` -> `templates`
* `inherit` -> `template`

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


tests/test_preprocess_model_data.py
irm-codebase, line 1: Nice-to-have: consider adding the small docstring to the modified tests, so that we know what they do in the future.

Feel free to skip.

irm-codebase (approved):
I must say: I love this change.  Very simple to understand, and will lead to users handling less terminology.

Bryn's idea to just merge everything into `templates` is excellent.

irm-codebase (approved):
Excellent!

brynpickering:
@sjpfenninger @irm-codebase one thing I missed is that you still end up with the two input variables in the xarray dataset: `techs_inheritance` and `nodes_inheritance`. For techs/nodes, they define the inheritance chain that was used to create that tech/node's definition. They need to be separated as they are indexed only over their respective dimensions. Should these be `tech_templates` and `node_templates`, or does `inheritance` make sense here?

irm-codebase:
@brynpickering looking at the code, it seems they achieve very similar things, but I do not think they match `template`'s purpose, since `template` is more of a 'user-side' thing, and this seems more 'internal'...

I think you are right in just going for `inheritance`.

docs/creating/techs.md
```diff
@@ -13,9 +13,10 @@ This establishes the basic characteristics in the optimisation model (decision v
 * `conversion`: Converts from one carrier to another.
 
 ??? info "Sharing configuration through inheritance"
```
sjpfenninger, line 15:
```suggestion
??? info "Sharing configuration through inheritance from templates"
```

sjpfenninger (dismissed):

sjpfenninger:
> @sjpfenninger @irm-codebase one thing I missed is that you still end up with the two input variables in the xarray dataset: `techs_inheritance` and `nodes_inheritance`. For techs/nodes, they define the inheritance chain that was used to create that tech/node's definition. They need to be separated as they are indexed only over their respective dimensions. Should these be `tech_templates` and `node_templates`, or does `inheritance` make sense here?

I agree with @irm-codebase that we can stick to inheritance here which marks this as a separate (but related) variable.

PR #656, opened 2024-08-07
Add `uses` cross-ref to math docs & collapsed admonitions

### LINKED ISSUE #651: Possible duplicate links
### Description

Decision variables contain a link to themselves in their `Used in: section`, which could be removed to make the list a bit shorter.

### Related links

- [`flow_cap`](https://calliope.readthedocs.io/en/latest/math/base/#flow_cap)
- other decision variables

### Version

latest (48148b814a8ba0cd64fa9b48decff4b5b60eb0d7)

### Proposed change

_No response_

brynpickering:
Fixes #651
Fixes point 3 in #643 

* Added `Uses` which is the inverse of `Used in`
* Added collapsed admonitions for `Uses` and `Used in` when building math docs with mkdocs features enabled
* Explicitly removed self-references in `Uses` and `Used in`

## Reviewer checklist

- [x] Documentation updated
- [x] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ### User-facing changes
 
+|new| Math component cross-references in both directions ("uses" and "used in") in Markdown math documentation (#643).
+
+|fixed| Duplicated links in math documentation (#651).
+
 |changed| `node_groups` and `tech_groups` changed to a general top-level `templates` key,
 accessed via the `template` key (replacing `inherit`) in `nodes` and `techs` (#600).
 
--- a/docs/hooks/generate_math_docs.py
+++ b/docs/hooks/generate_math_docs.py
@@ -122,7 +122,7 @@ def write_file(
 
     nav_reference["Pre-defined math"].append(output_file.as_posix())
 
-    math_doc = model.math_documentation.write(format="md", mkdocs_tabbed=True)
+    math_doc = model.math_documentation.write(format="md", mkdocs_features=True)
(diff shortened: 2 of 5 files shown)


brynpickering:
@sstroemer I can't add you as a review (will fix this), but welcome your comments

sstroemer:
That is really fancy! :astonished: 

---

Just one minor thing I noticed - you mentioned something related to "order of the entries" somewhere previously, not sure where exactly - when looking at [`flow_cap`](https://calliope--656.org.readthedocs.build/en/656/math/base/#flow_cap), it lists:

1. `flow_capacity_per_storage_capacity_min`
2. `ramping_down`
3. `area_use_per_flow_capacity`
4. ...

All of those are constraints, but are neither in the order that they appear in `base.yaml`, nor some other order I could guess (for stuff like `flow_cap_min` it also lists variables before constraints)? Maybe sticking to the same order as in the sidebar (if possible) would make it easier to read?

brynpickering:
> Maybe sticking to the same order as in the sidebar (if possible) would make it easier to read?

I have changed it to alphabetical because getting it in the same order as given in the math docs is a bit difficult. 
I have also alphabetically sorted the math docs in general. It is reasier to navigate (as you previously suggested @sstroemer) and it probably doesn't matter about getting the order right for defining global expressions as that isn't really what's important when communicating the math in this way.

sjpfenninger (approved):
Looks fantastic.

PR #666, opened 2024-08-19
Remove array copy

brynpickering:
Fixes slow decision variable creation (noticed on building a larger model)

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/backend/backend_model.py
+++ b/src/calliope/backend/backend_model.py
@@ -1071,13 +1071,14 @@ class BackendModel(BackendModelGenerator, Generic[T]):
                 name,
                 f"Applying bound according to the {bound} parameter values.",
             )
-            bound_array = self.get_parameter(bound).copy()
+            bound_array = self.get_parameter(bound)
             fill_na = bound_array.attrs.get("default", fill_na)
             references.add(bound)
         else:
             bound_array = xr.DataArray(bound)
-        bound_array.attrs = {}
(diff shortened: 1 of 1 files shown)


src/calliope/backend/backend_model.py
```diff
@@ -1071,13 +1071,14 @@ def _get_variable_bound(
...
-            bound_array = self.get_parameter(bound).copy()
+            bound_array = self.get_parameter(bound)
             fill_na = bound_array.attrs.get("default", fill_na)
             references.add(bound)
         else:
             bound_array = xr.DataArray(bound)
-        bound_array.attrs = {}
-        return bound_array.fillna(fill_na)
+        filled_bound_array = bound_array.fillna(fill_na)
+        filled_bound_array.attrs = {}
```
irm-codebase, line 1080: I am assuming this empty dict is necessary. Otherwise, i'd go with `del filled_bound_array.attrs`

irm-codebase (approved):

src/calliope/backend/backend_model.py
brynpickering: Why prefer `del` over `= {}`?

irm-codebase: I'm wrong here. It seems like `xarray` automatically puts an empty `dict` in `attrs` by default.
So your approach is correct.
image

PR #685, opened 2024-09-17
Update cost expression names & split out expressions

### LINKED ISSUE #645: Potential misinterpretation of (internal) cost calculation related to OM cost term
### What happened?

> Preface: The description below is way too long for how small the potential change is. It might be that I'm misreading
> the current status, then please ignore and just close this. It might also be entirely intended the way it currently is,
> in which case a small addition to the documentation of `cost_om_annual_investment_fraction` would render this obsolete.
> Text and calculations given at the end may be wrong, and include typos, etc. since time was limited, so please double-check.

---

## Description

Currently, the internal annualisation of costs is done in [`cost_investment`](https://calliope.readthedocs.io/en/latest/math/base/#cost_investment). The way it's done may be different to what some users may expect, based on the naming conventions, and may lead to confusion or misinterpretation of input data (and therefore wrong results).

The main motivation is the following observation of the calculation of `cost_investment`, which roughly looks like:

$$cost\textunderscore investment = \alpha \cdot (\delta \cdot total\textunderscore capex \cdot (1 + FOM_{\\%}) + FOM_{absolute})$$

Here $\delta$ is some sort of depreciation/annualisation factor ($\alpha$ just allows to generalize that for periods different to a single year) which weights the total capital expenditure (**correct**), but also the parts of the fixed operation and maintenance costs ($FOM$). The latter may not necessarily be expected by users, because these are often given as either `EUR/MW/a` (represented by `cost_om_annual`, here $FOM_{absolute}$, **correct**) or `%/a` (represented by `cost_om_annual_investment_fraction`, here $FOM_{\\%}$ - **potentially misleading**).

For both cases, it is a recurring fixed annual payment, which means it's already "per year" and does not have to be annualised. In the end, this leads to considerably reduced costs if using the "FOM are 2-5% depending on technology" approach (e.g., directly using _Danish Energy Agency_ data would be fine, using the processed _PyPSA technology-data_ would lead to wrong results, since they convert everything to %).

If that is actually a problem (and not intended), the easiest fix could be reworking to:

$$cost\textunderscore investment = \alpha \cdot ((\delta + FOM_{\\%}) \cdot total\textunderscore capex + FOM_{absolute})$$

Everything below just contains additional thoughts, that could be relevant when doing the (breaking) change.

## Thoughts

I'd like to point out / propose the following changes:

1. Renaming of the global "expressions", including some changes to the operating costs.
2. Changes to the calculation of the annualised costs, followed by changes to the operating costs.
3. Updates in the documentation to reflect the changes (not discussed, just mentioned for completeness).

### Renaming

1. [`cost_var`](https://calliope.readthedocs.io/en/latest/math/base/#cost_var) could be renamed to `cost_operation_var` (or something similar) - these are commonly referred to as **VOM**, variable operation and maintenance costs, which could be reflected by using `cost_vom`.
2. Introduction of `cost_operation_fix` for fixed costs of operation. Similar to the above, these are commonly referred to as **FOM**, fixed operation and maintenance costs, which could be reflected by using `cost_fom`.
3. Renaming of [`cost_investment`](https://calliope.readthedocs.io/en/latest/math/base/#cost_investment) to `cost_investment_total`, to better communicate that this is a (total) sum expression.
4. Finally, the introduction of `cost_investment_total_annualised`, indicating that this accounts for annualisation.

> Note 1: For consistency purposes, if doing a rename, it could be worth discussing to stick to a common convention for abbreviations. For example either: `cost_var` and `cost_invest`, or `cost_variable` and `cost_investment`; or any similar choice, but either both abbreviated or both as full names.

> Note 2: I get the choice of "depreciation" due to the first two branches of the calculations, but it may be worth to discuss whether the name "annualisation weight" may be confused with the calculation of `depreciation_rate` in the third case [link](https://calliope.readthedocs.io/en/latest/math/base/#cost_investment). This might be one of the main use cases (?), using an "annuity (factor)" for the annualisation of long-term costs under consideration of some cost-of-capital, whereas `annualisation_weight` only handles the proper scaling for non 1-year modeling periods.

#### Initial suggestion

1. `cost_vom` is given by the calculation of `cost_var` as it is.
2. `cost_investment_total` is given as "the inner sum of investment costs" of the current `cost_investment`, roughly `sum(cost_investment_flow_cap, over=carriers) + cost_investment_storage_cap + cost_investment_source_cap + cost_investment_area_use + cost_investment_purchase`.
3. `cost_fom = annualisation_weight * (sum(cost_om_annual * flow_cap, over=carriers) + cost_investment_total * cost_om_annual_investment_fraction)`.
4. `cost_investment_total_annualised = annualisation_weight * depreciation_rate * cost_investment_total`.
5. The final / overall costs are then updated to: `cost = cost_vom + cost_fom + cost_investment_total_annualised`.

Drawbacks:

- This introduces more global expressions.
- `annualisation_weight` is now needed in two separate locations.

#### An alternative way

This looks like a bigger change, which is why I did not want to propose this as the main way to go, but it may be worth considering (steps 1 and 2 are identical):

1. `cost_vom` is given by the calculation of `cost_var` as it is.
2. `cost_investment_total` is given as "the inner sum of investment costs" of the current `cost_investment`, roughly `sum(cost_investment_flow_cap, over=carriers) + cost_investment_storage_cap + cost_investment_source_cap + cost_investment_area_use + cost_investment_purchase`.
3. `cost_fom = sum(cost_om_annual * flow_cap, over=carriers) + cost_investment_total * cost_om_annual_investment_fraction`.
4. The final / overall costs are then updated to: `cost = cost_vom + annualisation_weight * (depreciation_rate * cost_investment_total + cost_fom)`.

This introduces less global expressions, and `annualisation_weight` is only needed in one location.

- [ ] macOS
- [ ] Windows
- [X] Linux

### Version

latest (4fc6b84fe3f17d6533765c5543192e773deb05a9)

### Relevant log output

_No response_

brynpickering:
Fixes #645 

The only thing I have chosen to _not_ implement from #645 is `cost_investment` -> `cost_investment_total`. I'm not sure that `_total` adds anything...

* the total capital cost is now `cost_investment` 
* the annualised capital cost is now `cost_investment_annualised`
* fixed operation costs are now `cost_operation_fixed`
* variable operation costs are now `cost_operation_variable`
* total `cost` is the sum of  `cost_investment_annualised`, `cost_operation_fixed`, and `cost_operation_variable`.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


brynpickering:
@sstroemer can you check that this is what you were thinking w.r.t. updating expressions?

src/calliope/math/base.yaml
```diff
@@ -874,27 +873,33 @@ global_expressions:
...
+          sum(default_if_empty(cost_investment_flow_cap, 0), over=carriers) +
+          default_if_empty(cost_investment_storage_cap, 0) +
+          default_if_empty(cost_investment_source_cap, 0) +
+          default_if_empty(cost_investment_area_use, 0) +
+          default_if_empty(cost_investment_purchase, 0)
+
+  cost_investment_annualised:
+    title: Equivalent annual investment costs
+    description: >-
+      An annuity factor has been applied to scale lifetime investment costs to annual values that can be directly compared to operation costs.
```
sstroemer: Maybe the use of `annualisation_weight` could be included here, to prevent someone taking the _[...] to annual values [...]_ too literally?

Example:

> An annuity factor has been applied to scale lifetime investment costs to annual values that can be directly compared to operation costs. If the modeling period is not equal to one full year, this will be scaled accordingly.

src/calliope/math/base.yaml
```diff
@@ -906,6 +911,23 @@ global_expressions:
             (cost_interest_rate * ((1 + cost_interest_rate) ** lifetime)) /
             (((1 + cost_interest_rate) ** lifetime) - 1)
 
+  cost_operation_fixed:
+    title: Total fixed operation costs
+    description: >-
+      The fixed, annual operation costs of a technology, which are calculated relative to investment costs.
```
sstroemer: Same comment as above. Example:

> The fixed, annual operation costs of a technology, which are calculated relative to investment costs. If the modeling period is not equal to one full year, this will be scaled accordingly.

sstroemer:
> @sstroemer can you check that this is what you were thinking w.r.t. updating expressions?

@brynpickering Looks really good!

sjpfenninger (approved):
Looks good to me, much clearer than before!

PR #687, opened 2024-09-25
Add data source dimension name mapping option

### LINKED ISSUE #680: Allow mappings for tabular data
### What can be improved?

Being able to load tabular data is a wonderful new feature.
But it is currently limited by forcing the data to 'match' calliope's dimension names in certain cases. Enforcing strict naming by default makes a lot of sense: you avoid ambiguity and you also avoid the risks of relying on column position.

However, it will often be too inflexible.

## Reasoning

Some of our names might lead to files being less human readable, or finicky: 
- if a model is national resolution, `nodes` is less informative than `country`
- people will often go for `technology` or `tech` instead of `techs`, since it can be intuitive to name columns in singular
- people might prefer `time` or `utc_timestamp`, over `timesteps`...

This will lead to a lot of 'boilerplate' code that just shapes the data to fit Calliope's naming. See the following 3 examples for different names used for timeseries in Euro Calliope with v6.10:

image
image
image

All 3 are equally 'human' readable, but since they do not specify `timesteps`, they won't load into Calliope.

# Proposal

An option to use mappings would solve this issue.
For example, you could load one of the timeseries above this way:

```
data_sources:
  demand_elec_timeseries:
    source: timeseries/demand/electricity.csv
    columns: nodes
    rows: {timesteps: time}
    add_dims:
      techs: demand_elec
      parameters: sink_use_equals
```

This is still strict, but more flexible.

### Version

v0.7.0.dev3

brynpickering:
Fixes #680 

The only question I have is whether it is better described as `rename_dims` instead of `map_dims`.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


brynpickering:
@irm-codebase `map_dims` or `rename_dims`?

irm-codebase:
@brynpickering I like `map_dims` a bit more.

`rename_dims` is a bit ambiguous (is it renaming the model's dims? the data's?). `map_dims` is a bit clearer, and kind of similar to some `pandas` functionality, so I brain likes it more  ¯\_ (ツ)_/¯.

brynpickering:
@sjpfenninger I'll let you cast your vote on the naming. If it's unanimous then we'll stick with `map_dims`.

brynpickering:
On the side of `rename_dims`, it does align with pandas [`rename_axis`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.rename_axis.html) and xarray [`rename_dims`](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.rename_dims.html)

irm-codebase:
Hmmm... I see the point.

I think the difference is that in pandas and xarray you affect the dataframe / dataarray itself (i.e., the 'thing' the library handles), and my mind links the command to `calliope`, not `pandas` (`timesteps` is still `timesteps`, we are just telling `data_tables` to read `time` as `timesteps`). 

But looking at how the `yaml` file would look like, it's perhaps less ambiguous than I thought.
`rename_dims` is perfectly fine!

irm-codebase (dismissed):
Looking good.
Feel free to refactor to `rename_dims`, too!

sjpfenninger:
I vote for `rename_dims`: I think it's clearer in this context

PR #695, opened 2024-10-08
Add templates for data tables

### LINKED ISSUE #676: Data sources and templates
### What can be improved?

I am not entirely sure if this is possible in the new math, so I'm filing this as a _potential_ feature request.
Otherwise, it might be a documentation oversight...

------------------------

In v0.6.10 it used to be possible to set data sources for `tech_groups` (now `templates`):

```yaml
tech_groups:
    wind_onshore:  # from [@JRC:2014] Table 4
        essentials:
            name: Onshore wind
            carrier: electricity
            parent: supply
        constraints:
            resource: file=supply/capacityfactors-wind-onshore.csv
            resource_unit: energy_per_cap
            lifetime: 25
```

Since data sources are now a separate section, I assume this is no longer possible, since you need to provide dimensions to them (which `templates` do not have, since the `tech` is set later).

Is there a way to support this?
People can define this for each tech, but it diminishes the usefulness of templates a bit.

### Version

v0.7.0.dev3

brynpickering:
Fixes #676 

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


src/calliope/util/tools.py
```diff
@@ -47,3 +51,52 @@ def listify(var: Any) -> list:
     else:
         var = [var]
     return var
+
+
+def climb_template_tree(
+    dim_item_dict: "AttrDict",
```
sjpfenninger, line 57: Call this `input_dict`, might be easier to understand?

src/calliope/util/tools.py
```diff
@@ -47,3 +51,52 @@ def listify(var: Any) -> list:
...
+    Model definition elements can inherit template entries (those in `templates`).
+    Template entries can also inherit each other, to create an inheritance chain.
+
+    This function will be called recursively until a definition dictionary without `template` is reached.
+
+    Args:
+        dim_item_dict (AttrDict): Dictionary (possibly) containing `template`.
+        templates (AttrDict): Dictionary of available templates.
+        item_name (str):
+            The current position in the inheritance tree.
```
sjpfenninger, line 73:
```suggestion
            The current position in the inheritance tree. Used only for a more expressive KeyError.
```
(maybe this could be an optional argument?)

src/calliope/util/tools.py
```diff
@@ -47,3 +51,52 @@ def listify(var: Any) -> list:
...
+    templates: "AttrDict",
+    item_name: str,
+    inheritance: list | None = None,
+) -> tuple["AttrDict", list | None]:
+    """Follow the `template` references from model definition elements to `templates`.
+
+    Model definition elements can inherit template entries (those in `templates`).
+    Template entries can also inherit each other, to create an inheritance chain.
+
+    This function will be called recursively until a definition dictionary without `template` is reached.
```
sjpfenninger, line 67: The wording here in the docstring suggests that the function calls itself recursively, but that doesn't seem to actually be the case.

sjpfenninger (changes requested):
A great idea. Just some comments on making `climb_template_tree()` easier to understand.

src/calliope/util/tools.py
brynpickering: It is the case, there is a call to `climb_template_tree` inside `climb_template_tree`.

sjpfenninger (approved):

PR #706, opened 2024-11-08
Update description; add image

### LINKED ISSUE #699: Fixing Circular Definitions in Docs
### Description

This was in the getting familiar with Calliope definitions. These definitions are self referential and relatively difficult to understand without real examples. 

Current:
Technology: a technology that produces, consumes, converts or transports carriers.
Node: a site which can contain multiple technologies and which may contain other nodes for carrier balancing purposes.
Source: a source of commodity that can (or must) be used by a technology to introduce carriers into the system.
Sink: a commodity sink that can (or must) be used by a technology to remove carriers from the system.
Carrier: a carrier that groups technologies together into the same network, for example electricity or heat.

Also, what does "grouping technologies" mean? The example, "electricity or heat" does not constitute a grouping. 

### Related links

https://calliope.readthedocs.io/en/latest/creating/

### Version

v0.7.0

### Proposed change

I propose that each definition should not be self referential and also have a tangible example. One definition may be:

Technology: an entity that produces, consumes, converts or transports carriers.
- You could, for instance, define a solar technology to produce, battery to cosume energy...

brynpickering:
Fixes #699 

Updates to docs to improve legibility when first introducing a Calliope model. Includes a new visualisation of the system.

@jeisenman23 you can see what it looks like rendered if you click on the `readthedocs` CI tab at the bottom of this page.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


brynpickering:
@FLomb and @irm-codebase would benefit from your review on this while @sjpfenninger is away - does it make things easier to understand?

docs/creating/index.md
```diff
@@ -6,14 +6,24 @@ The model specifies one or more **nodes** along with the technologies allowed at
 **Demand technologies** remove carriers from the system through a **sink**, while **storage technologies** can store carriers at a specific node.
 Putting all of these possibilities together allows a modeller to specify as simple or as complex a model as necessary to answer a given research question.
 
+![Visual description of the Calliope terminology.](../img/description_of_system.svg)
+
 ??? info "An overview of the Calliope terminology"
-    The terminology defined here is used throughout the documentation and the model code and configuration files:
+    The terminology defined here is used throughout the documentation, and the model code and configuration files:
```
irm-codebase, line 12: I would not hide this section with `???`. This is quite important, and users might miss it.

docs/img/description_of_system.svg
irm-codebase, line 1: Some suggestions:
- I'd consider reducing the amount of text... you already have the info section.
- consider using the terminology users will face when configuring a model (supply, demand, node, etc.)
- I'd avoid using both rectangles and circles for the nodes. It's a tad confusing.
- I'd use the same style as other sections in the documentation. The national example is really nice, visually: https://calliope.readthedocs.io/en/v0.7.0.dev4/examples/national_scale/

Here is a draft. Unfortunately I do not know which program was used for the national example, so it is by no means a direct replacement.
image

docs/creating/index.md
```diff
@@ -6,14 +6,24 @@ The model specifies one or more **nodes** along with the technologies allowed at
...
+    * **Technology**: an entity that produces, consumes, converts, stores or transports _carriers_.
+    > E.g., a _wind turbine_ that produces electricity, a _heat pump_ that converts electricity to heat, or a _battery_ that stores electricity.
+    * **Node**: a geographic point or area which can contain multiple _technologies_ and at which _carrier_ flows in and out must balance.
+    > E.g., a building in a district, an administrative region in a country, a country in a continent.
+    * **Source**: a commodity that can (or must) be used by a _technology_ to introduce carriers into the system.
+    Unlike _carriers_, _sources_ are not explicitly tracked as they are outside the system boundary.
+    > E.g., energy from the sun to power a solar panel, natural gas to power a gas turbine.
+    * **Sink**: a commodity that can (or must) be used by a _technology_ to remove carriers from the system.
+    Unlike _carriers_, _sinks_ are not explicitly tracked as they are outside system boundary.
+    > E.g., electricity consumed by household appliances, liquid fuel consumed by vehicles.
```
irm-codebase, line 25: This section is a better explanation than the text that starts this section (In short, a Calliope model works....).

How about trimming that text, and making it a  friendlier introduction?

----

A Calliope model is a collection of interconnected technologies, nodes and carriers describing a real world energy system. **Technologies** can be anything in between transmission lines, storage facilities, power supplies or energy demanding appliances. **Nodes** are groupings of technologies describing a region of the world, which could be a country, municipality or a single house. Finally, **carriers** describe the different types of energy being produced, consumer or transmitted. This could be electricity, heat or fuels needed for all kinds of different activities.

Putting all of these possibilities together allows a modeller to create a model that is as simple or as complex as necessary to answer a given research question, and to do so in a way that is easy to understand to other researchers thanks to Calliope's straightforward syntax.

irm-codebase (commented):
Some suggestions for this section

docs/img/description_of_system.svg
brynpickering: Not considering technologies as being "inside" nodes is a source of confusion in its own right. I prefer the visualisation that there are flows _within_, _between_, and _to_/_from_ nodes. 

Renaming to base tech types makes sense. As does having a key.

docs/creating/index.md
FLomb: I like the more straightforward version you propose, @irm-codebase, but I guess that the original one is more generalisable as in there is no mention of specific energy-related terms such as 'lines', 'power supplies', 'energy', etc. This is possibly meant to keep open the option to model _whatever_ system in Calliope, which makes sense since it has already happened several times.

I would try to strike a balance between too-generic (and therefore, not straightforward) descriptions and too-energy-focussed ones. Perhaps one solution would be to make it explicit that the framework is generalised to to other systems and flows than energy ones, but we make an example description for an energy system as the most typical application.

docs/img/description_of_system.svg
FLomb: I agree that having techs inside nodes is desirable for understandability. And yet, they are outside the nodes in the visualisations of the tutorials. So, whatever we choose, it is desirable to have uniformity across all visualisations.

docs/creating/index.md
brynpickering: Danger is that there's repetition, right?

jeisenman23: Hey, repetitive or not, I love these changes!

irm-codebase: @brynpickering yep, repetition would be the main issue. I think we should aim to be as concise as possible with the docs, to avoid leaving old stuff as the software changes.

docs/img/description_of_system.svg
brynpickering: Calliope drawings - Frame 1

brynpickering: I agree that the "nodal balance" part is a bit out of place. Any ideas on how to simplify that part?

brynpickering: Calliope drawings - Frame 1(1)

irm-codebase: This is really good!
The last change I'd make is to add an arrow towards 'sink', to keep it consistent with the rest of the graph.

docs/creating/index.md
brynpickering: OK, I've gone with what is hopefully both more concise and clearer.

docs/creating/index.md
```diff
@@ -1,35 +1,22 @@
...
-    * **Source**: a commodity that can (or must) be used by a _technology_ to introduce carriers into the system.
-    Unlike _carriers_, _sources_ are not explicitly tracked as they are outside the system boundary.
-    > E.g., energy from the sun to power a solar panel, natural gas to power a gas turbine.
-    * **Sink**: a commodity that can (or must) be used by a _technology_ to remove carriers from the system.
-    Unlike _carriers_, _sinks_ are not explicitly tracked as they are outside system boundary.
-    > E.g., electricity consumed by household appliances, liquid fuel consumed by vehicles.
+**Nodes** contain groups of _technologies_ and are usually geographic, e.g., a country, municipality or a single house.
 
+Flows can enter the system from **Sources**, e.g., energy from the sun to power a solar panel, and can exit it into **Sinks**, e.g., electricity consumed by household appliances.
+Unlike _carriers_, we do not explicitly track the type of commodity described by sources and sinks.
```
irm-codebase, line 14: Optional: I recommend not capitalising sources and sinks.
Mostly because, in the code, we do not really require it. I find that keeping this consistency helps people tie things to the code.

Even the new image keeps lowercase.

irm-codebase (approved):
Approved!

There is one small nice to have. Otherwise this looks good.

docs/creating/index.md
brynpickering: good catch, they were copy-past jobs from the old section.

irm-codebase (approved):

PR #707, opened 2024-11-12
Add simplistic mypy configuration

irm-codebase:
This PR attempts to bring a bit more standardisation on the way we use mypy.

I've noticed that our current usage of the tool is varied, and may result in back and forth on some changes.
This attempts to reduce redundant changes by using a simplistic 'hands-off' mypy configuration.

Notably, I do not add any CI testing for it, as it would fail. Instead, this attempts to help us clean up the code over time.

* Added mypy to the `dev.txt` dependencies, using the newest version available.
* Added minimalistic configuration for the tool in `pyproject.toml`.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,10 @@ count = ''
 quiet-level = 3
 ignore-words-list = "socio-economic"  # British english spelling that isn't covered by the inbuilt dictionary
 
+[tool.mypy]
+ignore_missing_imports = true
+files = "src/"
+
 [tool.setuptools.packages.find]
 where = ["src"]
 include = ["calliope*"]
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -6,6 +6,7 @@ mkdocs-jupyter >= 0.24, < 0.24.7
 mkdocs-macros-plugin >= 1.0, < 2
 mkdocs-material >= 9.5, < 10
 mkdocstrings-python >= 1.7, < 2
+mypy >= 1.13.0, < 2
 pandas-stubs
 plotly >= 5, < 6
 pre-commit < 4


irm-codebase:
This `mypy` configuration throws out ~20 errors.
Over time, these should diminish, but for now I think it's OK if we just generally align on how the tool is used throughout our code.

https://github.com/user-attachments/assets/c080e002-a48a-42e7-ac2b-053e442586ef

irm-codebase:
Seems like CI fails due to a separate issue caused by the newest gurobi version (https://github.com/calliope-project/calliope/issues/710)

pyproject.toml
```diff
@@ -75,6 +75,10 @@ count = ''
 quiet-level = 3
 ignore-words-list = "socio-economic"  # British english spelling that isn't covered by the inbuilt dictionary
 
+[tool.mypy]
+ignore_missing_imports = true
```
brynpickering, line 79: why do we want to ignore missing imports?

brynpickering (commented):
This change makes perfect sense! Just need to decide on the mypy config, i.e. what do we want to ignore.

pyproject.toml
irm-codebase: In most cases, pyomo will mark imported libraries that do not use it as a warning/errors.
This leads to a lot of "mypy spam" that only goes away by using comments or by adding unofficial 'stub' libraries.
The first is bothersome, the second would inflate our `dev.txt` file.

Since the amount of false positives is often quite big (it more than quintuples in our case), I often turn it off.

This still allows us to check our own function annotations using mypy, which is the main goal, I think.

brynpickering (approved):

PR #711, opened 2024-11-15
Add config obj:  simplification suggestions

irm-codebase:

Some suggestions stemming from my review of the `pydantic` proposal. This mostly focuses on removing stuff I think may not be needed to save us some complexity.

* Some slight naming changes for `this_build_config` and similars. Since the configuration is updated, I think just using `build_config` makes sense.
* Removed `mode` setting from `solve`. We only really use it for one comparison, so just using the one in `build` saves a lot of code.
* Added `pydantic` to the `base.txt` environment
* Moved `config` import out of `TYPE_CHECKING`
* Made the config schema extraction more flexible, and avoided redundancies of its use in the documentation
* Split `AttrDict.to_yaml()` into into serialisation and file generation.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/attrdict.py
+++ b/src/calliope/attrdict.py
@@ -329,7 +329,7 @@ class AttrDict(dict):
             d[k] = self.get_key(k)
         return d
 
-    def as_yaml_str(self) -> str:
+    def to_yaml(self, path: str | None = None) -> str:
         """Return a serialised YAML string."""
         result = self.copy()
         yaml_ = ruamel_yaml.YAML()
@@ -357,17 +357,12 @@ class AttrDict(dict):
 
         stream = io.StringIO()
         yaml_.dump(result, stream)
-        return stream.getvalue()
-
-    def save_yaml(self, path: str) -> None:
-        """Save AttrDict as a yaml file.
-
-        Args:
-            path (str): path of saved YAML.
-        """
-        yaml_str = self.as_yaml_str()
-        with open(path, "w") as f:
-            f.write(yaml_str)
+        yaml_str = stream.getvalue()
+        if path:
+            yaml_str = self.to_yaml()
+            with open(path, "w") as f:
+                f.write(yaml_str)
+        return yaml_str
 
     def keys_nested(self, subkeys_as="list"):
         """Returns all keys in the AttrDict, including nested keys.
--- a/src/calliope/attrdict.py
+++ b/src/calliope/attrdict.py
@@ -329,12 +329,8 @@ class AttrDict(dict):
             d[k] = self.get_key(k)
         return d
 
-    def to_yaml(self, path=None):
-        """Conversion to YAML.
-
-        Saves the AttrDict to the ``path`` as a YAML file or returns a YAML string
-        if ``path`` is None.
-        """
+    def as_yaml_str(self) -> str:
+        """Return a serialised YAML string."""
         result = self.copy()
         yaml_ = ruamel_yaml.YAML()
         yaml_.indent = 2
@@ -359,13 +355,19 @@ class AttrDict(dict):
         # handle multi-line strings.
         walk_tree(result)
 
-        if path is not None:
-            with open(path, "w") as f:
-                yaml_.dump(result, f)
-        else:
-            stream = io.StringIO()
-            yaml_.dump(result, stream)
-            return stream.getvalue()
+        stream = io.StringIO()
+        yaml_.dump(result, stream)
+        return stream.getvalue()
+
+    def save_yaml(self, path: str) -> None:
+        """Save AttrDict as a yaml file.
+
+        Args:
+            path (str): path of saved YAML.
+        """
+        yaml_str = self.as_yaml_str()
+        with open(path, "w") as f:
+            f.write(yaml_str)
 
     def keys_nested(self, subkeys_as="list"):
         """Returns all keys in the AttrDict, including nested keys.
--- a/src/calliope/attrdict.py
+++ b/src/calliope/attrdict.py
@@ -359,7 +359,6 @@ class AttrDict(dict):
         yaml_.dump(result, stream)
         yaml_str = stream.getvalue()
         if path:
-            yaml_str = self.to_yaml()
             with open(path, "w") as f:
                 f.write(yaml_str)
         return yaml_str
--- a/src/calliope/config.py
+++ b/src/calliope/config.py
@@ -82,19 +82,15 @@ class ConfigBaseModel(BaseModel):
         self._kwargs = update_dict
         return updated
 
-    def model_json_schema(self, replace_refs=False) -> AttrDict:
-        """Generate an AttrDict with the schema of this class.
-
-        Args:
-            replace_refs (bool, optional): If True, replace $ref/$def for better readability. Defaults to False.
+    def model_no_ref_schema(self) -> AttrDict:
+        """Generate an AttrDict with the schema replacing $ref/$def for better readability.
 
         Returns:
             AttrDict: class schema.
         """
         schema_dict = AttrDict(super().model_json_schema())
-        if replace_refs:
-            schema_dict = AttrDict(jsonref.replace_refs(schema_dict))
-            schema_dict.del_key("$defs")
+        schema_dict = AttrDict(jsonref.replace_refs(schema_dict))
+        schema_dict.del_key("$defs")
         return schema_dict
 
     @property
--- a/src/calliope/config.py
+++ b/src/calliope/config.py
@@ -342,28 +342,3 @@ class CalliopeConfig(ConfigBaseModel):
     init: Init = Init()
     build: Build = Build()
     solve: Solve = Solve()
-
-    @model_validator(mode="before")
-    @classmethod
-    def update_solve_mode(cls, data):
-        """Solve mode should match build mode."""
-        data["solve"]["mode"] = data["build"]["mode"]
-        return data
-
-    def update(self, update_dict: dict, deep: bool = False) -> Self:
-        """Return a new iteration of the model with updated fields.
-
-        Updates are validated and stored in the parent class in the `_kwargs` key.
-
-        Args:
-            update_dict (dict): Dictionary with which to update the base model.
-            deep (bool, optional): Set to True to make a deep copy of the model. Defaults to False.
-
-        Returns:
-            BaseModel: New model instance.
-        """
-        update_dict_temp = AttrDict(update_dict)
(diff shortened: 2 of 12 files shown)


irm-codebase:
@brynpickering ready for review!

All the changes relate to my suggestions. Most of this aims to reduce lines of code

src/calliope/attrdict.py
brynpickering, line 1: I'm not sure I agree with this change. It's quite common for these kinds of dictionary wrappers to have a "to_yaml" method which accepts a filename and returns a string if it doesn't receive a filename (see [Box](https://github.com/cdgriffith/Box/wiki/Converters#YAML) and [benedict](https://github.com/fabiocaccamo/python-benedict?tab=readme-ov-file#to_yaml)).

src/calliope/config.py
```diff
@@ -82,27 +82,20 @@ def update(self, update_dict: dict, deep: bool = False) -> Self:
...
 
-    @overload
-    def model_yaml_schema(self, filepath: str | Path) -> None: ...
-
-    @overload
-    def model_yaml_schema(self, filepath: None = None) -> str: ...
-
-    def model_yaml_schema(self, filepath: str | Path | None = None) -> None | str:
-        """Generate a YAML schema for the class.
+    def model_json_schema(self, replace_refs=False) -> AttrDict:
```
brynpickering, line 85: I would _not_ override a parent method. We can just remove this entirely and have a utility function that converts the json schema output to a referenceless schema.

src/calliope/model.py
```diff
@@ -341,26 +341,23 @@ def solve(self, force: bool = False, warmstart: bool = False, **kwargs) -> None:
         else:
             to_drop = []
 
-        kwargs["mode"] = self.config.build.applied_keyword_overrides.get(
-            "mode", self.config.build.mode
-        )
-
-        this_solve_config = self.config.update({"solve": kwargs}).solve
+        solve_config = self.config.update({"solve": kwargs}).solve
+        mode = self.config.build.mode
```
brynpickering, line 345: This isn't valid. If a user updated the mode as a kwarg to `build` then `self.config.build.applied_keyword_overrides` will store that update, but `self.config.build` will not. This allows `self.config.build` to always be used to create a new temporary build config on each call to `build` (i.e., kwargs are temporary and so do not get stored in the core config, only in the separated kwarg dict). So, here you'd need to call `self.config.update(self.config.applied_keyword_overrides).build.mode`. Or we add a new config object method which returns the config with kwarg overrides applied automatically.

brynpickering (changes requested):
I like the cleanliness achieved in most parts, however I'm not sure of the switch to two different `yaml` dumping methods and build config `mode` attr needs to be correctly accessed.

src/calliope/attrdict.py
irm-codebase: I generally prefer to keep functions as small as possible because it keeps our code slim, but it's a style thing.
Still, I'll revert to the old form.

src/calliope/model.py
irm-codebase: @brynpickering This is an issue, yeah... keeping `config.build` to the initial setup complicates things a bunch, it seems.

I think I have two suggested solutions:

### Solution 1
Changing `model._is_built` from a bool to a string flag that holds the latest built mode, which is set `build()`

This way, if `solve` is ran:
- if `_is_built` is `None` -> throw an error
- if `_is_built` does not match `build.mode`: solve but throw a warning to notify that the mode was changed
- if `_is_built` matches `build.mode`: just solve normally

### Solution 2

Just updating build directly, changing `applied_keyword_overrides` to `first_configuration` (or a better name), which holds the first ever configuration, and adding a method that 'recovers' the initial values.

src/calliope/attrdict.py
irm-codebase: The only change I made is always returning the string, even if the path is saved.
It's just lost if not stored in a variable, which is fine IMO.

src/calliope/model.py
brynpickering: I guess we don't want to give `mode` such a special position in things - it could be that other config elements need to be passed between the two stages in future. From that perspective, I wouldn't use solution 1. 

I like having `build.config` always reflecting the loaded, first config and then having the ad-hoc overrides living elsewhere. Then, every time you want to re-build / re-solve, you know what base config you are working with and don't accidentally expect some former kwarg config to still be in place. However, it's easily missed. The reason for not actually updating the pydantic model is that updating it in-place is a pain if you have frozen it, a feature I would keep (i.e., you don't allow direct updates to the config options - `config.solve.solver = "gurobi"` is not allowed). Instead, you get a copy of the model, and constantly updating the config object with a copy of itself is a bad idea.

src/calliope/attrdict.py
```diff
@@ -359,13 +355,14 @@ def to_yaml(self, path=None):
...
+        yaml_str = stream.getvalue()
+        if path:
+            yaml_str = self.to_yaml()
             with open(path, "w") as f:
-                yaml_.dump(result, f)
-        else:
-            stream = io.StringIO()
-            yaml_.dump(result, stream)
-            return stream.getvalue()
+                f.write(yaml_str)
```
brynpickering, line 364: why do you need `self.to_yaml()` here? you can use the constructed `yaml_str` directly. You can also use pathlib to avoid the context manager: `Path(path).write_text(yaml_str)`

src/calliope/model.py
irm-codebase: I am not sure what you mean by frozen in this case... I can easily alter the configuration by just writing to it:
image

I'd argue that having the option to 'reset' the model configuration to the initial state is similarly simple to understand (although a change from the current norm). We could even add methods to save both configurations (latest and initial) when calling `model.save_netcdf()`. This also brings the additional advantage that the saved data and the 'loaded' configuration will always correspond.

No approach will save us from the fact that python just lets you modify stuff as you see fit. Here is the current config again, but with a wrongful value:
image

irm-codebase: The question here is how to keep the current approach manageable and reduce future accidental bugs (as I did here...). I think we should shift the way we see the problem to 'let's ensure things are right when needed' rather than 'lets shield the config from everything'. 

`pydantic` allows us to easily re-run the validation whenever needed. But this only easily achievable if the config object is up-to-date:

image

src/calliope/attrdict.py
irm-codebase: Whoops. That's what you get from copy pasting!

src/calliope/model.py
brynpickering: OK, I didn't actually freeze `build` and `solve` in the end! But you can't do `model.config.init.name = "new_name"` as there is a way to stop pydantic model attributes being set on-the-fly. I planned for this to extend to `build` and `solve`, it's just a matter of adding `"frozen": True` to the pydantic model config.

RE re-validating, that's what the new `update` method does. It applies new config and forces re-validation, returning the new pydantic model with the changes made. I had intended for this to be the only way to allow users to update the config (i.e., `model.config.build.mode = "give coconut"` would not be allowed), so that we get re-validation _every_ time. It's a runtime hit, but negligible given how small the pydantic model is.

The question then is whether we update in-place or not. [Returning an updated copy of the model is what pydantic recommends when models are otherwise frozen](https://docs.pydantic.dev/2.9/concepts/serialization/#model_copy). So, we can freeze it and shield the config by giving only one route for config updates which are _always_ re-validated. But we have a copy of the config returned to us rather than it updating in-place. This seems to be more the direction projects are going in (e.g. xarray has moved to returning copies of objects rather than allowing in-place operations). We could make the model with the current kwarg overrides applied more easily accessible with something like `model.config.with_kwargs_overrides`...?

irm-codebase: Oh! If freezing the config is possible, then avoiding in-place modifications is the way to go.

One last argument for making `model.config` reflect the current state rather than the initial state: I'd say it's more 'natural' to how most code / programs operate. I suspect most people will expect similar behaviour.

If someone does `model.config = model.config.update(my_kwargs)`, they'd most likely expect their changes to be reflected in `model.config`, rather than `model.config._kwargs` or `model.config.with_overrides`.

We can go either way, though. This is just minor preferences.

brynpickering: Yeah, `model.config = model.config.update(my_kwargs)` would be feasible if we didn't freeze that level, but it isn't an _in-place_ operation (just `model.config.update(my_kwargs)` and expecting contents of config to be updated would be). But it feels safer to not allow `model.config` to be updated either (otherwise we have to validate that it is an appropriate type! What would stop you from doing `model.config = "foo"` other than the fact that model.config itself can be a frozen pydantic model?). 

I should say that I agree with you entirely that having to call/check `model.config.with_kwargs_overrides` is an annoyance that might cause issues down the line, so I'm keen to iterate to find a better solution to it!

irm-codebase: Ok. I think this will start taking form once the whole approach to `config` is set up.
Here is my proposal for now: I use `self.config.update(self.config.applied_keyword_overrides).build.mode` in this PR to get it in the other one.

Then we decide the final approach once the `CalliopeConfig` object is fully complete (also with techs, nodes, data tables, etc)?

irm-codebase: I already have `data_tables` up and going (including testing!) [here](https://github.com/calliope-project/calliope/tree/add-config-obj-replace-schemas). I think it's better to put a pin on this until the rest is finished, to make a better decision.

brynpickering: Sounds good @irm-codebase . I suspect it will affect the config and the math. Both are places where we can update the dict via kwargs at different stages in the process. The data tables won't be so affected by it, I hope!

irm-codebase: This is how I think the configuration object would end up looking like:

```python
class CalliopeConfig(ConfigBaseModel):
    """Calliope configuration class."""
    model_config = {"title": "config"}
    init: Init = Init()
    build: Build = Build()
    solve: Solve = Solve()
    techs: dict[str, Tech]
    nodes: dict[str, Node]
    overrides: dict[str, Self] = None
    scenarios: dict[str, UniqueList[str]] = None
    data_tables: dict[str, DataTable] = None
```

I already made sure that `data_tables` are possible. I think the most complicated bits will be overrides, templates (which we'll most likely need to handle pre-pydantic) and some aspects of the math.

irm-codebase:
@brynpickering temporary fix is in. Please let me know if there is anything else you'd like me to change.
I'll continue with sections beyond the `data_tables` meanwhile.

brynpickering (approved):

PR #715, opened 2024-11-19
Add `broadcast_param_data` config option and default it to False.

### LINKED ISSUE #615: Unwanted spillover and overwriting across cost dimensions
### What happened?

In the new version, the addition of a cost dimension requires extreme care. It is very easy for a new cost dimension to end up overriding existing cost dimensions, even if this was not the intention. 

Let's assume we define a new `spores_score` for the technology  `ccgt` in the national-scale example, like this:     
```
techs:
        ccgt.cost_flow_cap: {'data': 0, 'index': 'spores_score', 'dims': 'costs'}

```
Then this would overwrite the `monetary` cost class, which I should have explicitly re-defined with the same cost as in the default `techs.yaml` file. Now, this may seem little extra effort, but it may be daunting for very large models (such as Euro-Calliope) to ensure that this repetition always happens. And also, it makes the override files unnecessarily large.

Moreover, the model does not provide any warning about this, leading to a model solution in which (without noticing) the `ccgt` ended up having no monetary cost.

- [ ] macOS
- [ ] Windows
- [X] Linux

### Version

v0.7.0dev3

### Relevant log output

_No response_

brynpickering:
Fixes #615 

* `config.init.broadcast_param_data` boolean option available as per suggestion from @irm-codebase. It defaults to False, so must be set knowingly by the user to True to allow for indexed parameter data broadcasting.
* Updated example models to use CSV cost data (national scale) and to have a top-level interest rate (since we don't change it per technology in our example models).
* Updated docs to include this info and reflect updates to example models.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/docs/examples/national_scale/index.md
+++ b/docs/examples/national_scale/index.md
@@ -1,3 +1,10 @@
+---
+costs:
+    file: "src/calliope/example_models/national_scale/data_tables/costs.csv"
+    header: [0, 1]
+    index_col: 0
+---
+
 # National Scale Example Model
 
 This example consists of two possible power supply technologies,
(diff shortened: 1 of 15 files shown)


docs/examples/national_scale/index.md
```diff
@@ -25,6 +32,29 @@ It does not contain much data, but the scaffolding with which to construct and r
...
+To do this we reference data tables under the `data_tables` key:
+
+```yaml
+--8<-- "src/calliope/example_models/national_scale/model.yaml:data-tables"
+```
+
+In the Calliope national scale example model, we load both timeseries and cost data from file.
+As an example, the data in the cost CSV file looks like this:
+
+{{ read_csv(page.meta.costs.file, header=page.meta.costs.header, index_col=page.meta.costs.index_col) }}
```
FLomb, line 47: If I look at the 'built' markdown file here on GitHub, the links to files and to the header do not work properly, but I assume this is because these files are built into documentation by Calliope in a different way than simple markdown visualisation. So perhaps someone else who knows how to build the docs locally can double-check that all is as expected

FLomb (commented):
Thanks for this PR @brynpickering; it looks like a very good solution to issue #615, while it simultaneously enriches the documentation by providing an example of cost data loading from tabular files.

I couldn't check that the docs build properly (see specific comment), but the rest looks ready to approve. Before giving the green light, I would like to perform only a test with some of the models that led me to find the issue in the first place so that I can confirm this solution works in practice. I will do so later today.

brynpickering:
@FLomb thanks for checking. You can see the rendered documentation by clicking on the RTD build "details" in the checks:

image

FLomb (approved):
Ok, I have tested this on a practical case, and it works as expected; it also nicely warns me if I have a dimension mismatch when switching to broadcast to False.

irm-codebase (approved):
Gave a glance at how the code is arranged for this change. Looks fine to me!

In the future we might want to turn this type of broadcasting individually per parameter, but let's keep things simple for now.

brynpickering:
Yep, agreed that per-param broadcasting config is a nice-to-have @irm-codebase - it can be added to the longer-term roadmap.

PR #716, opened 2024-11-20
Introduce SPORES in v0.7.0 as a generalisable mode

FLomb:

* Introduces a `spores` scenario as part of the national-scale example model
* Introduces a custom math file in the national-scale example model, which is activated in the `spores` scenario to create a slack cost constraint
* It provides a dedicated notebook to generate SPORES iteratively based on the `spores` scenario overrides and a custom Python function for assigning penalties at each iteration.

Based on internal discussions, we have come to realise that there are many ongoing SPORES developments that may be hard to always reconduct to a monolithic run mode in the core code. Instead, it may be best to implement SPORES as a method that simply allows you to run a model in `plan` mode once, and then N number more times with some external function being applied between each run, with the user providing that function. This way, this mode could be used not only for SPORES but also for robustness analysis and other things. 

This PR aims to show how SPORES can be indeed already used in v0.7.0 based on default software functionality plus a simple for-loop and a custom Python function to assign penalties for decision variables at each iteration of the loop. To move from this to the generalised mode described above, we would only need to decide how to make some bits of this code (such as the number of iterations, and the custom Python function) a default request for inputs to the user.

The new `run-iteratively-with-a-function` mode (a better name must be found, could also be `spores-and-more` if the main application remains SPORES), would essentially require the following three key inputs: 1) how many iterations to run; 2) what function to apply at each iteration; and 3) what backend parameter to updated based on the function.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### User-facing changes
 
+|changed| Single data entries defined in YAML indexed parameters will not be automatically broadcast along indexed dimensions.
+To achieve the same functionality as in `<v0.7.dev4`, the user must set the new `init` configuration option `broadcast_param_data` to True (#615).
+
 |changed| Helper functions are now documented on their own page within the "Defining your own math" section of the documentation (#698).
 
 |new| `where(array, condition)` math helper function to apply a where array _inside_ an expression, to enable extending component dimensions on-the-fly, and applying filtering to different components within the expression (#604, #679).
--- a/docs/creating/config.md
+++ b/docs/creating/config.md
@@ -41,6 +41,10 @@ To test your model pipeline, `config.init.time_subset` is a good way to limit yo
     Various capabilities are available to adjust the temporal resolution of a model on-the-fly, both by resampling or using externally-provided clustering.
     See our [time adjustment page](../advanced/time.md) for more details.
 
+!!! info "See also"
+    The full set of available configuration options is documented in the [configuration schema][model-configuration-schema].
+    This provides you with a description of each configuration option and the default which will be used if you do not provide a value.
+
 ## Deep-dive into some key configuration options
 
 ### `config.build.backend`
--- a/docs/creating/parameters.md
+++ b/docs/creating/parameters.md
@@ -53,3 +53,29 @@ Which will add the new dimension `my_new_dim` to your model: `model.inputs.my_ne
 
 !!! warning
     The `parameter` section should not be used for large datasets (e.g., indexing over the time dimension) as it will have a high memory overhead on loading the data.
+
+## Broadcasting data along indexed dimensions
+
+If you want to set the same data for all index items, you can set the `init` [configuration option](config.md) `broadcast_param_data` to True and then use a single value in `data`:
+
+=== "Without broadcasting"
+
+    ```yaml
+    my_indexed_param:
+      data: [1, 1, 1, 1]
+      index: [my_index_val1, my_index_val2, my_index_val3, my_index_val4]
+      dims: my_new_dim
+    ```
+
+=== "With broadcasting"
+
+    ```yaml
+    my_indexed_param:
+      data: 1  # All index items will take on this value
+      index: [my_index_val1, my_index_val2, my_index_val3, my_index_val4]
+      dims: my_new_dim
+    ```
+
+!!! warning
+    The danger of broadcasting is that you maybe update `index` as a scenario override without realising that the data will be broadcast over this new index.
+    E.g., if you start with `!#yaml {data: 1, index: monetary, dims: costs}` and update it with `!#yaml {index: [monetary, emissions]}` then the `data` value of `1` will be set for both `monetary` and `emissions` index values.
--- a/docs/examples/national_scale/index.md
+++ b/docs/examples/national_scale/index.md
@@ -1,3 +1,10 @@
+---
+costs:
+    file: "src/calliope/example_models/national_scale/data_tables/costs.csv"
+    header: [0, 1]
+    index_col: 0
+---
+
 # National Scale Example Model
 
 This example consists of two possible power supply technologies,
@@ -24,6 +31,29 @@ It does not contain much data, but the scaffolding with which to construct and r
 ```
 
 ## Model definition
+
+### Referencing tabular data
+
+As of Calliope v0.7.0 it is possible to load tabular data completely separately from the YAML model definition.
+To do this we reference data tables under the `data_tables` key:
+
+```yaml
+--8<-- "src/calliope/example_models/national_scale/model.yaml:data-tables"
+```
+
+In the Calliope national scale example model, we load both timeseries and cost data from file.
+As an example, the data in the cost CSV file looks like this:
+
+{{ read_csv(page.meta.costs.file, header=page.meta.costs.header, index_col=page.meta.costs.index_col) }}
+
+You'll notice that in each row there is reference to a technology, and in each column to a cost parameter and a comment on the units being used.
+Therefore, we reference `techs` in our data table _rows_, and `parameters` and `comment` in our data table _columns_.
+The `comment` information is metadata that we don't need in our Calliope model object, so we _drop_ it on loading the table.
+Since all the data refers to the one cost class `monetary`, we don't add that information in the CSV file, but instead add it on as a dimension when loading the file.
+Where there is no data for that combination of technology and cost parameter, the value is Not-a-Number (NaN) and this combination will be ignored on loading the table.
+
+!!! info
+    You can read more about loading data from file in [our dedicated tutorial][loading-tabular-data].
 
 ### Indexed parameters
 
@@ -110,23 +140,6 @@ The costs are more numerous as well, and include monetary costs for all relevant
 * carrier conversion capacity
 * variable operational and maintenance costs
 
-### Interlude: inheriting from templates
-
-You will notice that the above technologies _inherit_ `cost_dim_setter`.
-Templates allow us to avoid excessive repetition in our model definition.
-In this case, `cost_dim_setter` defines the dimension and index of costs, allowing us to keep our definition of technology costs to only defining `data`.
-By defining `data`, the technologies override the `null` setting applied by `cost_dim_setter`.
-We also use it to set the `interest_rate` for all technologies, which will be used to annualise any investment costs each technology defines.
-
-Technologies and nodes can inherit from anything defined in `templates`.
-items in `templates` can also inherit from each other, so you can create inheritance chains.
-
-`cost_dim_setter` looks like this:
-
-```yaml
---8<-- "src/calliope/example_models/national_scale/model_config/techs.yaml:cost-dim-setter"
-```
-
 ### Storage technologies
 
 The second location allows a limited amount of battery storage to be deployed to better balance the system.
@@ -184,8 +197,16 @@ Transmission technologies look different to other technologies, as they link the
 `free_transmission` allows local power transmission from any of the csp facilities to the nearest location.
 As the name suggests, it applies no cost or efficiency losses to this transmission.
 
+### Interlude: inheriting from templates
+
 We can see that those technologies which rely on `free_transmission` inherit a lot of this information from elsewhere in the model definition.
 `free_transmission` is defined in `templates`, which makes it inheritable.
+[Templates](../../creating/templates.md) allow us to avoid excessive repetition in our model definition.
+
+Technologies and nodes can inherit from anything defined in `templates`.
+items in `templates` can also inherit from each other, so you can create inheritance chains.
+
+The `free_transmission` template looks like this:
 
 ```yaml
 --8<-- "src/calliope/example_models/national_scale/model_config/techs.yaml:free-transmission"
--- a/docs/examples/national_scale/spores_run.py
+++ b/docs/examples/national_scale/spores_run.py
@@ -1,0 +1,154 @@
+# ---
+# jupyter:
+#   jupytext:
+#     text_representation:
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.16.4
+#   kernelspec:
+#     display_name: Python 3 (ipykernel)
+#     language: python
+#     name: python3
+# ---
+
+# %% [markdown]
+# # Generating SPORES
+# An interactive example of how to generate near-optimal system designs (or SPORES) out of a Calliope v0.7.0 model. This example relies solely on default software functionality and a custom Python function to determine how to assign penalties (scores) to previously explored system design options.
+
+# %%
+# Importing the required packages
+import calliope
+import xarray as xr
+
+# %% [markdown]
+# ## Cost-optimal model run and extraction of SPORES-relevant outputs
+
+# %%
+# Loading model files and building the model
+model = calliope.Model(
+    "example_models/latest_national_scale/model.yaml", scenario="simplified_spores"
+)
+model.build()
+
+# Solving
+model.solve()
+
+# Extracting SPORES-relevant data
+least_feasible_cost = model.results.cost.loc[{"costs": "monetary"}].sum().sum()
+print("The minimum cost for a feasible system design is {}".format(least_feasible_cost.values))
+
+
+# %% [markdown]
+# ## SPORES model run
+# ### Definition of the penalty-assignment methods
+
+# %%
+def scoring_integer(results, backend):
+    # Filter for technologies of interest
+    spores_techs = backend.inputs["spores_tracker"].notnull()
+    # Look at capacity deployment in the previous iteration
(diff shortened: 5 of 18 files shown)


FLomb:
I see that building the docs fails because I am using matplotlib, which isn't part of the default Calliope environment. I will remove the plotting part, for now. We can later re-implement it via a supported library, or just leave it out.

brynpickering:
@FLomb I've updated the approach to minimise the input requirements from the user. The main change is that there is now no longer a spores-specific cost class. Instead, the objective is changed in the math. Updating this objective with user-defined math would allow a user to update how the spores scores are translated to penalties in the optimisation problem.

The SPORES mode stuff is currently untested although you can see that it works in the example notebook.

I hope that some of the hardcoding of params in `model.py` can go away with the implementation of #712. 

The SPORES mode stuff will also move out of `model.py` (initial work on that is also in #712). Hopefully, though, you can see that it is on the verge of being generalisable - there is an "iterative optimisation" controller that calls a model updater. My idea would be to scrub "spores" from this controller and just have it so that it can run with a supplied updater function, the SPORES mode one being one of them. This would allow for other iterative updaters (including custom SPORES algorithms) to be supplied by the user and managed by the controller.

brynpickering:
@irm-codebase & @FLomb if you could check this PR that would be great!

src/calliope/backend/pyomo_backend_model.py
```diff
@@ -201,6 +203,16 @@ def _objective_setter(
...
+    def set_objective(self, name: str) -> None:  # noqa: D102, override
+        for obj_name, obj in self.objectives.items():
+            if obj.item().active and obj_name != name:
+                self.log("objectives", obj_name, "Objective deactivated.", level="info")
+                obj.item().deactivate()
+            if obj_name == name:
+                obj.item().activate()
+                self.log("objectives", obj_name, "Objective activated.", level="info")
+        self.objective = name
+
```
irm-codebase, line 215: I'd consider directly querying the requested name and using sets to deactivate the rest to reduce condition checks and to 'fail fast'.

Less checks, but same strength.

```suggestion
    def set_objective(self, name: str) -> None:  # noqa: D102, override
        self.objectives[name].activate()
        self.objective = name
        self.log("objectives", name, "Objective activated.", level="info")
        for key in self.objectives.keys() - {name}:
            if self.objectives[key].active:
                self.objectives[key].deactivate()
                self.log("objectives", key, "Objective deactivated.", level="info")
```

src/calliope/model.py
```diff
@@ -266,6 +267,16 @@ def build(
...
+        elif mode == "spores":
+            backend_input = self._model_data.copy()
+            if "spores_score" not in backend_input:
+                backend_input["spores_score"] = xr.DataArray(0).assign_attrs(
+                    is_result=0
+                )
+            if "spores_baseline_cost" not in backend_input:
+                backend_input["spores_baseline_cost"] = xr.DataArray(
+                    np.inf
+                ).assign_attrs(is_result=0)
```
irm-codebase, line 279: I do not like the idea of hidden values being set here...
Shouldn't we just fail in this case instead?

irm-codebase: Perhaps this is something that the schema should take care of.

src/calliope/model.py
```diff
@@ -491,7 +504,7 @@ def _prepare_operate_mode_inputs(
 
         return sliced_inputs
 
-    def _solve_operate(self, **solver_config) -> xr.Dataset:
+    def _solve_operate(self, solver_config: config.Solve) -> xr.Dataset:
         """Solve in operate (i.e. dispatch) mode.
 
         Optimisation is undertaken iteratively for slices of the timeseries, with
```
irm-codebase, line 510: This is missing a description for `solver_config`.

src/calliope/model.py
```diff
@@ -506,7 +519,9 @@ def _solve_operate(self, **solver_config) -> xr.Dataset:
 
         LOGGER.info("Optimisation model | Running first time window.")
 
-        step_results = self.backend._solve(warmstart=False, **solver_config)
+        step_results = self.backend._solve(
+            warmstart=False, **solver_config.model_dump()
+        )
```
irm-codebase, line 524: Why use `model_dump`? Shouldn't we just pass `solver_config` directly?

src/calliope/model.py
```diff
@@ -571,3 +588,103 @@ def _recalculate_storage_initial(self, results: xr.Dataset) -> xr.DataArray:
...
+            xr.Dataset: Results dataset.
+        """
+        LOGGER.info("Optimisation model | Resetting SPORES parameters.")
+        self.backend.update_parameter(
+            "spores_score", self.inputs.get("spores_score", xr.DataArray(0))
+        )
+        self.backend.update_parameter(
+            "spores_baseline_cost",
+            self.inputs.get("spores_baseline_cost", xr.DataArray(np.inf)),
+        )
```
irm-codebase, line 608: Nice to have: 
A `for` might be more maintainable in the long run... Both do the same, after all.

src/calliope/model.py
irm-codebase, line 1: As a general comment: we might want to start moving 'modes' functions out of `model.py` and into a separate subdirectionry in the future to make this file easier to maintain.

src/calliope/model.py
```diff
@@ -571,3 +588,103 @@ def _recalculate_storage_initial(self, results: xr.Dataset) -> xr.DataArray:
...
+                warmstart=False, **solver_config.model_dump()
+            )
+            results_list.append(step_results)
+
+            if spores_config.save_per_spore_path is not None:
+                LOGGER.info(f"Optimisation model | Saving SPORE {spore} to file.")
+                step_results.assign_coords(spores=spore).to_netcdf(
+                    spores_config.save_per_spore_path / f"spore_{spore}.nc"
+                )
+
```
irm-codebase, line 644: This part is a bit hard to follow. Maybe consider adding a short comment to explain what `results_list` is doing here to someone unfamiliar with the SPORES approach...

src/calliope/model.py
```diff
@@ -571,3 +588,103 @@ def _recalculate_storage_initial(self, results: xr.Dataset) -> xr.DataArray:
...
+                spores_config.tracking_parameter, xr.DataArray(True)
+            ).notnull()
+            & self.inputs.definition_matrix
+        )
+
+        # Look at capacity deployment in the previous iteration
+        previous_cap = previous_results["flow_cap"].where(spores_techs)
+
+        # Make sure that penalties are applied only to non-negligible deployments of capacity
+        min_relevant_size = 0.1 * previous_cap.max(["nodes", "techs"])
```
irm-codebase, line 675: `0.1` is a 'magic' value. Shouldn't it be in the spores configuration?

irm-codebase (commented):
Looks good overall, although a bit hard to follow in places. Added a few suggestions.

src/calliope/model.py
brynpickering: yeah there's no reason why not tbh. It was using `model_dump` to avoid excessive changes in the PR that introduced the solver config

src/calliope/model.py
brynpickering: Yeah, I have that implemented on a different feature branch...

src/calliope/model.py
brynpickering: probably, yes.

src/calliope/config.py
```diff
@@ -215,6 +215,13 @@ class SolveSpores(ConfigBaseModel):
     tracking_parameter: str | None = None
     """If given, an input parameter name with which to filter technologies for consideration in SPORES scoring."""
 
+    score_iteration_threshold_relative: float = Field(default=0.1, ge=0)
+    """A factor to apply to flow capacities above which they will increment the SPORES score.
```
irm-codebase, line 219: Should the minimum be `gt` (greater than) rather than `ge` (greater or equal)?

src/calliope/model.py
irm-codebase: Makes sense. For later, then!

src/calliope/config.py
irm-codebase: Also, I find the name slightly confusing... I'd include the 'unit' explicitly and shorten it a bit.
Would `score_threshold_factor` work?

irm-codebase (approved):
Changes look good.

I am tentatively approving, with some suggestions on naming.

src/calliope/config.py
brynpickering: I think zero is a reasonable value to assign to it (i.e., there is no threshold). I'll rename!

brynpickering:
@irm-codebase I've updated the implementation slightly to get the SPORES maths showing in the docs (you can see it under "pre-defined math" in the PR build). It requires an update to our math docs to show all defined objectives by default. To differentiate between those objectives I'm now adding a suffix to the objective name to say whether or not it is activated. Question is, is this confusing for e.g. the `plan` mode math (that it has `(active)` after the objective name)?

irm-codebase (approved):
@brynpickering changes look fine, but I am not sure why the code difference is showing changes from `main`...

Approved, but let's be careful just in case to avoid overwriting things on  `main`

brynpickering:
> but I am not sure why the code difference is showing changes from main.

Which changes is it showing?

irm-codebase:
@brynpickering ignore, I was incorrect (just goes to show that you should do reviews while well-rested).
For some reason I confused the `sense` in the math with the pydantic PR, but that is not the case.

src/calliope/schemas/config_schema.py
```diff
@@ -134,45 +133,34 @@ class SolveSpores(CalliopeBaseModel):
...
-                raise ValueError("`save_per_spore_path` must be a directory.")
-        return self
+    tracking_parameter: str | None = None
+    """If given, an input parameter name with which to filter technologies for consideration in SPORES scoring."""
+
+    score_threhsold_factor: float = Field(default=0.1, ge=0)
+    """A factor to apply to flow capacities above which they will increment the SPORES score.
+    E.g., if the previous iteration flow capacity was `100` then, with a threshold value of 0.1,
+    only capacities above `10` in the current iteration will cause the SPORES score to increase for that technology at that node.
+    If, say, the current iteration's capacity is `8` then the SPORES score will not change for that technology (as if it had no )
```
FLomb, line 152: Something missing at the end of this sentence?

docs/examples/modes.py
```diff
@@ -0,0 +1,190 @@
...
+# Here, we compare installed capacities between the baseline run (== `plan` mode) and the SPORES.
+# Note how the baseline SPORE is the same as `plan` mode and then results deviate considerably.
+
+# %%
+fig_flows_plan = plot_capacity(model_plan.results)
+fig_flows_plan.update_layout(title="Plan mode capacities")
+
+# %%
+fig_flows_spores = plot_capacity(model_spores.results, facet_col="spores")
+fig_flows_spores.update_layout(title="SPORES mode capacities")
```
FLomb, line 190: Where is this shown in the published docs? I cannot find it.

src/calliope/math/spores.yaml
```diff
@@ -2,9 +2,37 @@ import:
   - plan.yaml
 
 constraints:
-  cost_sum_max:
+  total_system_cost_max:
+    description: >
+      Limit total system cost. Conceived for use in SPORES mode to apply a maximum
+      relaxation to the system cost compared to the least-cost feasible option.
+    foreach: [costs]
```
FLomb, line 9: Is it normal for this to iterate over all types of costs? In the docs, we say the default is 'monetary', but this doesn't seem to reflect it. Or is it intended that it is open to all costs here, but 'by default' users tend to then specify only one type of cost in their use of this?

FLomb (commented):
Most things look in order, apart from a couple of minor comments I have made just to double-check. 

However, I am not quite following how the suggested implementation relates to the idea of a generalisable/open version of the mode. It looks very specific to the actual (old) SPORES method, including hard-coded scoring logics in `model.py`. What am I missing?

src/calliope/math/spores.yaml
brynpickering: True, currently it would apply a constraint limiting all cost classes to `least_cost_val * (1 + spores_slack)`  (note now that the SPORES score _isn't_ a cost). I suppose this would matter if we were tracking other costs in the model (e.g. CO2 emissions). We could limit this to `monetary` in the math.

brynpickering:
> However, I am not quite following how the suggested implementation relates to the idea of a generalisable/open version of the mode. It looks very specific to the actual (old) SPORES method, including hard-coded scoring logics in model.py. What am I missing?

You're right that it hasn't yet been well generalised. I've mentioned what might need to happen to make it possible in [this comment](https://github.com/calliope-project/calliope/pull/716#issuecomment-2561335028).

What I could imagine is that we have "plugins" like our math helper function subclass functionality. That is, there would be a `CalliopeMode` abstract base class that can be subclassed to create custom modes (and from which `operate` and `spores` mode would be defined). It could have methods like `pre_build`, `post_build`, `pre_solve`, and `solve` which would allow a user to define what (if anything) should happen at each of these steps. And then they link to it in their config when defining their mode. However, it's beyond the scope of this PR right now. It will have to wait until post-0.7 unless we can get more resources to deliver it.

docs/examples/modes.py
FLomb: Ok, but this part was completely detached from the SPORES docs, making it unlikely to find when looking at SPORES. I've made a commit to facilitate the connection.

FLomb (commented):
The functionality looks good as a basic implementation of SPORES. However, it' still outdated and less powerful compared to what we implemented for the study ['What is redundant and what is not'](https://doi.org/10.1016/j.apenergy.2023.121002), and available on a [branch of mine of v0.6.8](https://github.com/FLomb/calliope/blob/0.6.8-turbo-spores-methods/calliope/backend/run.py). 

Given that the idea of having SPORES as a highly customisable mode has been postponed, I think it would make sense for v0.7.0 to at least implement the additional scoring methods as a default option, since many partners have provided feedback and empirical that they are sometimes (depending on the model specifics) critically needed.

src/calliope/model.py
```diff
@@ -567,3 +574,181 @@ def _recalculate_storage_initial(self, results: xr.Dataset) -> xr.DataArray:
...
+                .clip(min=1, max=1)
+                .fillna(0)
+                .where(spores_techs)
+            )
+            return new_score
+
+        def _score_relative_deployment() -> xr.DataArray:
+            """Relative deployment scoring algorithm."""
+            previous_cap = latest_results["flow_cap"].where(spores_techs)
+            relative_cap = previous_cap / self.inputs["flow_cap_max"].where(
```
FLomb, line 678: A problem that I've never addressed properly is what happens if no cap_max is defined (i.e., it is `inf` as per the default). Normally, I'd just make sure that every tech has a cap_max, even when this is not due to obvious reasons (e.g., land constraints). For instance, batteries are typically unconstrained in our models, and I would arbitrarily assign them a cap_max that makes sense based on the order of magnitude of other things. 

To automate this, should we say that where no cap_max exists, the maximum relative potential is capped at a value that has the same order of magnitude as the highest cap_max in the model?

src/calliope/model.py
```diff
@@ -567,3 +574,181 @@ def _recalculate_storage_initial(self, results: xr.Dataset) -> xr.DataArray:
...
+                previous_cap.fillna(0)
+                .where(previous_cap.isnull(), other=np.random.rand(*previous_cap.shape))
+                .where(spores_techs)
+            )
+
+            return new_score
+
+        def _score_evolving_average() -> xr.DataArray:
+            """Evolving average scoring algorithm."""
+            previous_cap = latest_results["flow_cap"]
```
FLomb, line 703: In all other methods, it is always: `previous_cap = latest_results["flow_cap"].where(spores_techs)`. In 'evolving_average', the `.where(spores_techs)` bit is gone. Is that on purpose?

FLomb (commented):
This looks very good. Thanks a lot, @brynpickering, for converting the scoring methods from my branch based on an older Calliope version.

I have left a couple of minor comments for things that either need a double-check or could be marginally improved.

src/calliope/model.py
brynpickering: It's just applied later on

FLomb: Ok, that's what I figured. I guess it saves some lines of code to do it only on the final product in this case. But could this also be potentially less efficient if you must do the averaging, subtraction and division operations for larger xarray objects? Imagine a case in which you have a humongous model, but only do spores for a couple of techs.

FLomb (commented):
Thanks for the additional commits @brynpickering. I think this looks good now and can be merged!

src/calliope/model.py
brynpickering: These are quite efficient operations as they're just standard arithmetic applied to known types, so shouldn't be an issue. We can revisit if it does seems slow.

Issue #240, opened 2019-05-22
Clean up organisation of tests and test models


sjpfenninger:
## Problem description

Test models are not very well organised at the moment.

This includes:

* file naming consistency
* use of simple models vs fully-blown models
* use of overrides/scenarios

Better organised test models could be used as simple example models to demonstrate key functionality and be integrated into the docs, too.

## Calliope version

0.6.4-dev

sjpfenninger:
We will be able to simplify the tests substantially with new YAML-based constraint formulation, by unit-testing the functionality rather than each constraint separately

irm-codebase:
This would be lovely, I feel like we have lots of repeated tests.
Since `pytest` can look within folders recursively, here are some proposals to make leaner:

Important stuff:
- Always mirror the folder/file structure of `calliope`, to make tests quick to find. (e.g., `util/schema.py` tests are in `util_test/schema_test.py`). Right now we combine many features into single test files, which clutters them.
- Prefer `xxxx_test` over `test_xxxx`. This makes test searching faster on the CLI and often helps when searching things in IDEs since the test and the code will often appear together if the naming is set right.
- **Always** add a single line docstring explaining the test imperatively (e.g., `"Build step kwarg overrides shall not modify the initial configuration."`).

PR #717, opened 2024-11-21
Add config obj replace schemas

### LINKED ISSUE #709: Add support for descriptive/passthrough data within the YAML schema
### What can be improved?

Calliope currently takes almost any value in the YAML and adds it to the underlying model along one or more of the indices. It would be nice to have an option to add relevant technology data that doesn't affect the Calliope optimization (a description or parameters needed for post-processing or a downstream model) in the YAML without it being added to the underlying problem. Something like an `extras` key in a tech/node block that is ignored by the schema and attribute dict code would allow users to put additional context directly into the configuration. The alternative of having the extra data in a readme or other metadata file would work, but having that context directly next to the model configuration would make it easier to review models and diagnose issues.

### Version

v0.7.0.dev4

irm-codebase:
Compliments PR #704 by extending `pydantic` validation to other sections of model `.yaml` files

New schemas for the following:
- Model Math
- Dimension data (techs, nodes)
- Data Tables
- Model definition

This PR does _not_:

- Replace `model_def_schema.yaml` because it has parameter defaults. This needs to be removed in a subsequent PR.
- Integrate the use of the new schemas in type annotations. Mostly because the parameters / dims PR will affect this significantly.
- Replace `preprocess/model_math.py` with the new schema. Once again, the parmeters / dims PR should take care of this.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


irm-codebase:
@brynpickering before continuing on this, we need to ensure we have a common vision on where / how stuff is checked, and the pros/cons of certain approaches. Particularly for places were our configuration setup conflicts with `pydantic`'s typing-based approach.

A key assumption, based on previous discussions, is that we are attempting to remove _all_ schema `.yaml` files in favour of a `pydantic` approach. This would lead to a final `CalliopeConfig` composed of several `pydantic` structures, each replacing one of our separate schemas (`data_table_schema.yaml`, `math_schema.yaml`...).

The final object would look like this, sort of:

```python
class CalliopeConfig(ConfigBaseModel):
    """Calliope configuration class."""

    model_config = {"title": "config"}
    init: Init = Init()
    build: Build = Build()
    solve: Solve = Solve()
    techs: dict[str, Tech]
    nodes: dict[str, Node]
    data_tables: dict[str, DataTable] | None = None
    parameters: Parameter | None = None  # this might go in Math later, here for now
    math: Math | None = None  # inserted by us using build.mode???
    overrides: None | dict[str, Self] = None
```

## The role of `pydantic` in checking user model files

From a pure SW validation / data validation perspective: should we strive to catch invalid cases with `pydantic` as much as possible? To what extent should `pydantic` be used to ensure that the data going into later stages is 'good'? What is the effect of going either way?

Some examples of trivial checks with `pydantic`: 
- Ensure all `techs` used in `nodes.techs` are defined in the default or in overrides.
- Run tests to ensure that all dimensions given in `dims:` sections (e.g., in `techs`, `data_tables`, etc) are present in a `dimensions` dictionary section or in overrides of said section.
- Ensuring that all parameter names used in `techs` or `nodes` sections are present in our definition or in user provided parameter definitions.
- Ensuring that parameters that go 'into' the optimisation have numeric values.
- etc

## Where an when to introduce our basic parameter math and user extended math

At the moment a good chunk of our parameter defaults / checks is in `model_def_schema.yaml`, which we cram into `model._data.attrs`. `pydantic` might change this

- Should these defaults be placed in `model.config` (and later `model.math`) by defining them in a `pydantic` model?
- How flexible should this be? Some values can be either parameters or variables depending on the mode (e.g, `flow_cap`).
- Hard-defining all these parameters within the `pydantic` model might prove to be too inflexible. Should we do so?

## How to support user-provided 'passthrough' settings without weakening validation

This stems from issue #709. Since `pydantic` is type based, our grouping of items in the dictionary matters quite a bit on how strong / weak our validation becomes. Our current flat structure has advantages when it comes to user simplicity, but at a possibly heavy cost of validation strength.

As an example, our current `techs` definition:
```yaml
supply_gas:
    name: "Natural gas import"
    color: "#C98AAD"
    carrier_out: gas
    source_use_max: .inf
    flow_cap_max: 2000
    lifetime: 25
    cost_flow_cap:
      data: 1
      index: monetary
      dims: costs
    cost_flow_in:
      data: 0.025 # 2.5p/kWh gas price #ppt
      index: monetary
      dims: costs
```

This would require a `Tech` model structured like this. The big issue here is that you need to allow extras at the same level of our hard-defined attributes! I worry this will get messy fast. A less flat structure, which separates math parameters, code-bound values (e.g., `carrier_out`) and user 'extras' would be stronger, but less user friendly if done poorly.

```python
class Tech(BaseModel):
    model_config = {
        "extra": "allow"
    }
    name: str
    color: str | None = None
    carrier_out: str
    flow_cap_max: PositiveReal  = float('inf')  # Assume we define PositiveReal some place else
    ...
```

irm-codebase:
We can also keep all of the elements in `CalliopeConfig` in separate places. It's only an illustration of our possible checks.

brynpickering:
Thanks @irm-codebase - this is what I had in mind w.r.t. a move to pydantic! One thing I think it will offer us is the ability to remove AttrDict entirely 👀... We still need to be able to load our flavour of YAML, but that could be something we do with less maintenance overhead.

RE techs, we have a few weird parameters, but keep in mind that none of them are really "hard" requirements in terms of _YAML_ definitions. You can define all that info in data tables and then have almost nothing defined in YAML. At that point pydantic's magic is useless to us anyway, hence why I wouldn't put too much focus on maximising the value of pydantic; we'll only ever apply it to a subset of the data when data tables are used. Data validation should wait a bit longer and be applied once YAML and data table info has been merged (i.e. on the xarray arrays). This means that _no_, parameter settings should not be in the pydantic model (as they shouldn't be in a YAML schema). Instead, they should be in a YAML definition (e.g. as given in #712) whose _structure_ is described by a pydantic model.

Still, we have to define some things in pydantic under tech defs since they are weird parameters structures that we let through (namely, `carrier_[in/out/export]`) to make the user journey easier. Otherwise, we say that a tech setting/option/parameter can be any single value _or_ an indexed parameter. These kinds of things can be part of the pydantic model and have some fancier cross-validation done if possible, at least in terms of the dimensions that have been defined?

RE the config options we include. We should consider how best to handle overrides and templates - should they be validated separately _or_ should the pydantic model only be generated based on the resolved dict (all overrides and template data applied)?

brynpickering:
> Understood. One important caveat of this is that we cannot easily offer some features that users might want (i.e., setting max/min ranges for a numeric parameter) without having to write this code ourselves. This is fine as long as it is understood.

That would be dealt with by the parameter config being introduced in #712 . We can have min/max or a more general way to  define validation rules that will act on the xarray dataset (using the format defined in https://github.com/calliope-project/calliope/blob/959fea7096ce47dfaf6886be1e1fa9445b92bc69/src/calliope/config/model_data_checks.yaml). This is what I mean about data validation that is beyond the scope of pydantic as it is no longer working on a dictionary of data, but arrays in an xarray dataset.

> This comes back to my point on our current flat structure being troublesome.

I don't think it's the flat structure that causes the problem so much as allowing for user-defined keys. One option would be to consider dynamic fields. For instance, you can [set type annotations for extra fields](https://docs.pydantic.dev/latest/concepts/models/#extra-fields) such that they have to at least follow a specific structure (as we define with our patternproperties currently). So the flat structure for a tech could be:

```
class TechFlat(BaseModel):
    model_config = {
        "extra": "allow"
    }
    __pydantic_extra__: dict[RegexStrWithExclusions, IndexedParam | bool | int | float | str] = Field(union_mode='left_to_right')
    base_tech: Literal[...]
    carrier_in: str | list[str] | None = None
    carrier_out: str | list[str] | None = None
    carrier_export: str | list[str] | None = None
    # + computed fields for to/from based on `base_tech`?
```

brynpickering:
To update on the idea of using `__pydantic_extra__`, you could do this:

```python

DATA_T = bool | int | float | str

class Param(BaseModel):
    """Uniform dictionairy for parameters."""

    data: DATA_T | list[bool | int | float | str]
    index: list[list[str]] = [[]]
    dims: list[str] = []

class Tech(BaseModel):
    model_config = {
         "extra": "allow"
     }
    __pydantic_extra__: dict[str, Param | DATA_T] = Field(union_mode='left_to_right')
    base_tech: Literal["supply", "demand"]  # etc.
    carrier_in: str | list[str] | None = None
    carrier_out: str | list[str] | None = None
    carrier_export: str | list[str] | None = None


class Techs(BaseModel):
    model_config = {
         "extra": "allow"
     }
    __pydantic_extra__: dict[str, Tech]

class Model(BaseModel):
    techs: Techs
```

Which would work to manage a dict like:

```python
{"techs": {"foo": {"bar": {"data": 1}, "baz": "foobar", "base_tech": "supply", "carrier_in": ["elec", "heat"]}}}
```

And would give you appropriate dot notation afterwards (e.g. `techs.foo.bar.carrier_in`)

brynpickering:
@irm-codebase I've edited my previous comment based on playing around with the `update` method to handle the fact that an update to the data might add/remove a Param class from a technology.

irm-codebase:
@brynpickering thanks for the update!
It is similar to the design I had in mind, but I was not aware of `__pydantic_extra__`...

Is forcing subdicts to update to their pydantic model necessary, though? I thought that was done automatically (or it seemed to be during my tests).

brynpickering:
> Is forcing subdicts to update to their pydantic model necessary, though? I thought that was done automatically (or it seemed to be during my tests).

I found that it just generated subdicts in my updated pydantic model, not nested pydantic models. If there's a way to do it without the extra lines I've added, then that's great!

irm-codebase:
> > Is forcing subdicts to update to their pydantic model necessary, though? I thought that was done automatically (or it seemed to be during my tests).
> 
> I found that it just generated subdicts in my updated pydantic model, not nested pydantic models. If there's a way to do it without the extra lines I've added, then that's great!

@brynpickering seems like this can be achieved with `revalidate_instances`? See [here](https://github.com/pydantic/pydantic/issues/7387#issuecomment-1753296144)

brynpickering:
@irm-codebase I've brought in the changes from main now that #704 has been merged in. It should now be ready to work from!

irm-codebase:
@brynpickering question regarding some of the validation done in `model_def_schema.yaml` that is proving very difficult to implement in a way that is not messy.

Currently, we execute some extra checks for cases where indexed parameters start with `cost_`. E.g.:

```yaml
      boiler:
        cost_flow_cap:
          data: 43.1 # different boiler costs
          index: [[monetary, heat]]
          dims: [costs, carriers]
```

Namely, if the parameter starts with `cost_`, we check that has 'costs' as one of its dimensions.

```yaml
        patternProperties: &pattern_def
          '^(cost_[\w]+)$':
            $ref: "#/$defs/TechCostNullNumber"

  TechCostNullNumber:
    oneOf:
      - type: ["null", number]
        x-type: float
      - type: object
        description: Indexed tech-level cost.
        additionalProperties: false
        required: ["data", "index", "dims"]
        properties:
          data:
            description: >-
              parameter value(s).
              If data is one value, will be applied to all dimension members.
              If a list, must be same length as the index array.
            oneOf:
              - type: "null"
              - type: number
              - type: array
                minItems: 1
                items:
                  oneOf:
                    - type: "null"
                    - type: number
          index:
            $ref: "#/$defs/IndexedParamIndex"
          dims:
            $ref: "#/$defs/TechCostDims"
```

I am wondering if it's even necessary, as it does not really protect users that use names that do not start with `cost_`.
I'd rather skip this check for simplicity, and to avoid messy code. Opinions?

brynpickering:
Yeah that's fine. As you say, it was always a bit questionable. I'd prefer to embed it in the param definition (e.g. setting required dimensions for that param).

irm-codebase:
@brynpickering I've run into a rather unfortunate issue

Our use of `from` / `to` for `transmission` technologies breaks `pydantic` because `from` is a protected term.
See [here](https://docs.pydantic.dev/latest/concepts/models/#basic-model-usage) (scroll down to the Warning section).

We have two options, none of them ideal.
- Change the names to something else. Will impact users of v0.7
- Use an [alias](https://docs.pydantic.dev/latest/concepts/alias/) (e.g., `from_`). Could lead to user confusion and bugs on our side since we must always dump the model using `model_dump(by_alias=True)`, and we must always use `from_` when accessing it in the code.

brynpickering:
Interesting, although I'm surprised that `from` is protected (I've never seen it used in type annotations!). 

Any ideas on names we could change it to? `start` & `end`? @jnnr any thoughts?

If we don't have a viable alternative we realistically will have to go with aliasing.

irm-codebase:
It's protected due to imports: `from calliope import AttrDict`.

I like `start` / `end`, and I think that changing the terms is the best solution in the long run.
@sjpfenninger might also want to pitch in on this one.

irm-codebase:
Some more options @jnnr @brynpickering @sjpfenninger :
- `into` / `out_of` | `in` / `out` | `input` / `output`: potentially confusing with `carrier_in` / `out`
- `start` / `end`: fine, although a bit weird (could imply time rather than location)
- `link_to` / `link_from`: a bit long, but clear and links to old usage previous to v7. I think this is the better option

brynpickering:
`here` / `there` ? ;)

jnnr:
Regarding the naming of link connections nodes:
- I have seen source, target sometimes.
- I like link_from, link_to.
- My overall favorite has not been mentioned: from_node, to_node.

irm-codebase:
Even worse, this means that stuff defined in data tables kind of always 'jumps' over the validation because it goes straight to the `xarray` dataset. For example, it probably allows invalid `node.latitude` values to go through (even in the current approach) because the limits check is in the schema...

irm-codebase:
@brynpickering ready for review!

Although this is a lengthy PR, the line difference number that `github` is giving is kind of deceptive. Most of it was due to a `.csv` file change.

I've updated the first comment of the PR to reflect the changes and omissions of this PR. In general: 
- `model_def_schema.yaml` remains because it contains parameter defaults.
- `preprocess/model_math.py` likely needs to be merged into the math schema. For now, I kept them separate because the params / dims update will impact this and this PR is long enough already.

brynpickering (changes requested):
Pydanic models look good! Our test suite really does not cleaning. Need to work out a better way to have minimal models to load and check. This is a task for #240 though.

sjpfenninger (dismissed):
Scanned this and it looks good to me, nothing to add on top of what Bryn noted!

src/calliope/preprocess/model_definition.py
```diff
@@ -73,40 +73,39 @@ def _load_scenario_overrides(
...
 
-    # Second pass of applying override dict after applying scenarios,
-    # so that scenario-based overrides are overridden by override_dict!
-    if override_dict is not None:
+    # Second pass of applying override_dict after applying scenarios,
+    # so it can overwrite scenarios (it has the highest priority)
+    if overrides:
+        # Remove scenarios/overrides since they were already applied.
+        overrides.pop("scenarios", None)
+        overrides.pop("overrides", None)
```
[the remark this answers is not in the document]
irm-codebase: It was actually added because the new schema catched that in certain cases these were 'left overs'.
So it's more of a case of the schema working pretty well :)

src/calliope/schemas/data_table_schema.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    Mapping between dimension names in the data table being loaded to equivalent Calliope dimension names.
+    For instance, the "time" column in the data table would need to be mapped to "timesteps": `{"time": "timesteps"}`.
+    """
+
+    @model_validator(mode="after")
+    def check_row_and_columns(self) -> Self:
+        """Ensure users specify a valid data table shape."""
+        drop = set(listify(self.drop))
+        rows = set(listify(self.rows)) - drop
```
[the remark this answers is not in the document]
irm-codebase: This is a pretty sweet idea, actually... I think it'll affect a bunch of sections, though.
I'll create feature request to track this one.

irm-codebase:
I've added most of your suggestions / fixes @brynpickering.
I left a comment for those that I did not implement.

src/calliope/schemas/general.py
```diff
@@ -35,11 +35,16 @@ def _validate_unique_list(v: list) -> list:
     AfterValidator(_validate_unique_list),
     Field(json_schema_extra={"uniqueItems": True}),
 ]
+"""A list with no repeated values."""
 # ==
 NonEmptyList = Annotated[list[T], Len(min_length=1)]
+"""A list with at least one value in it."""
 NonEmptyUniqueList = Annotated[UniqueList[T], Len(min_length=1)]
+"""A list with at least one value in it and no repeated values."""
 AttrStr = Annotated[str, Field(pattern=r"^[^_^\d][\w]*$")]
```
brynpickering, line 44: Something that cropped up recently on another project is that `\w` is annoyingly not strictly snakecase - upper case letter are allowed... I'm kinda OK with it since it makes the regex simpler but if we want strict snakecase it would probably need to be`"^[^_^\d][a-z0-9_]*$"`. Actually, that isn't so complex :P

brynpickering (approved):
One minor comment on the snakecase regex, otherwise good to go from my perspective!

src/calliope/schemas/general.py
irm-codebase: I thought the same, but this breaks our `bigM` parameter.

I did not switch because I think this not worth enforcing. Some people like CamelCase, and that is fine.

brynpickering: Maybe then update the docstring to not say snakecase specifically?

PR #724, opened 2024-12-04
Release v0.7.0.dev5

brynpickering:
Release PR

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.7.0.dev5 (Unreleased)
+## 0.7.0.dev5 (2024-12-04)
 
 ### User-facing changes
 


irm-codebase (dismissed):
Seems good.

All the math / params / dims / etc upgrades will go in the next, correct?

irm-codebase:
Wait, @brynpickering I do not see the version number being updated. I assume this is not necessary?

brynpickering:
> All the math / params / dims / etc upgrades will go in the next, correct?

correct

brynpickering:
The version bumping _should_ have been needed, but it only works if after each dev release we update the version to `v0.7.0.dev6-alpha` or something like that, we've instead been updating it to `v0.7.0.dev6` directly.

PR #736, opened 2025-01-27
Allow node params on links

jnnr:
Fixes the following issue:

I am not allowed to set a constraint on `flow_out` of transmission links. I will get this error:

```
calliope.exceptions.ModelError: (data_sources, boundary_link_flow_out) | Cannot define transmission technology data over the `nodes` dimension.
``` 

I am trying to provide the model with a data table like this:

```
nodes,techs,timesteps,fix_link_flow_out
ESP,ESP_to_FRA,2017-01-01,120
```

And a constraint like this:

```
constraints:
  boundary_constraint_link_flow_out_group_ub:
    foreach: [nodes,techs,carriers,timesteps]
    where: "base_tech=transmission AND carriers=electricity AND fix_link_flow_out"
    equations:
      - expression: flow_out == fix_link_flow_out
```

I convert the error message into a warning.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/preprocess/data_sources.py
+++ b/src/calliope/preprocess/data_sources.py
@@ -154,7 +154,7 @@ class DataSource:
                     techs_incl_inheritance[tech].get("base_tech", None)
                     == "transmission"
                 ):
-                    self._raise_error(
+                    self._raise_warning(
                         "Cannot define transmission technology data over the `nodes` dimension"
                     )
                 else:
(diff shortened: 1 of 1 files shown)


src/calliope/preprocess/data_sources.py
```diff
@@ -154,7 +155,7 @@ def node_dict(self, techs_incl_inheritance: AttrDict) -> AttrDict:
                     techs_incl_inheritance[tech].get("base_tech", None)
                     == "transmission"
                 ):
-                    self._raise_error(
+                    self._raise_warning(
                         "Cannot define transmission technology data over the `nodes` dimension"
```
jnnr, line 159: Shall we change that to something like this?
`"Trying to define transmission technology data over the `nodes` dimension. Be sure that you know what you are doing."`

brynpickering:
Thanks for the contribution @jnnr. My only concern is that this was in there for a reason. If you define transmission data at a node in YAML, what happens with this change? It would require some tests to check for what happens.

src/calliope/preprocess/data_sources.py
irm-codebase: We probably want a more descriptive message, stating the possible error that this might create.

jnnr:
Thanks for your comments, @brynpickering and @irm-codebase. Apparently, we need to first better understand how link data is treated in preprocessing to understand the scenarios of failure we are opening up. I will take some time with to study this with @irm-codebase at the whiteboard and come back to this soon.

brynpickering:
Any progress on identifying the link @jnnr ?

PR #744, opened 2025-03-14
Update deps incl cbc

brynpickering:
Now that [CBC is available cross-platform](https://anaconda.org/conda-forge/coin-or-cbc), we should move to depending on it by default.

I've also removed upper bound pins for all our deps where possible as they are causing issues when using calliope as a dependency... You win @irm-codebase ;)

* Added coin-or-cbc to dev deps
* Updated install instructions to include coin-or-cbc in the recommended install snippet, and updated text comparing GLPK and CBC.
* Updated all pins to lower bounds only, except xarray (due to a bug which hasn't yet been resolved)

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


brynpickering:
As in other PRs, link check is failing on <https://www.gnu.org/software/glpk/>. A manual check confirms that this link is fine so I don't really know how we overcome that. Maybe just point to [GLPK recipe on conda forge](https://anaconda.org/conda-forge/glpk)?

irm-codebase (dismissed):
Looks great!
Any PR that removes more lines than it adds brings joy to my heart.

README.md
```diff
@@ -37,7 +37,7 @@ Calliope comes with several built-in analysis and visualisation tools. Having so
 
 ## Quick start
 
-Calliope can run on Windows, macOS and Linux. Installing it is quickest with the `mamba` package manager by running a single command: `mamba create -c conda-forge -n calliope calliope`.
+Calliope can run on Windows, macOS and Linux. Installing it is quickest with the `mamba` package manager by running a single command: `mamba create -n calliope -c conda-forge/label/calliope_dev -c conda-forge calliope`.
```
irm-codebase, line 41: maybe something like `mamba create -n calliope conda-forge/label/calliope_dev::calliope` would work here?

docs/installation.md
```diff
@@ -48,34 +49,28 @@ mamba activate calliope
...
 ## Updating an existing installation
 
-If following the recommended installation method above, the following command, assuming the mamba environment is active, will update Calliope to the newest version
-
-```bash
-mamba update -c conda-forge calliope
-```
+We recommend that if you want to update an existing installation, you re-create your environment from scratch.
+This will mitigate any possible unwanted updates to additional dependencies that you may have installed into your environment since you first created it.
+Therefore, run the same command found in the [recommended installation instructions](#recommended-installation-method) and say yes when prompted about whether you want to overwrite the existing environment.
```
irm-codebase, line 54: Is this section really necessary? It's not really related to core calliope stuff, just `conda` basics.

irm-codebase (approved):
CI passes, and I do not see anything wrong in the changes.
Just a couple of minor comments.

docs/installation.md
brynpickering: Fair enough.

PR #746, opened 2025-03-24
Release v0.7.0.dev6

brynpickering:
I'd like to get this iteration through so we have access to the minor fixes required to make our new lower bound pinning work, and to access the SPORES algorithm in some projects.

Param & dim config will need to wait for the next iteration.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/.github/workflows/commit-ci.yml
+++ b/.github/workflows/commit-ci.yml
@@ -46,7 +46,7 @@ jobs:
 
     - name: run tests without coverage and without time intensive tests
       if: github.ref != 'refs/heads/main'
-      run: pytest --no-cov -m "not time_intensive"
+      run: pytest --no-cov -m 'not time_intensive' -n 'not needs_gurobi_license'
 
     - name: run tests with coverage
       if: github.ref == 'refs/heads/main'
--- a/.github/workflows/commit-ci.yml
+++ b/.github/workflows/commit-ci.yml
@@ -44,9 +44,9 @@ jobs:
     - name: Install jupyter kernel
       run: python -m ipykernel install --user --name calliope
 
-    - name: run tests without coverage
+    - name: run tests without coverage and without time intensive tests
       if: github.ref != 'refs/heads/main'
-      run: pytest
+      run: pytest --no-cov -m "not time_intensive"
 
     - name: run tests with coverage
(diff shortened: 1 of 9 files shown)


irm-codebase:
@brynpickering the release bits look fine.

I'm not so sure about the unexpected change to the backend...
Having a property that creates a dictionary every time and then having to pass the name to it does not seem like an improvement. Using a constant seems more sensible.

If there is another reason for the change, please let me know

irm-codebase (changes requested):
@brynpickering the release bits look fine.

I'm not so sure about the unexpected change to the backend...
Having a property that creates a dictionary every time and then having to pass the name to it does not seem like an improvement in terms of efficiency or readability. 

Using a constant seems more sensible to me...

brynpickering:
Yeah, the change came about because otherwise calliope triggers a gurobipy
import on init. We don't have gurobipy as a core dependency so the pip
build fails as it was. It would need an extra conditional in the class to
check if gurovipy can be imported...

irm-codebase (dismissed):
I'm approving, but the fact that python is immediately trying to import this as a base dependency seems to hint at a general issue with our setup.

We probably should look into how projects with multiple optional dependencies (e.g., `pandera`) are handling these cases.

brynpickering:
I wouldn't say it's done especially cleanly over there either. There is [a boolean check for the dependency of e.g. geopandas](https://github.com/unionai-oss/pandera/blob/main/pandera/engines/geopandas_engine.py), but there's no special effort within the code base to isolate it. We also have a conditional import of gurobipy. I've now stopped the GurobiBackendModel class from being imported until it is actively requested by the user, so one shouldn't see this error until the point gurobipy is really _needed_.

brynpickering:
Minor coverage reduction to have the gurobi objective sense dictionary be conditional (it's hard to drop a dependency on the fly in the test suite) and test-pypi is failing because it only ever allows one upload per version (stupid way to run a sandpit!), a successful one of which has already made its way there. I've checked that and it installs fine on my device.

brynpickering:
Triple-checked the pip build and install on my local machine (incl. running `calliope --help`) just to be certain we're all good.

Issue #510, opened 2023-10-27
New object-oriented backend API


brynpickering:
## Problem description

In #429, I introduced a backend api available at `Calliope.Model.backend`. E.g., to view parameters: `model.backend.parameters`; to update a parameter: `model.backend.update_parameter("param_name", ...)`, etc.

If we wrapped the backend component arrays in a class, we could move some of this functionality more nested objects with associated methods, e.g., `model.backend.parameters` has each parameter as an attribute. these paramter classes can then be updated `model.backend.parameters.param_name.update(...)` or simply viewed: `model.backend.parameters.param_name.data`. The benefit of this is that you get tab autocompletion for free, and it is probably the more "pythonic" way to implement the API. 

However, the current approach works. So maybe this is low priority to implement.

## Calliope version

v0.7.0.dev

irm-codebase:
@brynpickering I believe we can close this issue?
The combination of the new backend and schema approaches fits your description above pretty closely.

brynpickering:
No, this would be about having the individual math components be their own objects so you could do `model.backend.constraints.add`, `model.backend.variables.<my-variable>.fix` etc. Should be solved by moving to linopy as our main backend as we then get those things out of the box. But it isn't resolved yet!

PR #754, opened 2025-05-02
Fix save per SPORE & add SPORES docs

### LINKED ISSUE #750: Misleading / not up-to-date SPORES documentation
### What happened?

In the documentation (https://calliope.readthedocs.io/en/latest/examples/modes/), the SPORES mode is called from the national example model via the `spores` scenario. Such a scenario - as defined in the `scenarios.yaml` file - lacks the definition of a `spores_score` or any other cost metric to work with SPORES. In fact, in the latest implementation of SPORES, this appears to be something that the code deals with internally - creating a `spores_score` automatically without the user needing to define one. Instead, based on the SPORES documentation (https://calliope.readthedocs.io/en/latest/advanced/mode/), this is something that should be made explicit by the user.

I believe that the documentation is not up-to-date with the latest version of the code. This discrepancy makes it confusing for a reader of the documentation to understand how to use SPORES, likely leading to conflicts/errors in the model config.

- [ ] macOS
- [ ] Windows
- [x] Linux

### Version

v0.7.0dev6

### Relevant log output

```Shell

```

### LINKED ISSUE #751: Unable to export SPORES results to NetCDF
### What happened?

Despite some issues (see #750 ), the example scenario for SPORES in v0.7.0dev6 produces multiple solutions, accounted for in the model results via an extra `spores` dimension. 

However, trying to export the multi-dimensional array to NetCDF fails due to the following error:

```
ValueError: unable to infer dtype on variable 'spores'; object array contains mixed native types: str, int
```

It looks like there's a mix of integer and strings somewhere in some data related to the `spores` dimension that prevents saving. I'm not sure what this could be due to.

- [ ] macOS
- [ ] Windows
- [x] Linux

### Version

v0.7.0dev6

### Relevant log output

```Shell

```

### LINKED ISSUE #752: save_per_spore_path does not work and is not documented
### What happened?

The option to save each SPORE generated in `spores` mode separately via the `save_per_spore_path` parameter does not seem to work. Moreover, the documentation does not provide any guidance regarding how to use this feature.

Looking at the core code and at the tests for this function, I tried using it in two different ways, neither of which worked for me:
1. Setting just a path, such as: `save_per_spore_path: "results/spores"`
2. Setting a path and a string for the file name to be formatted based on the spore index, such as: `save_per_spore_path: "results/spores/spore_{}.nc"`

In both cases, the error is as follows:
`TypeError: Invalid value for attr 'references': set(). For serialization to netCDF files, its value must be of one of the following types: str, Number, ndarray, number, list, tuple`

- [ ] macOS
- [ ] Windows
- [x] Linux

### Version

v0.7.0dev6

### Relevant log output

```Shell

```

brynpickering:
Fixes #751 
Fixes #750 
Fixes #752 

@FLomb please check whether the updated docs make sense.

* Saving per SPORE uses the Calliope I/O netcdf saver to ensure appropriate serialisation
* All SPORES index items are set to strings ("baseline", "1", etc.) to avoid type issues with saving to NetCDF.
* Documentation on SPORES update to reflect new implementation. This new implementation reduces the expected user input to initialise a SPORES model, as described in #716. 
* Side-fix: when updating the config model, the return of `model_validate` is used, to preserve any type casting from that validation process (e.g. string -> Path).

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/model.py
+++ b/src/calliope/model.py
@@ -604,13 +604,14 @@ class Model:
         if spores_config.save_per_spore_path is not None:
             spores_config.save_per_spore_path.mkdir(parents=True, exist_ok=True)
             LOGGER.info("Optimisation model | Saving SPORE baseline to file.")
-            baseline_results.assign_coords(spores="baseline").to_netcdf(
-                spores_config.save_per_spore_path / "baseline.nc"
+            io.save_netcdf(
+                baseline_results.assign_coords(spores="baseline"),
+                spores_config.save_per_spore_path / "baseline.nc",
             )
 
         # We store the results from each iteration in the `results_list` to later concatenate into a single dataset.
         results_list: list[xr.Dataset] = [baseline_results]
-        spore_range = range(1, spores_config.number + 1)
+        spore_range = [str(i) for i in range(1, spores_config.number + 1)]
         LOGGER.info(
             f"Optimisation model | Running SPORES with `{spores_config.scoring_algorithm}` scoring algorithm."
         )
@@ -623,8 +624,9 @@ class Model:
 
             if spores_config.save_per_spore_path is not None:
                 LOGGER.info(f"Optimisation model | Saving SPORE {spore} to file.")
-                iteration_results.assign_coords(spores=spore).to_netcdf(
-                    spores_config.save_per_spore_path / f"spore_{spore}.nc"
+                io.save_netcdf(
+                    iteration_results.assign_coords(spores=spore),
+                    spores_config.save_per_spore_path / f"spore_{spore}.nc",
                 )
 
         spores_dim = pd.Index(["baseline", *spore_range], name="spores")
--- a/src/calliope/schemas/general.py
+++ b/src/calliope/schemas/general.py
@@ -80,8 +80,8 @@ class CalliopeBaseModel(BaseModel):
                 )
                 new_dict[key] = val
         updated = super().model_copy(update=new_dict, deep=deep)
-        updated.model_validate(updated)
-        return updated
+        updated_with_type_casting = updated.model_validate(updated)
+        return updated_with_type_casting
 
     @classmethod
     def model_no_ref_schema(cls) -> AttrDict:
--- a/tests/test_core_model.py
+++ b/tests/test_core_model.py
@@ -263,6 +263,21 @@ class TestSporesMode:
         spores_model, _ = spores_model_and_log
         assert spores_model.backend.config.mode == "spores"
 
+    def test_io_save(self, spores_model_and_log, tmp_path):
+        """Verify that we have run in spores mode"""
+        spores_model, _ = spores_model_and_log
+        filepath = tmp_path / "test_io_save.nc"
+        spores_model.to_netcdf(filepath)
+        assert filepath.exists()
+
+    def test_io_load(self, spores_model_and_log, tmp_path):
+        """Verify that we have run in spores mode"""
+        spores_model, _ = spores_model_and_log
+        filepath = tmp_path / "test_io_load.nc"
+        spores_model.to_netcdf(filepath)
+        new_model = calliope.read_netcdf(filepath)
+        xr.testing.assert_allclose(spores_model._model_data, new_model._model_data)
+
     def test_spores_mode_success(self, spores_model_and_log_algorithms):
         """Solving in spores mode should lead to an optimal solution."""
         spores_model, _ = spores_model_and_log_algorithms
@@ -272,7 +287,7 @@ class TestSporesMode:
         """Solving in spores mode should lead to 3 sets of results."""
         spores_model, _ = spores_model_and_log_algorithms
         assert not set(spores_model.results.spores.values).symmetric_difference(
-            ["baseline", 1, 2]
+            ["baseline", "1", "2"]
         )
 
     def test_spores_scores(self, spores_model_and_log_algorithms):
@@ -316,7 +331,7 @@ class TestSporesMode:
             spores_model._model_data.spores_score_cumulative.diff("spores") > 0
         )
         numpy.testing.assert_array_equal(
-            has_cap.shift(spores=1).sel(spores=[1, 2]), spores_score_increased
+            has_cap.shift(spores=1).sel(spores=["1", "2"]), spores_score_increased
         )
 
     def test_use_tech_tracking(self, spores_model_with_tracker):
@@ -335,7 +350,7 @@ class TestSporesMode:
         out_dir = model.config.solve.spores.save_per_spore_path
         assert len(list(out_dir.glob("*.nc"))) == 3
 
-    @pytest.mark.parametrize("spore", ["baseline", 1, 2])
+    @pytest.mark.parametrize("spore", ["baseline", "1", "2"])
     def test_save_per_spore(self, spores_model_save_per_spore_and_log, spore):
         """We expect SPORES results to be saved to file once per iteration."""
         model, _ = spores_model_save_per_spore_and_log
@@ -344,7 +359,7 @@ class TestSporesMode:
         result = xr.open_dataset((out_dir / filename).with_suffix(".nc"))
         assert result.spores.item() == spore
 
-    @pytest.mark.parametrize("spore", ["baseline", 1, 2])
+    @pytest.mark.parametrize("spore", ["baseline", "1", "2"])
     def test_save_per_spore_log(self, spores_model_save_per_spore_and_log, spore):
         """We expect SPORES results saving to be logged."""
         _, log = spores_model_save_per_spore_and_log
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 0.7.0.dev7
+
+### User-facing changes
+
+|new| documentation on SPORES-specific configuration options (#750, #752).
+
+|fixed| SPORES mode models not being appropriately serialised on saving to NetCDF (#751, #752).
+
 ## 0.7.0.dev6 (2025-03-24)
 
 ### User-facing changes
--- a/docs/advanced/mode.md
+++ b/docs/advanced/mode.md
@@ -75,48 +75,90 @@ As an example, if you wanted to generate 10 SPORES, all of which are within 10% 
 ```yaml
 config.build.mode: spores
 # The number of SPORES to generate:
-config.solve.spores.number: 10:
+config.solve.spores.number: 10
 # The fraction above the cost-optimal cost to set the maximum cost during SPORES:
 parameters.spores_slack: 0.1
 ```
 
-You will now also need a `spores_score` cost class in your model.
-The `spores_score` is the cost class against which the model optimises in the generation of SPORES.
-The recommended approach is to initialise it in your model definition for all technologies and locations that you want to limit within the scope of finding alternatives.
-Technologies at locations with higher scores will be penalised in the objective function, so are less likely to be chosen.
-In the [national scale example model](../examples/national_scale/index.md), this would look something like:
-
-```yaml
-templates:
-  add_spores_score:
-    template: cost_dim_setter
-    cost_flow_cap:
-      data: [null, null]
-      index: ["monetary", "spores_score"]
-      dims: costs
-    cost_interest_rate:
-      data: [0.1, 1]
-      index: ["monetary", "spores_score"]
-      dims: costs
-
-techs:
-  ccgt:
-    template: add_spores_score
-    cost_flow_cap.data: [750, 0]
-  csp:
-    template: add_spores_score
-    cost_flow_cap.data: [1000, 0]
-  battery:
-    template: add_spores_score
-    cost_flow_cap.data: [null, 0]
-  region1_to_region2:
-    template: add_spores_score
-    cost_flow_cap.data: [10000, 0]
-```
-
-!!! note
-    We ourselves use and recommend using `spores_score` to define the cost class that you will now optimise against.
-    However, it is user-defined, allowing you to choose any terminology that best fits your use case.
-
 To get a glimpse of how the results generated via SPORES compare to simple cost optimisation, check out our documentation
 on [comparing run modes](../examples/modes.py).
+
+### Limiting the search for alternatives to specific technologies
+
+By default, all technologies at all nodes will be subject to SPORES scoring.
+This means that they are all equally likely to be removed from the system when generating alternative system designs.
+You may want to instead focus on specific technologies when searching for alternatives.
+To do so, set a _tracking parameter_.
+This will be a parameter you set to `True` for all technologies that you want SPORES scores to be applied to.
+Any technology _not_ being tracked will not be penalised in the optimisation for having a non-zero capacity.
+
+!!! example
+
+    ```yaml
+    config.solve.spores.tracking_parameter: my_tracking_parameter
+    parameters:
+      my_tracking_parameter: # defines which techs are going to be subject to SPORES scoring
+        data: [true, true, true]
+        index: [ccgt, csp, battery]
+        dims: techs
+    ```
+
+    Or, at the technology level:
+
+    ```yaml
+    config.solve.spores.tracking_parameter: my_tracking_parameter
+    techs:
+      ccgt:
+        my_tracking_parameter: true
+      csp:
+        my_tracking_parameter: true
+      battery:
+        my_tracking_parameter: true
+      pv:
+        my_tracking_parameter: false # will not be tracked
+    ```
+
+### Saving results per SPORE
+
+Optimisation runs can be resource intensive.
+You should not lose all your results if the optimisation fails part way through your SPORES runs.
+To mitigate this, you can _save results per SPORE run_ to capture results up to any point of failure.
+
+!!! example
+
+    ```yaml
+    config.build.mode: spores
+    # The number of SPORES to generate:
+    config.solve.spores:
+      number: 10
+      save_per_spore_path: results/spores
+    # The fraction above the cost-optimal cost to set the maximum cost during SPORES:
+    parameters.spores_slack: 0.1
+    ```
+
+    Here, 11 result files will be present in the results, 1 for the baseline (non-SPORES) run + 10 SPORES runs.
+    Each SPORE run will be labelled `spore_<run number>.nc`, e.g. `results/spores/spore_10.nc`.
+
+!!! note
+    - The `save_per_spore_path` directory path will be considered as relative to the current working directory unless given as an absolute path.
+    - Even if you choose to save results per SPORE, the results will also be stored in memory.
+      Following the successful completion of all SPORES runs, results will be available in the `model.results` dataset.
+
+### Skipping the baseline optimisation run
+
+The `baseline` run does not set any SPORES scores or a system cost slack.
+It is equivalent to running the model in [plan](#plan-mode) mode.
+If you already a model with baseline results, you need not re-run that optimisation.
+Instead, you can _skip the baseline run_.
+
+!!! example
+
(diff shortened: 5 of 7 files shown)


src/calliope/model.py
```diff
@@ -604,13 +604,14 @@ def _solve_spores(self, solver_config: config_schema.Solve) -> xr.Dataset:
         if spores_config.save_per_spore_path is not None:
             spores_config.save_per_spore_path.mkdir(parents=True, exist_ok=True)
             LOGGER.info("Optimisation model | Saving SPORE baseline to file.")
-            baseline_results.assign_coords(spores="baseline").to_netcdf(
-                spores_config.save_per_spore_path / "baseline.nc"
+            io.save_netcdf(
+                baseline_results.assign_coords(spores="baseline"),
+                spores_config.save_per_spore_path / "baseline.nc",
             )
```
irm-codebase, line 610: Shouldn't this be in the `if` above, ensuring we only save baseline results once?

src/calliope/schemas/general.py
```diff
@@ -80,8 +80,8 @@ def update(self, update_dict: dict, deep: bool = False) -> Self:
                 )
                 new_dict[key] = val
         updated = super().model_copy(update=new_dict, deep=deep)
-        updated.model_validate(updated)
-        return updated
+        updated_with_type_casting = updated.model_validate(updated)
+        return updated_with_type_casting
```
irm-codebase, line 85: Great catch. I think other parts of the code have this problem (namely math and configuration)...

tests/test_core_model.py
```diff
@@ -263,6 +263,21 @@ def test_backend_build_mode(self, spores_model_and_log):
...
 
+    def test_io_save(self, spores_model_and_log, tmp_path):
+        """Verify that we have run in spores mode"""
+        spores_model, _ = spores_model_and_log
+        filepath = tmp_path / "test_io_save.nc"
+        spores_model.to_netcdf(filepath)
+        assert filepath.exists()
+
+    def test_io_load(self, spores_model_and_log, tmp_path):
+        """Verify that we have run in spores mode"""
```
irm-codebase, line 274: Same docstring as above, which is a bit confusing.

irm-codebase (changes requested):
Requesting a few changes / clarifications.
Only minor stuff though.

tests/test_core_model.py
```diff
@@ -263,6 +263,21 @@ def test_backend_build_mode(self, spores_model_and_log):
...
+        spores_model, _ = spores_model_and_log
+        filepath = tmp_path / "test_io_load.nc"
+        spores_model.to_netcdf(filepath)
+        new_model = calliope.read_netcdf(filepath)
+        xr.testing.assert_allclose(spores_model._model_data, new_model._model_data)
+
```
FLomb, line 280: I am not sure how this test succeeds. I have tested the functionality on a model of mine and, while saving works fine, loading the results leads to the error below:

```
--------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[20], line 1
----> 1 spore_0 = calliope.read_netcdf('results[/spores/baseline.nc](http://localhost:8888/spores/baseline.nc)')

File [~/GitHub/calliope/src/calliope/model.py:34](http://localhost:8888/home/fl/GitHub/calliope/src/calliope/model.py#line=33), in read_netcdf(path)
     32 """Return a Model object reconstructed from model data in a NetCDF file."""
     33 model_data = io.read_netcdf(path)
---> 34 return Model(model_definition=model_data)

File [~/GitHub/calliope/src/calliope/model.py:101](http://localhost:8888/home/fl/GitHub/calliope/src/calliope/model.py#line=100), in Model.__init__(self, model_definition, scenario, override_dict, data_table_dfs, **kwargs)
     96     self._init_from_model_definition(
     97         model_definition, scenario, override_dict, data_table_dfs, **kwargs
     98     )
    100 self._model_data.attrs["timestamp_model_creation"] = timestamp_model_creation
--> 101 version_def = self._model_data.attrs["calliope_version_defined"]
    102 version_init = self._model_data.attrs["calliope_version_initialised"]
    103 if version_def is not None and not version_init.startswith(version_def):

KeyError: 'calliope_version_defined'
```

FLomb (changes requested):
Thanks a lot, @brynpickering for taking care of the opened issues. The changes look good, overall. However, when testing them on a model of mine, I incur an error when reading the NetCDF files back, which is related to the 'calliope_version_defined' (see specific comment in my review)

I do not incur the same error when loading back model results generated in `plan` mode, so it must be something specific to the `spores` mode, although the error message doesn't seem to be tightly related to it.

src/calliope/model.py
irm-codebase: @FLomb I meant to say that if the baseline is not run (`skip_baseline_run is True`) we should not save baseline results. Otherwise, `baseline.nc` will be created even if the baseline run was skipped. I think this risks either overwriting baseline results or generating double files in a case were they were skipped.

I am assuming that `skip_baseline_run` is only True in cases where you are 're-loading' a model from a file, though. If there is another case then this might be ok.

tests/test_core_model.py
brynpickering: That's because you're trying to load back in one of the individually saved results (using save_per_spore_path). We don't store any of the calliope metadata with these files, just the results. The reason for that is to keep the file size small. If we saved inputs with every result then we'd have a lot of duplicated data stored in memory... 

I could add the metadata in so that you can use `m = calliope.read_netcdf(...)`, you would just have to be OK with the idea of not having any inputs available in those files (i.e. nothing in `m.inputs`.)

FLomb: I see; I assumed this was an option because it was possible in the earlier implementation of the save-per-spore functionality in v0.6.x. So, essentially, I'd need to read the NetCDF explicitly via xarray, right? 

I would have no problem with that but, for casual users, I think it would be odd to find that this is (possibly) the only instance in the software in which they are required to be familiar with xarray's native functionality instead of being able to use Calliope's file reading functions. I'd definitely go for the solution you propose, even though inputs wouldn't be there.

brynpickering: OK, I've updated it to being loadable with `calliope.read_netcdf(...)` but without any contents in `model.inputs`.

src/calliope/model.py
```diff
@@ -597,37 +599,31 @@ def _solve_spores(self, solver_config: config_schema.Solve) -> xr.Dataset:
...
             iteration_results = self.backend._solve(solver_config, warmstart=False)
             results_list.append(iteration_results)
 
-            if spores_config.save_per_spore_path is not None:
-                LOGGER.info(f"Optimisation model | Saving SPORE {spore} to file.")
-                io.save_netcdf(
-                    iteration_results.assign_coords(spores=spore),
-                    spores_config.save_per_spore_path / f"spore_{spore}.nc",
-                )
+            self._spores_save_model(baseline_results, spores_config, spore)
```
irm-codebase, line 626: Nitpick: `baseline_results` is a misleading name here. consider renaming to `results` instead?

src/calliope/model.py
```diff
@@ -597,37 +599,31 @@ def _solve_spores(self, solver_config: config_schema.Solve) -> xr.Dataset:
...
-                baseline_results.assign_coords(spores="baseline"),
-                spores_config.save_per_spore_path / "baseline.nc",
-            )
-
+        # Update the slack-cost backend parameter based on the calculated minimum feasible system design cost
+        constraining_cost = baseline_results.cost.groupby("costs").sum(..., min_count=1)
+        self.backend.update_parameter("spores_baseline_cost", constraining_cost)
+        self.backend.set_objective("min_spores")
         # We store the results from each iteration in the `results_list` to later concatenate into a single dataset.
         results_list: list[xr.Dataset] = [baseline_results]
```
irm-codebase, line 613: I fear that this list might result in HUGE memory use.
Should we consider just saving the SPORES in a location by default, and forgetting the data afterwards?

Is this list really necessary if we make `save_per_spore_path` mandatory?

irm-codebase (changes requested):
Great fixes!

I am adding a comment regarding memory concerns. I am perfectly willing to approve if it is justified, though!

src/calliope/model.py
irm-codebase: Just to clarify: I am referring to `results_list` here. This will get quite big if people are running 100+ spores.

brynpickering: You would never be able to run so many spores in one go, I don't think. At least, we rarely do. You usually exhaust an avenue after 5-10 SPORES, then have to update your scoring mechanism / focus technologies. Is that reasonable to assume @FLomb?

irm-codebase: Looking at the spores functions, it really does seem like this list is needed... although I am not sure if that is truly the case. I'll mark this as resolved. We can implement these improvements later, if need be.

src/calliope/model.py
irm-codebase: That's the model attributes being annoying again.
Attempt to change `combine_attrs` from `"no_conflicts"` to `"override"`, and see if this still happens.

FLomb: It looks like override solves the problem, but at the same time, it brings back the error when saving to NetCDF from which we started from

src/calliope/model.py
```diff
@@ -597,37 +599,31 @@ def _solve_spores(self, solver_config: config_schema.Solve) -> xr.Dataset:
...
-            LOGGER.info("Optimisation model | Saving SPORE baseline to file.")
-            io.save_netcdf(
-                baseline_results.assign_coords(spores="baseline"),
-                spores_config.save_per_spore_path / "baseline.nc",
-            )
-
+        # Update the slack-cost backend parameter based on the calculated minimum feasible system design cost
+        constraining_cost = baseline_results.cost.groupby("costs").sum(..., min_count=1)
+        self.backend.update_parameter("spores_baseline_cost", constraining_cost)
+        self.backend.set_objective("min_spores")
```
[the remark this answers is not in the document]
brynpickering: You would just update the objective math expression, so it should be relatively simple to do.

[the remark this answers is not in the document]
brynpickering: E.g.,

```yaml

objectives:
  min_spores:
    description: >-
      SPORES score minimisation objective, activated for all iterations after the baseline optimisation.
      The SPORES scores are applied to flow capacity for each technology at each node.
    equations:
      - expression: >-
          sum(flow_cap * spores_score, over=[nodes, techs, carriers]) +
          <MY-CUSTOM-MATH>
          + $unmet_demand
```

The main aim was to move the bulk of the work into the math expressions so they can be updated by the user as they like, rather than us having to predict how they might like to update the objective and providing custom parameters for that.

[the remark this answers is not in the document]
FLomb: Ok, that's a good solution. But the SPORES math rests in the core code; can it be updated via overrides/custom math?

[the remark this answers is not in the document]
brynpickering: Yeah, all the core math can be overridden with user-defined math. That's why we render it in YAML in the docs (e.g. [for SPORES](https://calliope.readthedocs.io/en/latest/math/spores/)) so you know what you have available to override.

src/calliope/math/spores.yaml
```diff
@@ -6,10 +6,10 @@ constraints:
     description: >
       Limit total system _monetary_ cost in SPORES mode.
       Applies a maximum relaxation to the system cost compared to the least-cost feasible option.
-    foreach: [costs]
-    where: "[monetary] in costs"
     equations:
-      - expression: sum(cost, over=[nodes, techs]) <= spores_baseline_cost * (1 + spores_slack)
```
FLomb, line 12: This looks like a good fix to account for baseline runs in which there is more than one cost class, e.g., co2 alongside monetary costs. The same cost classes would be kept throughout the SPORES runs, with the same objective cost weight, which means one of the (say, co2), can be also switched off from the baseline onwards.

What if, however, we want to run the typical SPORES case in which a given system feature, say `ccgt` capacity, is intensified (i.e., maximised or minimised) while generating SPORES around it? This was also achieved via multi-objective formulations. However, the `objective_cost_weights` would need to change between the baseline and the SPORES run, instead of being kept the same throughout (in previous implementations, we had 'regular' objective cost weights, and 'spores-specific' ones, which would override the originals only from SPORE 1 on).

brynpickering: You could provide this as user-defined math to override the base SPORES math:

```yaml
objectives:
  min_spores:
    equations:
      - expression: >-
          sum(flow_cap * spores_score, over=[nodes, techs, carriers]) +
          sum(intensifier, over=[nodes, techs]) +
          + $unmet_demand
```

And then, for your techs of interest, you would set the `intensifier` parameter to whatever you want:

```yaml
techs:
  pv:
    intensifier: 5
  wind:
    intensifier: 10
```

src/calliope/model.py
FLomb: @irm-codebase it is true that, with MGA, you tend to generate 100+ things in the end, but that's the beauty of SPORES. To generate 100 things, we can parallelise in batches of 5-10 SPORES each. That's what Bryn was hinting at.

src/calliope/math/spores.yaml
FLomb: Ok, I've tested the new formulation and I get the following error:

```
---------------------------------------------------------------------------
ModelError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 nuts2_model_spores.build()
      2 nuts2_model_spores.solve()

File [~/GitHub/calliope/src/calliope/model.py:281](http://localhost:8888/home/fl/GitHub/calliope/src/calliope/model.py#line=280), in Model.build(self, force, add_math_dict, **kwargs)
    276 model_math = preprocess.CalliopeMath(full_math_list, self.def_path)
    278 self.backend = backend.get_model_backend(
    279     self.config.build, backend_input, model_math
    280 )
--> 281 self.backend.add_optimisation_components()
    283 self.applied_math = model_math
    285 self._model_data.attrs["timestamp_build_complete"] = log_time(
    286     LOGGER,
    287     self._timings,
    288     "build_complete",
    289     comment="Model: backend build complete",
    290 )

File [~/GitHub/calliope/src/calliope/backend/backend_model.py:259](http://localhost:8888/home/fl/GitHub/calliope/src/calliope/backend/backend_model.py#line=258), in BackendModelGenerator.add_optimisation_components(self)
    257 self._add_all_inputs_as_parameters()
    258 if self.config.pre_validate_math_strings:
--> 259     self._validate_math_string_parsing()
    260 for components in typing.get_args(ORDERED_COMPONENTS_T):
    261     component = components.removesuffix("s")

File [~/GitHub/calliope/src/calliope/backend/backend_model.py:246](http://localhost:8888/home/fl/GitHub/calliope/src/calliope/backend/backend_model.py#line=245), in BackendModelGenerator._validate_math_string_parsing(self)
    243             validation_errors[f"{component_group}:{name}"] = parsed._errors
    245 if validation_errors:
--> 246     exceptions.print_warnings_and_raise_errors(
    247         during="math string parsing (marker indicates where parsing stopped, but may not point to the root cause of the issue)",
    248         errors=validation_errors,
    249     )
    251 LOGGER.info("Optimisation Model | Validated math strings.")

File [~/GitHub/calliope/src/calliope/exceptions.py:108](http://localhost:8888/home/fl/GitHub/calliope/src/calliope/exceptions.py#line=107), in print_warnings_and_raise_errors(warnings, errors, during, bullet)
    105     warn(f"Possible issues found during {during}:\n" + _indenter(warnings))
    107 if errors:
--> 108     raise ModelError(f"Errors during {during}:\n" + _indenter(errors))

ModelError: Errors during math string parsing (marker indicates where parsing stopped, but may not point to the root cause of the issue):
 * constraints:total_system_cost_max:
    * equations[0].expression (line 1, char 74): sum(sum(cost, over=[nodes, techs]) * objective_cost_weights, over=costs) ) <= sum(spores_baseline_cost * objective_cost_weights, over=costs) * (1 + spores_slack)
                                                                                                                          ^
```

FLomb (changes requested):
There seems to be a parsing error in the new mathematical formulation of the SPORES objective; not sure why it's not captured by the tests

src/calliope/math/spores.yaml
brynpickering: Fixed, sorry about that!

brynpickering:
> Unfortunately, still not enough to make the math correct. Gurobi doesn't like the form the constraint takes.

This is odd. I can't reproduce it. Can you share your model with me somehow @FLomb ?

src/calliope/backend/gurobi_backend_model.py
```diff
@@ -285,6 +288,8 @@ def _solve(
         ][0].lower()
         results.attrs["termination_condition"] = str(termination)
 
+        results.attrs["objective_function_value"] = objective_function_value
+
         return results
```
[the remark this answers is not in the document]
irm-codebase: This would also avoid failed cases we see in the tests at the moment. Putting model data in `.attrs` just leads to many possible conflicts.

irm-codebase (commented):
@brynpickering
I think the changes make sense. I am giving some suggestions that may help with the failing tests.

src/calliope/backend/gurobi_backend_model.py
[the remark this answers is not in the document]
brynpickering: Now done. Although this has required me to fast-track #761 and include it in this PR

src/calliope/backend/backend_model.py
```diff
@@ -805,30 +805,86 @@ def get_variable_bounds(self, name: str) -> xr.Dataset:
             xr.Dataset: Contains the arrays for upper ("ub", a.k.a. "max") and lower ("lb", a.k.a. "min") variable bounds.
         """
 
-    @abstractmethod
     def get_global_expression(
-        self, name: str, as_backend_objs: bool = True, eval_body: bool = True
+        self, name: str, as_backend_objs: bool = True, eval_body: bool = False
     ) -> xr.DataArray:
         """Extract global expression array from backend dataset.
```
irm-codebase, line 812: Nitpick (not necessary for this PR, but a possible future simplification).
This degree of combinations between `as_backend_objs`/ `eval_body` and the amount of `if` statements in `_get_expression` kind of indicates that it could be split into two functions: one for getting strings (should never fail) and one for values (fails if unoptimised).

Just food for thought.

irm-codebase (dismissed):
Really happy with the changes!

I've added a couple of nitpicks that might simplify our code/tests. These are for things that predate this PR, though, so I do not see them as requirements.

Consider the nitpicks as a way to get codecov to pass.

src/calliope/backend/backend_model.py
```diff
@@ -805,30 +805,86 @@ def get_variable_bounds(self, name: str) -> xr.Dataset:
...
 
         Args:
-            name (str): Name of global expression
+            name (str): Name of global expression.
```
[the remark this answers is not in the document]
brynpickering: Agreed. I'm hoping https://github.com/calliope-project/calliope/issues/510 would be the solution to this.

FLomb:
Hi,

Thanks for the latest improvements. They look fine on paper (in the code), but they still fail to produce diversity after 2 SPORES in the 'default' case in which there is only a simple SPORES objective for the model I'm testing with (which I have shared with you guys separately). The same doesn't occur for my 'external' version of the method. 

I've tried diving deeper into what happens to the scores in each iteration, but that seems to be difficult insofar as `spores_score` is now only a backend parameter. As such, I can inspect it after running a series of SPORES and see where it _landed_, but I can't see how it evolved across iterations (unless I spend some time adding printing statements to the core code).

Issue #753, opened 2025-05-02
Config model creation in tests failing with pydantic v2.11.0 - v2.11.3 (inclusive)


brynpickering:
### What happened?

We use `pydantic.create_model` in our test suite, setting a base class _and_ updating the config. Technically, this shouldn't have been possible until v2.11.4 (thanks to https://github.com/pydantic/pydantic/pull/11714). The loophole of providing `model_config` as a field was closed earlier in v2.11 meaning some of our tests won't complete successfully using v2.11.0 - v2.11.3. We'll need to wait for 2.11.4 to be available on conda forge (feedstock PR: https://github.com/conda-forge/pydantic-feedstock/pull/151) before we can allow users to install calliope with pydantic v2.11.

### Which operating systems have you used?

- [x] macOS
- [ ] Windows
- [ ] Linux

### Version

v0.7.0.dev6

### Relevant log output

```Shell
cls = <class 'pydantic._internal._config.ConfigWrapper'>, bases = (<class 'calliope.schemas.general.CalliopeBaseModel'>,)
namespace = {'__annotations__': {'foo': <class 'str'>, 'foobar': <class 'int'>, 'model_config': {'title': 'TITLE'}}, '__module__': 'tests.test_schemas.test_general', 'foo': 'bar', 'foobar': 1}
kwargs = {}

    @classmethod
    def for_model(cls, bases: tuple[type[Any], ...], namespace: dict[str, Any], kwargs: dict[str, Any]) -> Self:
        """Build a new `ConfigWrapper` instance for a `BaseModel`.
    
        The config wrapper built based on (in descending order of priority):
        - options from `kwargs`
        - options from the `namespace`
        - options from the base classes (`bases`)
    
        Args:
            bases: A tuple of base classes.
            namespace: The namespace of the class being created.
            kwargs: The kwargs passed to the class being created.
    
        Returns:
            A `ConfigWrapper` instance for `BaseModel`.
        """
        config_new = ConfigDict()
        for base in bases:
            config = getattr(base, 'model_config', None)
            if config:
                config_new.update(config.copy())
    
        config_class_from_namespace = namespace.get('Config')
        config_dict_from_namespace = namespace.get('model_config')
    
        raw_annotations = namespace.get('__annotations__', {})
        if raw_annotations.get('model_config') and config_dict_from_namespace is None:
>           raise PydanticUserError(
                '`model_config` cannot be used as a model field name. Use `model_config` for model configuration.',
[...]

PR #755, opened 2025-05-02
Pin pydantic

brynpickering:
Intermediate fix for #753 

Once v2.11.4 is released we'll want to set that as the lower bound. It's never ideal to have only one version of a dependency available to use, but their release cycle is quite frequent so I expect other minor/major versions to be available quite soon.

*
*
*

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.7.0.dev7
+
+### Internal changes
+
+|changed| Pinned upper bound of pydantic to < 2.11 while we await v2.11.4 to become available on conda forge (#753).
+
 ## 0.7.0.dev6 (2025-03-24)
 
 ### User-facing changes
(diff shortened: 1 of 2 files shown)


brynpickering:
This is really just a hotfix to enable the CI tests to pass in other PRs (e.g. #754) while we wait for the feedstock to be updated.

irm-codebase:
Got it! Next time, I'd suggest pinning in that PR and instead opening an issue saying that we should un-pin upon pydantic's next release.
Approved!

irm-codebase (approved):

irm-codebase:
Do open the issue, though!

brynpickering:
I mean, the issue is already there. This is in a separate PR so that it can be applied to _all_ branches irrespective of how long #754 takes to be merged in.

Issue #740, opened 2025-02-12
Calliope.model() from existing model issue, Windows OS


ddahawkins-TUDelft:
### Description

Hello. Whether an issue with the code itself, or just the documentation, I encountered issue with creating a Calliope model with calliope.model(...) from a previous model.

I was attempting to run a model in 'plan' mode and then use the capacity outputs from that model in a new model in 'operate' mode. Although it states that this is possible, it is not immediately clear from the documentation how to do this. Apologies, I do not have this code any longer to upload but to explain...

I created a plan model object: model_plan = calliope.model('model.yaml', ...) 
Then built and solve: model_plan.build()  >>  model_plan.solve()

I then tried to create an operate model object, taking the model_plan as an input argument for the 'model_definition' parameter i.e.: model_operate = calliope.model(model_plan,...)

This failed.

I expected this to work as the dev5 documentation (https://calliope.readthedocs.io/en/v0.7.0.dev5/reference/api/model/?h=calliope.model#calliope.Model) for calliope.model() suggests that 'model_definition' can be a string, path, dictionary, or xarray dataset. With model_plan, a calliope model object, being an xarray dataset, I expected it to be a valid argument. Instead I received 'encoding' errors which may or may not be unique to the Windows OS. I did notice in the calliope backend a few times where UTF-8 was hardcoded for os.reads but not os.writes. Perhaps it may be better to use OS default or at least make this encoding consisting for writing and reading.

What i have later been informed is that instead of importing the model_plan object, I could have saved this file as a .netcdf and run the calliope.read_netcdf('model_plan_path') as the input parameter for 'model_definition.' If this is the case, it is not clear from your documentation and in any case, i feel it is good practice that one should be able to import the object and not have to write>read in order to access an object for re-runs.

Apologies if any of the above if unclear, happy to discuss!

Kind regards,
Dano

### Related links

-https://calliope.readthedocs.io/en/v0.7.0.dev5/reference/api/model/?h=calliope.model#calliope.Model
-
...

### Version

v.0.7.0.dev5

### Proposed change

The simplest short-term solution might be an example in the documentation of chaining models i.e. running 'operate' from the output of a 'build.'

Further, a clear explanation within the calliope.model() documentation to explain the precise format of xarray permissible for the model_definition parameter.

irm-codebase:
@ddahawkins-TUDelft 
Ahhh, this makes sense.

The calliope `Model` class can only take in files or xarray objects as inputs, not other models.

[...]

Issue #749, opened 2025-04-25
`lb` / `up` are printed incorrectly


irm-codebase:
### What happened?

When printing the values of `flow_cap`, `inf` is displayed as a `0.0`, and lower bounds seem to be always printed as `0` regardless of the actual value.

![Image](https://github.com/user-attachments/assets/3c67acda-7e30-4b0a-84c5-34ca0ecfe8f6)

![Image](https://github.com/user-attachments/assets/85be372d-9467-4dc3-9a6f-2443de76641d)

@FLomb and I have confirmed that the actual value in the model math is correct.
The cause is this setup in the variables of `plan.yaml`:

![Image](https://github.com/user-attachments/assets/3ac390bf-87fb-4ceb-9b11-45dd8b470c86)

## Suggestion

- Find out why `inf` is displayed as `0.0`.
- Maybe we should consider isolating the per-unit `MILP` formulation to its own mode, to avoid this confusion in the `LP` version of `plan` mode?

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Relevant log output

```Shell

```

FLomb:
Thanks for opening the issue following up on our chat, @irm-codebase. I have nothing to add, but I'm available if further clarifications are needed.

PR #756, opened 2025-05-02
Load math data during init

### LINKED ISSUE #739: Use `pydantic` to validate math files
### What can be improved?

Two requirements:

- Math validation should use `pydantic` instead of YAML schemas.
- User math should be defined as a dictionary instead of a list, allowing users to combine different files using key names.

See #726 for more information

### Version

v0.7

irm-codebase:
Fixes #739, #763, #740

Based on the design given in [this comment](https://github.com/calliope-project/calliope/issues/642#issuecomment-2465389387) by @brynpickering, except that _all_ math files are loaded during init, not just external ones, to enable more consistency.

This is an attempt to make calliope's math handling more flexible _AND_ able to support knowing parameters / dimension indexes during init.

* All math is now loaded during init, specified in a new section (`config.init.math`).
    * `config.init.math.base` specifies the 'foundation' of the model (`plan.yaml` in most cases). It is _unmodifiable_ after being set (changing it essentially means a new model / data anyway).
    * `config.init.math.pre_defined` contains a list of standard files to load (`plan`, `operate`, `spores`).
    * `config.init.math.extra` can be used to specify new math, which can be mix and matched later on top of the base math.
* All loaded math is saved in the model definition object (which also contains config, data tables, etc) in `model._def.math` for easy serialisation and re-loading. No more broken model path references!
* Math modes and extras can be loaded during `build` in any combination by just referencing names.
    * `config.build.mode` works the same, and is used to trigger code / functions within calliope code (e.g., spores, operation mode reloads, etc).
    * `config.build.extra_math` can have additional math references. These only affect the formulation solved by the backend.
    * The order is: **base math** -> **mode** -> **extra math**

**Example 1: plan math with operate and extra**

```yaml
config:
  init:
    math:
      base: "plan"  # This is the default. Any other name in pre_defined or extra is possible!
      pre_defined: ["plan", "operate", "spores"]  # The default. Stuff most people will need.
      extra:  # User files!
        test_formulation1: "relative/path/to/file.yaml"
        test_formulation2: "/absolute/path/also/works.yaml"

  # Run base math + operate + test_formulation1
  build:
    ignore_base_math: False  # The default. You can still turn off the base math in fringe cases.
    mode: "operate"   # for math that affects calliope code operation
    extra_math: ["test_formulation1"]  # only affects backend optimisation
```

**Example 1: pathways math with different formulations**

```yaml
config:
  init:
    math:
      base: "pathways"  # You can also use any name in 'extra'
      pre_defined: []  # Do not load stuff you won't use!
      extra:  # 
        pathways: "math/pathways_perfect_foresight.yaml"
        salvage_values: "math/salvage_values.yaml"
        wacc: "math/wacc.yaml"
        vintaging: "math/powerplant_vintage_efficiency.yaml"

  # Run pathways using salvage values and vintaging
  # Order: pathways -> salvage_values -> vintaging
  build:
    ignore_base_math: False  # Still false! The base is now pathways
    mode: "base"
    extra_math: ["salvage_values", "vintaging"]
```

Other stuff:
- Math defaults set in `math_schema` are respected by ensuring we use the type-casted object produced when validating files. Similar to fixes introdued in  #754.
- `CalliopeMath` has been completely removed in favour of a `pydantic` approach.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/docs/creating/config.md
+++ b/docs/creating/config.md
@@ -77,13 +77,19 @@ It has a very high cost associated with its use, so it will only appear when abs
 
 ### `config.build.mode`
 
-In the `build` section we have `mode`.
+The `build.mode` option specifies pre-defined mathematical formulations that require additional processing.
 A model can run in `base`, `operate`, or `spores` mode.
-In `base` mode, capacities are determined by the model, whereas in `operate` mode, capacities are fixed and the system is operated with a receding horizon control algorithm.
-In `spores` mode, the model is first run in `base` mode, then run `N` number of times to find alternative system configurations with similar monetary cost, but maximally different choice of technology capacity and location (node).
+
+* In `base` mode, the default option, Calliope will not execute any additional processing beyond building the mathematical problem defined in `config.init.base_math`.
+* In `operate` mode, capacities are fixed and the system is operated with a receding horizon control algorithm.
+* In `spores` mode, the model is first run in `base` mode, then run `N` number of times to find alternative system configurations with similar monetary cost, but maximally different choice of technology capacity and location (node).
 
 In most cases, you will want to use the `base` mode.
-In fact, you can use a set of results from using `base` model to initialise both the `operate` (`config.build.operate.use_cap_results`) and `spores` modes.
+In fact, you can use a set of results from models run in `base` mode to initialise both the `operate` (via`config.build.operate.use_cap_results`) and `spores` modes.
+
+!!! warning
+
+    Both `operate` and `spores` modes are designed to work with our pre-defined math and may stop working if it is overridden with [user-defined math](../user_defined_math/customise.md#re-defining-calliopes-pre-defined-base-math).
 
 ### `config.solve.solver`
 
--- a/docs/creating/config.md
+++ b/docs/creating/config.md
@@ -10,7 +10,7 @@ config:
     name: 'My energy model'
     time_subset: ['2005-01-01', '2005-01-05']
   build:
-    mode: plan
+    mode: base
   solve:
     solver: cbc
 ```
@@ -78,12 +78,12 @@ It has a very high cost associated with its use, so it will only appear when abs
 ### `config.build.mode`
 
 In the `build` section we have `mode`.
-A model can run in `plan`, `operate`, or `spores` mode.
-In `plan` mode, capacities are determined by the model, whereas in `operate` mode, capacities are fixed and the system is operated with a receding horizon control algorithm.
-In `spores` mode, the model is first run in `plan` mode, then run `N` number of times to find alternative system configurations with similar monetary cost, but maximally different choice of technology capacity and location (node).
+A model can run in `base`, `operate`, or `spores` mode.
+In `base` mode, capacities are determined by the model, whereas in `operate` mode, capacities are fixed and the system is operated with a receding horizon control algorithm.
+In `spores` mode, the model is first run in `base` mode, then run `N` number of times to find alternative system configurations with similar monetary cost, but maximally different choice of technology capacity and location (node).
 
-In most cases, you will want to use the `plan` mode.
-In fact, you can use a set of results from using `plan` model to initialise both the `operate` (`config.build.operate.use_cap_results`) and `spores` modes.
+In most cases, you will want to use the `base` mode.
+In fact, you can use a set of results from using `base` model to initialise both the `operate` (`config.build.operate.use_cap_results`) and `spores` modes.
 
 ### `config.solve.solver`
 
--- a/docs/hooks/dummy_model/model.yaml
+++ b/docs/hooks/dummy_model/model.yaml
@@ -5,13 +5,13 @@ overrides:
       time_cluster: cluster_days.csv
     config.build:
       extra_math: ["storage_inter_cluster"]
-  spores:
-    config:
-      init.name: SPORES solve mode
-      build.mode: spores
-      solve.spores.number: 2
-    parameters:
-      spores_slack: 0.1
+  # spores:
+  #   config:
+  #     init.name: SPORES solve mode
+  #     build.mode: spores
+  #     solve.spores.number: 2
+  #   parameters:
+  #     spores_slack: 0.1
 
 config.init.name: base
 
--- a/docs/hooks/dummy_model/model.yaml
+++ b/docs/hooks/dummy_model/model.yaml
@@ -3,8 +3,6 @@ overrides:
     config.init:
       name: inter-cluster storage
       time_cluster: cluster_days.csv
-      math:
-        pre_defined: ["plan", "storage_inter_cluster"]
     config.build:
       extra_math: ["storage_inter_cluster"]
   spores:
--- a/docs/hooks/dummy_model/model.yaml
+++ b/docs/hooks/dummy_model/model.yaml
@@ -5,13 +5,13 @@ overrides:
       time_cluster: cluster_days.csv
     config.build:
       extra_math: ["storage_inter_cluster"]
-  # spores:
-  #   config:
-  #     init.name: SPORES solve mode
-  #     build.mode: spores
-  #     solve.spores.number: 2
-  #   parameters:
-  #     spores_slack: 0.1
+  spores:
+    config:
+      init.name: SPORES mode
+      build.mode: spores
+      solve.spores.number: 2
+    parameters:
+      spores_slack: 0.1
 
 config.init.name: base
 
--- a/docs/hooks/dummy_model/model.yaml
+++ b/docs/hooks/dummy_model/model.yaml
@@ -3,8 +3,10 @@ overrides:
     config.init:
       name: inter-cluster storage
       time_cluster: cluster_days.csv
+      math:
+        pre_defined: ["plan", "storage_inter_cluster"]
     config.build:
-      add_math: ["storage_inter_cluster"]
+      extra_math: ["storage_inter_cluster"]
   spores:
     config:
       init.name: SPORES solve mode
--- a/docs/user_defined_math/customise.md
+++ b/docs/user_defined_math/customise.md
@@ -18,16 +18,15 @@ storage_max:
 
 The other elements of the `storage_max` constraints have not changed (`foreach`, `where`, ...), so we do not need to define them again when adding our own twist on the pre-defined math.
 
-When defining your model, you can reference any number of YAML files containing the math you want to add in `config.init.math`.
+When defining your model, you can reference any number of YAML files containing the math you want to add in `config.init.extra_math`.
 Both absolute paths and paths relative to `model.yaml` are valid.
 
 ```yaml
 config:
   init:
-    math:
-      extra:
-        my_new_math_1: "my_new_math_1.yaml"
-        my_new_math_2: "/home/your_name/Documents/my_new_math_2.yaml"
+    extra_math:
+      my_new_math_1: "my_new_math_1.yaml"
+      my_new_math_2: "/home/your_name/Documents/my_new_math_2.yaml"
 ```
 
 You can then add this math during `build` by specifying it in `config.build.extra_math`.
@@ -59,9 +58,8 @@ This will tell Calliope to overwrite *all* of our pre-defined `base` math with y
 ```yaml
 config:
   init:
-    math:
-      base: "my_new_base_math"
-      extra: {my_new_base_math: your/base_math_file.yaml}
+    base_math: "my_new_base_math"
+    extra_math: {my_new_base_math: your/base_math_file.yaml}
 ```
 
 !!! danger
--- a/docs/user_defined_math/customise.md
+++ b/docs/user_defined_math/customise.md
@@ -1,10 +1,12 @@
-# Adding your own math to a model
+# Adding your own extra math to a model
 
 Once you understand the [math components](components.md) and the [formulation syntax](syntax.md), you'll be ready to introduce your own math to a model.
 
 You can find examples of additional math that we have put together in our [math example gallery](examples/index.md).
 
-Whenever you introduce your own math, it will be applied on top of the pre-defined math for your chosen run [mode](../creating/config.md#configbuildmode).
+Whenever you introduce your own extra math, it will be applied on top of the pre-defined math for your chosen run [mode](../creating/config.md#configbuildmode).
+The order of priority is **base math -> mode -> extra math**.
+
 Therefore, you can override the pre-defined math as well as add new math.
 For example, you may want to introduce a timeseries parameter to the pre-defined `storage_max` constraint to limit maximum storage capacity on a per-timestep basis:
 
@@ -16,25 +18,25 @@ storage_max:
 
 The other elements of the `storage_max` constraints have not changed (`foreach`, `where`, ...), so we do not need to define them again when adding our own twist on the pre-defined math.
 
-!!! note
-
-    If you prefer to start from scratch with your math, you can ask Calliope to _not_ load the pre-defined math for your chosen run mode by setting `#!yaml config.build.ignore_base_math: true`.
-
-When defining your model, you can reference any number of YAML files containing the math you want to add in `config.build`.
-The paths are relative to your main model configuration file:
+When defining your model, you can reference any number of YAML files containing the math you want to add in `config.init.math`.
+Both absolute paths and paths relative to `model.yaml` are valid.
 
 ```yaml
 config:
-  build:
-    add_math: [my_new_math_1.yaml, my_new_math_2.yaml]
+  init:
+    math:
+      extra:
+        my_new_math_1: "my_new_math_1.yaml"
+        my_new_math_2: "/home/your_name/Documents/my_new_math_2.yaml"
 ```
 
-You can also define a mixture of your own math and the [pre-defined math](../pre_defined_math/index.md):
+You can then add this math during `build` by specifying it in `config.build.extra_math`.
+It is even possible to define a mixture of your math and other [pre-defined math](../pre_defined_math/index.md):
 
 ```yaml
 config:
   build:
-    add_math: [my_new_math_1.yaml, storage_inter_cluster, my_new_math_2.md]
+    extra_math: [my_new_math_1, storage_inter_cluster, my_new_math_2]
 ```
 
 Finally, when working in an interactive Python session, you can add math as a dictionary at build time:
@@ -43,11 +45,28 @@ Finally, when working in an interactive Python session, you can add math as a di
 model.build(add_math_dict={...})
 ```
 
-This will be applied after the pre-defined mode math and any math from file listed in `config.build.add_math`.
+This will be applied after the pre-defined mode math and any extra math listed in `config.build.extra_math`.
 
 !!! note
 
     When working in an interactive Python session, you can view the final math dictionary that has been applied to build the optimisation problem by inspecting `model.applied_math` after a successful call to `model.build()`.
+
+## Re-defining Calliope's pre-defined base math
+
+If you prefer to start from scratch with your math, you can ask Callipe to use it as the base of the problem.
+This will tell Calliope to overwrite *all* of our pre-defined `base` math with your file.
+
+```yaml
+config:
+  init:
+    math:
+      base: "my_new_base_math"
+      extra: {my_new_base_math: your/base_math_file.yaml}
+```
+
+!!! danger
+
+    Modes and other pre-defined options such as `operate` and `spores` might not work as expected!
 
 ## Adding your parameters to the YAML schema
 
--- a/docs/user_defined_math/customise.md
+++ b/docs/user_defined_math/customise.md
@@ -18,7 +18,7 @@ The other elements of the `storage_max` constraints have not changed (`foreach`,
 
 !!! note
 
-    If you prefer to start from scratch with your math, you can ask Calliope to _not_ load the pre-defined math for your chosen run mode by setting `#!yaml config.build.ignore_mode_math: true`.
+    If you prefer to start from scratch with your math, you can ask Calliope to _not_ load the pre-defined math for your chosen run mode by setting `#!yaml config.build.ignore_base_math: true`.
 
 When defining your model, you can reference any number of YAML files containing the math you want to add in `config.build`.
 The paths are relative to your main model configuration file:
--- a/docs/user_defined_math/customise.md
+++ b/docs/user_defined_math/customise.md
@@ -1,13 +1,15 @@
-# Adding your own extra math to a model
+# Adding your own math to a model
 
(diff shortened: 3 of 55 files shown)


irm-codebase:
Tests are failing due to the issues seen in #754 
Before moving forward with doc

To do list:
- [ ] Clean up the docs
- [ ] Avoid setting defaults in the backend where relevant (we should respect the schema as much as possible).
- [ ] Change the default mode from 'plan' back to 'base' to avoid confusion when users decide to define their own base math.
- [ ] Remove `model.def_path` (should no longer be necessary since we now fully load all files during init).

irm-codebase:
Mini-diagram (again, based on https://github.com/calliope-project/calliope/issues/642#issuecomment-2465389387) showing the state of the model steps after this PR.
The main advantage is that we could know the dimensionality of our variables / parameters from the start (once we include them in the YAML). This should make extracting subsets of them (e.g., reduced timeseries) much simpler.

image

brynpickering:
Nice take on it. Initial thought based on your high-level comments, not having looked at the committed code yet:

You say that we would now `support knowing parameters / dimension indexes during init`, but is it not still the case that until we actually select our math in `config.build`, we can't be certain what math has been chosen (and therefore the params/dims)? 
E.g., in your first example, what if `test_formulation1` and `test_formulation2` set different parameters/dims? Or, more concretely, what if you set the `config.build.mode` as `base` vs. `operate`? These two have different math parameters.

It might be necessary to enforce the math (and mode) at `init` if we want to be able to process the params/dims at that point, too...

irm-codebase:
> Nice take on it. Initial thought based on your high-level comments, not having looked at the committed code yet:
> 
> You say that we would now `support knowing parameters / dimension indexes during init`, but is it not still the case that until we actually select our math in `config.build`, we can't be certain what math has been chosen (and therefore the params/dims)? E.g., in your first example, what if `test_formulation1` and `test_formulation2` set different parameters/dims? Or, more concretely, what if you set the `config.build.mode` as `base` vs. `operate`? These two have different math parameters.
> 
> It might be necessary to enforce the math (and mode) at `init` if we want to be able to process the params/dims at that point, too...

Hmm. This is true. Here is a general idea that may solve this issue (and even make switching a variable to a parameter easier):

1. We should consider variables and parameters as the same 'thing'. They are essentially the same data-wise (a value with indexes, one is know, the other isn't). During `init`, we still just extract those specified as parameters.
2. The math schema should guarantee that parameter / variable names (perhaps ALL math names?) are unique. The backend already enforces this through `valid_component_names`.
3. We forbid the addition of new parameters / variables in `mode` or `extra`, and we allow them to switch (e.g., use `flow_cap` as variable in `plan`, and as a parameter in `operate`). Dims are similarly 'hard set' after init.

Do you think this is sensible?

brynpickering:
It's definitely going in the right direction!

Not having `parameters` and `variables` as top-level keys is a bit strange, although I do agree that they are effectively the same thing once the model has finished optimising. I'd prefer that they are still under those top-level keys in YAML, if possible, unless we can think of the perfect name to encapsulate the two. We also want `parameters` to have a data type unless we very explicitly split out exclusively "decision" parameters (usually boolean or string, but could be numeric) that turn up in where strings and subsetting and "math" parameters (exclusively numeric) which are actually part of the equations.

Keep in mind that we would use the definition of parameters to type cast input data. Would we then say that if an input exists that is listed as a variable in the `init` math, we wouldn't use its metadata to validate that input dataset? Or would we only validate it at `build`, once all parameters<->variables switching has been done?

Could we say that math has to be defined at `init` and only a subset of it can only be overriden at `build`? e.g., only the objective, global expressions, and constraints can actually be edited at build but variables/parameters are fixed. This would mean e.g. an operate mode model would have to be set as such at `init` and you couldn't do run plan -> run operate without re-initialising the model.

irm-codebase:
@brynpickering some comments:

> Not having `parameters` and `variables` as top-level keys is a bit strange, although I do agree that they are effectively the same thing once the model has finished optimising. I'd prefer that they are still under those top-level keys in YAML, if possible, unless we can think of the perfect name to encapsulate the two. 

I agree that keeping them separate would be more 'natural' to users. We just need a method that swaps between them in the math schema, which should be easy to implement... Although, a benefit of joining them is that it might enable 'switching'  vars/params a calliope YAML math feature, which could be quite powerful.

Both seem reasonably doable.

>We also want `parameters` to have a data type unless we very explicitly split out exclusively "decision" parameters (usually boolean or string, but could be numeric) that turn up in where strings and subsetting and "math" parameters (exclusively numeric) which are actually part of the equations.
> 
> Keep in mind that we would use the definition of parameters to type cast input data. Would we then say that if an input exists that is listed as a variable in the `init` math, we wouldn't use its metadata to validate that input dataset? Or would we only validate it at `build`, once all parameters<->variables switching has been done?

This should be easy to do, too. We just type-cast both variables and parameters during init, and only allow variables to be type-casted to numeric values in the schema. If a user tries to 'swap' a string parameter to a variable, `pydantic` should deny it.

We could separate 'decision' parameters if it gets messy, but only if necessary.

> Could we say that math has to be defined at `init` and only a subset of it can only be overriden at `build`? e.g., only the objective, global expressions, and constraints can actually be edited at build but variables/parameters are fixed. This would mean e.g. an operate mode model would have to be set as such at `init` and you couldn't do run plan -> run operate without re-initialising the model.

Hmm... Not allowing people to switch modes kind of defeats the purpose of having modes in the first place, and it is wasteful since all the info is there after `plan` is run, at least for spores and operate. 

For now, here is how I see things:
- objectives / global expressions / constraints can be modified freely. As you say, they do not get in the way.
- It is probably fine to switch a parameter to a variable and vice-versa at `build`, as long as it is numeric and its dimensions are respected.
- It is probably fine for a mode / extra math to add a new variable at `build` _as long as it does not introduce new dimensions_.
- Parameters specific to modes / extra math are tricky due to the data extraction stage.
   - Having all parameters specified in the base math would be the easiest solution, but it would mean that a base file like `plan` would define parameters only used in `spores` (kind of like how `model_def_schema.yaml` does it at the moment). Basically, base files would have to be very intentional in saying what they 'support'.
   - An alternative would be to add `init` logic that 'detects' parameters only defined in mode / extra files, and uses those definitions during data extraction. I fear this would be complicated to implement and support, since it increases ambiguity.

brynpickering:
> Hmm... Not allowing people to switch modes kind of defeats the purpose of having modes in the first place, and it is wasteful since all the info is there after plan is run, at least for spores and operate.

Could we handle providing an existing Calliope model to `calliope.Model`, allowing you to re-initialise a model without needing to duplicate data in memory (we could default to a shallow copy)? We could then force setting the `mode` at init and then a user would do `init in plan` -> `solve in plan` -> `re-init in operate` -> `solve in operate`.

irm-codebase:
> Could we handle providing an existing Calliope model to `calliope.Model`, allowing you to re-initialise a model without needing to duplicate data in memory (we could default to a shallow copy)? We could then force setting the `mode` at init and then a user would do `init in plan` -> `solve in plan` -> `re-init in operate` -> `solve in operate`.

Possible, yeah. I am not sure if it's really necessary, though. You'd be 'duplicating' the model object (even if not the data), and it's a bit unintuitive. It'd be the same as just allowing the model to 're-initialise' some bits of itself using its own data (basically just the math?). No need to create a second object for that...

Also, it could lead to 'lossy' cases if we are not careful: if `operate` is layered on top of the base math and then _becomes_ the base math, it's not a state you can go back from. That model can never be run in `plan` mode again. In either case keeping base / modes / extra is still necessary.

Personally, I like the idea of switching between variables/parameters from the YAML math bit more... it feels 'cleaner', but I'd have to think about it more... that's a decision for this https://github.com/calliope-project/calliope/pull/712 behemoth (which comes after this one).

docs/hooks/dummy_model/model.yaml
```diff
@@ -3,8 +3,10 @@ overrides:
     config.init:
       name: inter-cluster storage
       time_cluster: cluster_days.csv
+      math:
+        pre_defined: ["plan", "storage_inter_cluster"]
```
brynpickering: Open question from Element discussion: do we need `config.init.math.pre_defined`? or is it sufficient to pre-load all pre-defined math and then leave it to be added as necessary via `config.build.mode` or in `config.build.extra_math`?

docs/user_defined_math/customise.md
```diff
@@ -18,7 +18,7 @@ The other elements of the `storage_max` constraints have not changed (`foreach`,
 
 !!! note
 
-    If you prefer to start from scratch with your math, you can ask Calliope to _not_ load the pre-defined math for your chosen run mode by setting `#!yaml config.build.ignore_mode_math: true`.
+    If you prefer to start from scratch with your math, you can ask Calliope to _not_ load the pre-defined math for your chosen run mode by setting `#!yaml config.build.ignore_base_math: true`.
```
brynpickering: I suppose this is slightly different functionality.

If you set `config.init.math.base: plan` and then `config.build.mode: operate`, what does `config.build.ignore_base_math` do?

src/calliope/backend/gurobi_backend_model.py
```diff
@@ -14,16 +14,17 @@
...
 from calliope.exceptions import BackendError, BackendWarning
 from calliope.exceptions import warn as model_warn
-from calliope.preprocess import CalliopeMath
 from calliope.schemas import config_schema
 
 if importlib.util.find_spec("gurobipy") is not None:
     import gurobipy
 
 T = TypeVar("T")
+# TODO-Ivan: below seems to be repated in several files? Fix.
```
brynpickering: I noticed this too!

src/calliope/model.py
```diff
@@ -269,19 +264,14 @@ def build(
         else:
             backend_input = self._model_data
 
-        init_math_list = [] if self.config.build.ignore_mode_math else [mode]
-        end_math_list = [] if add_math_dict is None else [add_math_dict]
-        full_math_list = init_math_list + self.config.build.add_math + end_math_list
-        LOGGER.debug(f"Math preprocessing | Loading math: {full_math_list}")
-        model_math = preprocess.CalliopeMath(full_math_list, self.def_path)
-
+        self.applied_math = preprocess.build_applied_math(
```
brynpickering: I would only attach this to `self` if the backend build succeeds as it hasn't actually been applied until then. Say that, attaching it first does act as a way to debug the resolved math definition if build fails...

src/calliope/model.py
```diff
@@ -111,7 +107,12 @@ def __init__(
     @property
     def name(self):
         """Get the model name."""
-        return self._model_data.attrs["name"]
+        return self._def.config.init.name
+
+    @property
+    def config(self) -> config_schema.CalliopeConfig:
+        """Get model configuration values."""
+        return self._def.config
```
brynpickering, line 115: What else is stored within `self._def` other than `config`?

src/calliope/preprocess/model_definition.py
```diff
@@ -8,40 +8,57 @@
...
     override_dict: dict | None = None,
+    **kwargs,
 ) -> tuple[AttrDict, str]:
     """Arrange model definition data folloging our standardised order of priority.
 
     Should always be called when defining calliope models from configuration files.
     The order of priority is:
 
-    - override_dict > scenarios > data section > template
+    - init_kwargs > override_dict > scenarios > data section > template
```
brynpickering, line 29: The override dict comes from the init kwargs. What else is there?

src/calliope/model.py
brynpickering: aha, `math`.

src/calliope/preprocess/model_math.py
```diff
@@ -24,156 +24,74 @@
...
+        ModelError: a given name was not found in the math dictionary.
+
+    Returns:
+        AttrDict: constructed and validated math dataset.
+    """
+    names = []
+    if not config.build.ignore_base_math:
+        names.append(config.init.math.base)
+    if config.build.mode != "base":
+        names.append(config.build.mode)
```
brynpickering: So what happens if `init.math.base: operate` and `build.mode: operate`? This will be a duplication? 

I realise this won't work as `base: operate` will fail because it is an override of `plan`. So whatever error comes from defining both as `operate` might be confusing to the user.

src/calliope/preprocess/model_math.py
```diff
@@ -24,156 +24,74 @@
...
-        self.history.append(name)
-        LOGGER.info(f"Math preprocessing | added file '{name}'.")
+            math.union(init_math[name], allow_override=True)
+        except KeyError:
+            raise ModelError(f"Requested math '{name}' was not initialised.")
+    if add_math:
+        LOGGER.info("Math build | appending additional math.")
+        math.union(add_math, allow_override=True)
+
+    return AttrDict(MathSchema(**math).model_dump())
```
brynpickering, line 97: I prefer the idea of keeping these as pydantic models as you get type hints, autocomplete, and tooltips when scripting. The data structure is lost entirely when dumped to an attribute dictionary.

src/calliope/schemas/config_schema.py
```diff
@@ -8,9 +8,9 @@
 
 from pydantic import Field
 
-from calliope.schemas.general import CalliopeBaseModel, UniqueList
+from calliope.schemas.general import AttrStr, CalliopeBaseModel, UniqueList
 
-Mode = Literal["plan", "operate", "spores"]
+Mode = Literal["base", "operate", "spores"]
```
brynpickering, line 13: This is how we make things clearer (plan -> base) w.r.t. operate/spores being an overlay on base?

brynpickering: Hmm, no, `plan` still exists. What if a user wants to do `base: operate`, `mode: plan`? As in, setting `plan` explicitly as the mode. I realise these are edge cases but I find it slightly confusing that I have to look back at `init.base` to know that `base` "mode" == `plan`

src/calliope/schemas/general.py
```diff
@@ -76,12 +76,11 @@ def update(self, update_dict: dict, deep: bool = False) -> Self:
                 new_dict[key] = key_class.update(val)
             else:
                 LOGGER.info(
-                    f"Updating {self.model_config['title']} `{key}`: {key_class} -> {val}"
+                    f"Updating {self.__class__.__name__} `{key}`: {key_class} -> {val}"
```
brynpickering, line 79: Isn't `title` more human-readable?

src/calliope/schemas/math_schema.py
brynpickering, line 1: should the backend/parsing.py type hint classes just inherit directly from this file so we don't duplicate that info?

tests/test_backend_general.py
```diff
@@ -354,6 +355,7 @@ def test_add_constraint(self, solved_model_func):
                 {"expression": "sum(flow_out, over=[nodes, timesteps]) >= 100"}
             ],
             "where": "carrier_out",  # <- no error is raised because of this
+            "active": True,
```
brynpickering, line 358: can this simply default to True in the pydantic model?

docs/hooks/dummy_model/model.yaml
irm-codebase: Given that `plan.yaml` might need to be split in different 'extras' (see https://github.com/calliope-project/calliope/issues/749), I think we should keep it with all of our files in it by default.

The thought of invalid math 'living' in the produced model file concerns me.
If that turns to be a problem in the future, we can move to a setup that only includes `plan` by default.

docs/user_defined_math/customise.md
irm-codebase: This has been eliminated in the latest version. This is no longer necessary, since you can replace the base math from the get-go.

src/calliope/model.py
irm-codebase: The full model definition schema, actually.

src/calliope/preprocess/model_definition.py
irm-codebase: It does not, actually.
`Model.init` has both `override_dict` and `**kwargs` (`init_kwargs` in this case).

src/calliope/preprocess/model_math.py
irm-codebase: Correct, this will result in a failure.
I think this is the kind of edge case that is best avoided with good docs rather than code.

src/calliope/preprocess/model_math.py
irm-codebase: I did try it in this PR, but it was a gargantuan task best left to a future PR.
The _entire backend_ needs to be changed in how it references to the math.

src/calliope/schemas/config_schema.py
irm-codebase: This would not work well because users can set their own files as `base`.
I think it's much less confusing, and much easier to maintain on our side, to specify `base` as the mode given that the name is ambiguous.

I've reverted our documentation to `base` to avoid this confusion.

src/calliope/schemas/general.py
irm-codebase: It's incompatible with the latest `pydantic` update. It was out of necessity.

tests/test_backend_general.py
irm-codebase: Not unless we move the full backend to support pydantic.
Which I'd love, but it proved too complex / large a task for this PR.

src/calliope/model.py
brynpickering: But none of the CSV data, ofc. This was the problem we had before with attaching the full model definition schema to the model object (and then storing it in the NetCDF) is that without the timeseries data it's questionable what value the non-config information has to be kept around long-term.

docs/hooks/dummy_model/model.yaml
brynpickering: We seem to be going in circles on this point. There are two steps right now:

1. Loading all _possible_ math files in and having them sit around waiting to be called up (`init`)
2. Calling up the math files of those that are available to use in building the optimisation problem (`build`)

It doesn't seem necessary to me to force users to specify which pre-defined math to load in (1) because it doesn't hurt anyone to load all of them alongside any user-defined math. They don't pollute anything until actually invoked by calling them up in (2). Yes, they all then "live" in the model but they are always lurking in the background anyway (anytime you import `calliope` you can technically access them). I just don't see the benefit of forcing the user to remember to load them _both_ at `init` and at `build`, especially since if you forget to load them at `init` and then save and re-load the model in future you have no way of accessing the pre-defined math you forgot to include in your `init` list.

irm-codebase:
@brynpickering thanks for the feedback!

Things to do, then:
1. I will axe `pre_defined`, and add all of our files.
2. I'll improve the error users get when requesting invalid math.

src/calliope/schemas/config_schema.py
```diff
@@ -19,15 +19,29 @@
...
+    """
+
+    base: AttrStr = "plan"
+    """Name of the math file to build on top of.
+    Can be any pre-defined math file or user-defined `extra` file.
+    """
+    extra: dict[AttrStr, str] = Field(default={})
+    "Dictionary with the names and paths of additional math files."
+
+
```
brynpickering: With this now only having two options and in an effort to reduce nesting where possible, could these now be `base_math` and `extra_math` params within `init`? It would align with having `extra_math` in `build` (or `add_math` in build?).

src/calliope/schemas/config_schema.py
```diff
@@ -19,15 +19,29 @@
...
+    Files will be loaded in the following order: base -> mode -> extra.
+    """
+
+    base: AttrStr = "plan"
+    """Name of the math file to build on top of.
+    Can be any pre-defined math file or user-defined `extra` file.
+    """
+    extra: dict[AttrStr, str] = Field(default={})
+    "Dictionary with the names and paths of additional math files."
```
brynpickering: In an effort to fail fast, could we have a validation step which checks whether `base` is either an extra key or a pre-defined math name?

src/calliope/schemas/config_schema.py
```diff
@@ -151,7 +164,7 @@ class SolveSpores(CalliopeBaseModel):
     """
 
     skip_baseline_run: bool = Field(default=False)
-    """If the model already contains `plan` mode results, use those as the initial base run results and start with SPORES iterations immediately."""
+    """If the model already contains `plan` results, use those as the initial base run results and start with SPORES iterations immediately."""
```
brynpickering:
```suggestion
    """If the model already contains `base` results, use those as the initial base run results and start with SPORES iterations immediately."""
```

Right? We've now removed all reference to `plan` in the docs at this point I think

docs/creating/config.md
```diff
@@ -78,12 +78,12 @@ It has a very high cost associated with its use, so it will only appear when abs
 ### `config.build.mode`
 
 In the `build` section we have `mode`.
-A model can run in `plan`, `operate`, or `spores` mode.
-In `plan` mode, capacities are determined by the model, whereas in `operate` mode, capacities are fixed and the system is operated with a receding horizon control algorithm.
-In `spores` mode, the model is first run in `plan` mode, then run `N` number of times to find alternative system configurations with similar monetary cost, but maximally different choice of technology capacity and location (node).
+A model can run in `base`, `operate`, or `spores` mode.
+In `base` mode, capacities are determined by the model, whereas in `operate` mode, capacities are fixed and the system is operated with a receding horizon control algorithm.
```
brynpickering: This could be confusing if someone updates `base` to their own math as `base` mode is then whatever they have defined and may not include capacities being determined by the model...

docs/user_defined_math/customise.md
```diff
@@ -1,10 +1,12 @@
...
+# Adding your own extra math to a model
 
 Once you understand the [math components](components.md) and the [formulation syntax](syntax.md), you'll be ready to introduce your own math to a model.
 
 You can find examples of additional math that we have put together in our [math example gallery](examples/index.md).
 
-Whenever you introduce your own math, it will be applied on top of the pre-defined math for your chosen run [mode](../creating/config.md#configbuildmode).
+Whenever you introduce your own extra math, it will be applied on top of the pre-defined math for your chosen run [mode](../creating/config.md#configbuildmode).
+The order of priority is **base math -> mode -> extra math**.
+
```
brynpickering: Again, it is missing the part where your extra math _is_ the `base` math.

brynpickering (changes requested):
Still some minor bits to clean up, but looks good to me overall! I can imagine some bits changing when we introduce YAML params/dims config as we still haven't quite addressed the issue of knowing all the proposed variables/params/dims in a model until `build` as that's when the math gets fully resolved. But that can wait for now.

brynpickering (approved):
LGTM! Thanks for the effort on this one @irm-codebase

PR #760, opened 2025-05-16
Fix typos

omahs:
Fix typos

### DIFF
--- a/src/calliope/backend/backend_model.py
+++ b/src/calliope/backend/backend_model.py
@@ -535,7 +535,7 @@ class BackendModelGenerator(ABC):
         return da
 
     def _raise_error_on_preexistence(self, key: str, obj_type: ALL_COMPONENTS_T):
-        """Detect if preexistance errors are present the dataset.
+        """Detect if preexistence errors are present in the dataset.
 
         We do not allow any overlap of backend object names since they all have to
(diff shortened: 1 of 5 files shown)


src/calliope/backend/backend_model.py
```diff
@@ -535,7 +535,7 @@ def _apply_func(
         return da
 
     def _raise_error_on_preexistence(self, key: str, obj_type: ALL_COMPONENTS_T):
-        """Detect if preexistance errors are present the dataset.
+        """Detect if preexistence errors are present in the dataset.
```
brynpickering, line 538:
```suggestion
        """Detect if preexistent errors are present in the dataset.
```

brynpickering (changes requested):
Thanks for picking up on these @omahs !

src/calliope/backend/backend_model.py
omahs: Thanks!

brynpickering (approved):
LGTM

irm-codebase:
Interesting.
Seems like this is a bot that is committing typo fixes to random repos.

omahs:
@irm-codebase Hey, I'm not a bot but for sure we will see more automated PRs as time goes.

PR #769, opened 2025-06-11
Docs fixes

brynpickering:
I noticed a few docs issues while building changes from #756 locally. These should resolve them.

## Reviewer checklist

- [x] Documentation updated
- [ ] Changelog updated


mkdocs.yml
```diff
@@ -112,6 +112,7 @@ nav:
   - Troubleshooting: troubleshooting.md
   - Pre-defined math:
     - index: pre_defined_math/index.md
+    - pre_defined_math/mode.md
```
irm-codebase: @brynpickering the reason why I removed this was to ensure the Base math section appears above the build modes one in the ToC.

image

irm-codebase: I figured that would be a better reading order.
Is there a way to re-define this?

irm-codebase (changes requested):

irm-codebase:
@brynpickering It seems like something in the SPORES formulation is random failures.
I've seen this a couple of times since https://github.com/calliope-project/calliope/pull/754 was merged, but I'm unsure of what is causing them...

It's not the topic of this PR, though. I created #770 to track it.

mkdocs.yml
brynpickering: It may well be a better reading order but without defining it it never turns up in the navigation (which is odd!). I'll update ordering.

irm-codebase (commented):
Nice fix on the SPORES problem @brynpickering !
Seems like the ToC is still a bit finicky, though.

image

irm-codebase:
Perhaps we should consider combining the build modes section and the `index.md` file?
It'd avoid the issue, and perhaps simplify our docs without making them more difficult to understand.

brynpickering:
@irm-codebase I've moved it into `index.md`; see what you think.

irm-codebase (approved):
LGTM!

Issue #738, opened 2025-02-06
Moving to `pydantic`: isolate dimension-related data (i.e., `techs`, `nodes`).


irm-codebase:
### What can be improved?

'Mutable' dimensional data should be isolated under its own key in the schema.

```yaml
config:  # structural, contains stuff that affects code behaviour
math: # structural, contains stuff that affects backend mathematics
data: # structural, contains data definition in YAML
    techs: # data for a dimension
    nodes: # data for a dimension
    vintages:  # data for a future 'vintages' dimension
data_tables: # structural, contains stuff that defines extraction from files
```

This would allow use to add / remove dimension functionalities without impacting other portions of the configuration files.

See #726 for more information

### Version

v0.7

PR #771, opened 2025-06-13
Update to inputs/results; separate attrs and store as pydantic model

brynpickering:
@irm-codebase in doing the dim/param config work I realised I was bloating it with lots of work to separate out the inputs/results datasets. So, here's a WIP of what I'm thinking for that.

Some tests still fail and I'm working on that but wanted to get your view on it in the meantime!

- The results and inputs datasets are completely separate and stored as independent "group" in the NetCDF
- The model attributes are never attached to these datasets and are instead a pydantic model attached to the calliope model which is saved into its own NetCDF "group" on saving
- This means that I have to currently pass parameter defaults explictly to the backend (they're not stored in the inputs attrs). This would then go with #712 
- Initialising a model is done with my pandas/xarray-esque command (`read_yaml`, `from_dicts`) to allow loading a model from dataset(s) (`from_datasets`). This removes the messy process we currently have for inferring how to process user inputs when they call `calliope.Model`. I'm not sure I like my current approach of having `calliope.Model()` returning an empty shell though. That needs some thought.
- `model._def` -> `model.model_def` as it is now part of a pydantic model and pydantic doesn't see private attrs as valid fields (so you can't update it). Might need a better name or perhaps even be better exploded out into its constituent parts in `model.attrs`.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -14,3 +14,4 @@ pytest >= 8
 pytest-cov >= 4
 pytest-order >= 1
 pytest-xdist >= 3 # pytest distributed testing plugin
+rich >= 14.0.0
--- a/src/calliope/__init__.py
+++ b/src/calliope/__init__.py
@@ -1,10 +1,16 @@
 """Public packaging of Calliope."""
 
+import xarray as xr
+from rich import pretty
+
 from calliope import examples, exceptions
 from calliope._version import __version__
 from calliope.attrdict import AttrDict
-from calliope.model import Model, read_netcdf
+from calliope.model import Model, read_netcdf, read_yaml
 from calliope.util.logging import set_log_verbosity
+
+xr.set_options(display_expand_attrs=False)
+pretty.install(max_depth=1)
 
 __title__ = "Calliope"
 __author__ = "Calliope contributors listed in AUTHORS"
--- a/src/calliope/__init__.py
+++ b/src/calliope/__init__.py
@@ -2,7 +2,7 @@
 
 from rich import pretty
 
-from calliope import examples, exceptions
+from calliope import examples
 from calliope._version import __version__
 from calliope.attrdict import AttrDict
 from calliope.model import Model, read_netcdf, read_yaml
--- a/src/calliope/__init__.py
+++ b/src/calliope/__init__.py
@@ -1,6 +1,5 @@
 """Public packaging of Calliope."""
 
-import xarray as xr
 from rich import pretty
 
 from calliope import examples, exceptions
@@ -9,7 +8,6 @@ from calliope.attrdict import AttrDict
 from calliope.model import Model, read_netcdf, read_yaml
 from calliope.util.logging import set_log_verbosity
 
-xr.set_options(display_expand_attrs=False)
 pretty.install(max_depth=1)
 
 __title__ = "Calliope"
--- a/src/calliope/__init__.py
+++ b/src/calliope/__init__.py
@@ -1,6 +1,6 @@
 """Public packaging of Calliope."""
 
-from rich import pretty
+import importlib
 
 from calliope import examples
 from calliope._version import __version__
@@ -8,7 +8,12 @@ from calliope.attrdict import AttrDict
 from calliope.model import Model, read_netcdf, read_yaml
 from calliope.util.logging import set_log_verbosity
 
-pretty.install(max_depth=1)
+try:
+    from rich import pretty
+
+    pretty.install(max_depth=1)
+except ModuleNotFoundError:
+    pass
 
 __title__ = "Calliope"
 __author__ = "Calliope contributors listed in AUTHORS"
--- a/src/calliope/backend/backend_model.py
+++ b/src/calliope/backend/backend_model.py
@@ -4,7 +4,6 @@
 
 from __future__ import annotations
 
-import functools
 import logging
 import time
 import typing
@@ -45,21 +44,6 @@ ALL_COMPONENTS_T = Literal["parameters", ORDERED_COMPONENTS_T]
 
 
 LOGGER = logging.getLogger(__name__)
-
-
-def add_new_component(func):
-    """When calling wrapped adder methods, store the math updates in the backend dataset."""
-
-    @functools.wraps(func)
-    def func_wrapper(self, name, component_dict):
-        component_type = func.__name__.removeprefix("add_") + "s"
-        if name not in self.math[component_type]:
-            self._dataset.attrs["applied_math"].union(
-                {f"{component_type}.{name}": component_dict}
-            )
-        return func(self, name, component_dict)
-
-    return func_wrapper
 
 
 class BackendModelGenerator(ABC, metaclass=ABCMeta):
@@ -380,6 +364,8 @@ class BackendModelGenerator(ABC, metaclass=ABCMeta):
         self._add_to_dataset(
             name, component_da, component_type, component_dict, references
         )
+        if name not in self.math[component_type]:
+            self._dataset.attrs["applied_math"].union(component_dict)
 
         return parsed_component
 
@@ -416,9 +402,7 @@ class BackendModelGenerator(ABC, metaclass=ABCMeta):
             model_data (xr.Dataset): Input model data.
             defaults (dict): Parameter defaults.
         """
-        for param_name, param_data in self.inputs.filter_by_attrs(
-            is_result=0
-        ).data_vars.items():
+        for param_name, param_data in self.inputs.data_vars.items():
             default_val = param_data.attrs.get("default", np.nan)
             self.add_parameter(param_name, param_data, default_val)
         for param_name, default_val in self.defaults.items():
@@ -434,7 +418,6 @@ class BackendModelGenerator(ABC, metaclass=ABCMeta):
                 "parameters", param_name, "Component not defined; using default value."
             )
             self.add_parameter(param_name, xr.DataArray(np.nan), default_val)
-            self.parameters[param_name].attrs["is_result"] = 0
         LOGGER.info("Optimisation Model | parameters | Generated.")
 
     @staticmethod
@@ -652,21 +635,6 @@ class BackendModel(BackendModelGenerator, Generic[T]):
         self._instance = instance
         self.shadow_prices: ShadowPrices
         self._has_verbose_strings: bool = False
-        for func in [
-            f"add_{component.removesuffix('s')}"
-            for component in typing.get_args(ORDERED_COMPONENTS_T)
-        ]:
-            if func in self.__dict__:
-                setattr(self, func, add_new_component(self.__dict__[func]))
-
-    def __init_subclass__(cls, **kwargs):
-        """When calling wrapped adder methods, store the math updates in the backend dataset."""
-        for func in [
-            f"add_{component.removesuffix('s')}"
-            for component in typing.get_args(ORDERED_COMPONENTS_T)
-        ]:
-            if func in cls.__dict__:
-                setattr(cls, func, add_new_component(cls.__dict__[func]))
 
     def add_piecewise_constraint(  # noqa: D102, override
         self, name: str, constraint_dict: parsing.UnparsedPiecewiseConstraint
--- a/src/calliope/backend/backend_model.py
+++ b/src/calliope/backend/backend_model.py
@@ -365,7 +365,9 @@ class BackendModelGenerator(ABC, metaclass=ABCMeta):
             name, component_da, component_type, component_dict, references
         )
         if name not in self.math[component_type]:
-            self._dataset.attrs["applied_math"].union(component_dict)
+            self._dataset.attrs["applied_math"].union(
+                {f"{component_type}.{name}": component_dict}
+            )
 
         return parsed_component
 
--- a/src/calliope/backend/backend_model.py
+++ b/src/calliope/backend/backend_model.py
@@ -4,10 +4,11 @@
 
 from __future__ import annotations
 
+import functools
 import logging
 import time
 import typing
-from abc import ABC, abstractmethod
+from abc import ABC, ABCMeta, abstractmethod
 from collections.abc import Callable, Iterable, Iterator
 from contextlib import contextmanager
 from copy import deepcopy
@@ -29,6 +30,7 @@ import xarray as xr
 from calliope import exceptions
 from calliope.attrdict import AttrDict
 from calliope.backend import helper_functions, parsing
+from calliope.exceptions import BackendError
 from calliope.exceptions import warn as model_warn
 from calliope.io import load_config, to_yaml
 from calliope.preprocess.model_math import ORDERED_COMPONENTS_T
@@ -38,8 +40,6 @@ from calliope.util.schema import MODEL_SCHEMA, extract_from_schema
 if TYPE_CHECKING:
     from calliope.backend.parsing import T as Tp
 
-from calliope.exceptions import BackendError
-
 T = TypeVar("T")
 ALL_COMPONENTS_T = Literal["parameters", ORDERED_COMPONENTS_T]
 
@@ -47,7 +47,22 @@ ALL_COMPONENTS_T = Literal["parameters", ORDERED_COMPONENTS_T]
 LOGGER = logging.getLogger(__name__)
 
 
-class BackendModelGenerator(ABC):
+def add_new_component(func):
+    """When calling wrapped adder methods, store the math updates in the backend dataset."""
+
+    @functools.wraps(func)
+    def func_wrapper(self, name, component_dict):
+        component_type = func.__name__.removeprefix("add_") + "s"
+        if name not in self.math[component_type]:
+            self._dataset.attrs["applied_math"].union(
+                {f"{component_type}.{name}": component_dict}
+            )
+        return func(self, name, component_dict)
+
+    return func_wrapper
+
+
+class BackendModelGenerator(ABC, metaclass=ABCMeta):
     """Helper class for backends."""
 
     LID_COMPONENTS: tuple[ALL_COMPONENTS_T, ...] = typing.get_args(ALL_COMPONENTS_T)
@@ -70,20 +85,25 @@ class BackendModelGenerator(ABC):
     """Optimisation problem objective name."""
 
     def __init__(
-        self, inputs: xr.Dataset, math: AttrDict, build_config: config_schema.Build
+        self,
+        inputs: xr.Dataset,
+        math: AttrDict,
+        build_config: config_schema.Build,
+        defaults: dict,
     ):
         """Abstract base class to build a representation of the optimisation problem.
 
         Args:
             inputs (xr.Dataset): Calliope model data.
             math (AttrDict): Calliope math.
-            build_config: Build configuration options.
+            build_config (config_schema.Build): Build configuration options.
+            defaults (dict): Parameter defaults.
         """
-        self._dataset = xr.Dataset()
+        self._dataset = xr.Dataset(attrs={"applied_math": AttrDict()})
         self.inputs = inputs.copy()
-        self.inputs.attrs = deepcopy(inputs.attrs)
         self.config = build_config
         self.math: AttrDict = deepcopy(math)
+        self.defaults: dict = deepcopy(defaults)
         self._solve_logger = logging.getLogger(__name__ + ".<solve>")
 
         self._check_inputs()
@@ -209,6 +229,7 @@ class BackendModelGenerator(ABC):
             "input_data": self.inputs,
             "build_config": self.config,
             "helper_functions": helper_functions._registry["where"],
+            "defaults": self.defaults,
             "apply_where": True,
             "references": set(),
         }
@@ -294,9 +315,6 @@ class BackendModelGenerator(ABC):
             parsing.ParsedBackendComponent | None: parsed component. None if the break_early condition was met.
         """
         references: set[str] = set()
-
-        if name not in self.math[component_type]:
-            self.math.union({f"{component_type}.{name}": component_dict})
 
         if break_early and not component_dict["active"]:
             self.log(
@@ -403,7 +421,7 @@ class BackendModelGenerator(ABC):
         ).data_vars.items():
             default_val = param_data.attrs.get("default", np.nan)
             self.add_parameter(param_name, param_data, default_val)
-        for param_name, default_val in self.inputs.attrs["defaults"].items():
+        for param_name, default_val in self.defaults.items():
             if param_name in self.parameters.keys():
                 continue
             elif (
@@ -618,6 +636,7 @@ class BackendModel(BackendModelGenerator, Generic[T]):
         inputs: xr.Dataset,
         math: AttrDict,
         build_config: config_schema.Build,
+        defaults: dict,
         instance: T,
     ) -> None:
         """Abstract base class to build backend models that interface with solvers.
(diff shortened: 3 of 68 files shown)


src/calliope/__init__.py
```diff
@@ -1,11 +1,17 @@
...
+from rich import pretty
+
 from calliope import examples, exceptions
 from calliope._version import __version__
 from calliope.attrdict import AttrDict
-from calliope.model import Model, read_netcdf
+from calliope.model import Model, read_netcdf, read_yaml
 from calliope.util.logging import set_log_verbosity
 
+xr.set_options(display_expand_attrs=False)
```
irm-codebase: Do we really need to do this, now that we intend to not hold anything there?
It risks messing with how users might expect xarray to behave, without needing it on our side.

src/calliope/backend/backend_model.py
```diff
@@ -70,20 +85,25 @@ class BackendModelGenerator(ABC):
...
         Args:
             inputs (xr.Dataset): Calliope model data.
             math (AttrDict): Calliope math.
-            build_config: Build configuration options.
+            build_config (config_schema.Build): Build configuration options.
+            defaults (dict): Parameter defaults.
         """
-        self._dataset = xr.Dataset()
+        self._dataset = xr.Dataset(attrs={"applied_math": AttrDict()})
         self.inputs = inputs.copy()
```
irm-codebase, line 87: I am quite confused here... Why do we need this, when we already have `self.math`, which is also a copy of the `applied_math` at the top layer?
This kind of duplication often leads to other issues down the line. Is there a way to avoid this?

src/calliope/backend/backend_model.py
```diff
@@ -38,16 +40,29 @@
...
 
 LOGGER = logging.getLogger(__name__)
 
 
-class BackendModelGenerator(ABC):
+def add_new_component(func):
+    """When calling wrapped adder methods, store the math updates in the backend dataset."""
+
+    @functools.wraps(func)
+    def func_wrapper(self, name, component_dict):
```
irm-codebase: Would it be possible to avoid the nested function? These have [heavy computation penalties](https://stackoverflow.com/questions/14055843/what-is-the-performance-overhead-of-nested-functions) (even ~20% is a ton, specially since this may be used in `for` loops).

`functools.wraps` is normally used to define decorators, which does not seem to be the case here...

src/calliope/backend/backend_model.py
```diff
@@ -618,20 +636,37 @@ def __init__(
...
+        super().__init__(inputs, math, build_config, defaults)
         self._instance = instance
         self.shadow_prices: ShadowPrices
         self._has_verbose_strings: bool = False
+        for func in [
+            f"add_{component.removesuffix('s')}"
+            for component in typing.get_args(ORDERED_COMPONENTS_T)
+        ]:
+            if func in self.__dict__:
+                setattr(self, func, add_new_component(self.__dict__[func]))
```
irm-codebase: This is quite hard to read (3 to 4 layers of 'nestedness')...

Suggestion:
~~How about adding `self.add_components` and `add_component`, each with a `c_type` attribute that defines what component it is (contraint, variable, etc.).~~
- ~~`self.add_components` does the loop we have here to add _all_ components of a type.~~
- ~~`self.add_component` can also be used to add a component from the user-side.~~

~~Should be easier to read, and should also remove the `removesuffix` struggle.~~

src/calliope/backend/backend_model.py
```diff
@@ -618,20 +636,37 @@ def __init__(
...
+
+    def __init_subclass__(cls, **kwargs):
+        """When calling wrapped adder methods, store the math updates in the backend dataset."""
+        for func in [
+            f"add_{component.removesuffix('s')}"
+            for component in typing.get_args(ORDERED_COMPONENTS_T)
+        ]:
+            if func in cls.__dict__:
+                setattr(cls, func, add_new_component(cls.__dict__[func]))
```
irm-codebase: Could you clarify the need for this?

src/calliope/backend/backend_model.py
irm-codebase: The above is not a good suggestion given that we already have stuff like `add_variable`.
Still, this is _very_ difficult to read, which means difficult to maintain.

The problem stems from a design clash between the specific functions (e.g., `add_variable`) and this complicated method of trying to _not call them explicitly_ in our code. Isn't there a simpler option?

src/calliope/model.py
```diff
@@ -28,212 +26,208 @@
...
+        **kwargs: initialisation overrides.
+
+    Returns:
+        Model: Calliope Model instance.
+    """
+    raw_data = io.read_rich_yaml(file)
+    definition_path = Path(file)
+    return Model.from_dict(
+        raw_data, scenario, override_dict, data_table_dfs, definition_path, **kwargs
+    )
```
irm-codebase: I love this new approach!

However, there is one **_big_** problem here: this essentially "duplicated" `Model.__init__`, which is highly risky from a software validation/verification stand-point.

This can be easily avoided by having `read_netcdf` and `read_yaml` here "standardise" stuff, and _then_ call `Model.__init__`. This provides the best of both worlds: less combinatorial complexity _and_ a single entry point to instantiate `Model`.

src/calliope/schemas/attributes.py
```diff
@@ -0,0 +1,73 @@
...
+    timings: dict[str, float] = Field(default_factory=dict)
+    """Dictionary of timings that is updated as the model is initialised/built/solved."""
+
+    termination_condition: str | None = None
+    """Indicates whether the optimisation problem solved to optimality (`optimal`) or not (e.g. `unbounded`, `infeasible`)."""
+
+    @property
+    def calliope_version_defined(self) -> str | None:
+        """Calliope version defined for this model."""
+        return self.model_def.config.init.calliope_version
```
irm-codebase: I'd consider moving this to `Model`, for intuitiveness.

src/calliope/schemas/attributes.py
irm-codebase, line 1: I like the idea of hard-defining all all of this in a schema!

Some things to think about:
1. We should consider simplifying names. Stuff like `calliope_version_defined` -> `version_defined` and `model_definition` -> `definition`.
2. I feel like `CalliopeModelDef` and `CalliopeModelAttrs` are serving the same function, since it's all 'dumped' when saving to netCDF. I see no problem in just combining them into one to 'flatten' this schema.

src/calliope/schemas/general.py
```diff
@@ -74,8 +74,22 @@ def update(self, update_dict: dict, deep: bool = False) -> Self:
             key_class = getattr(self, key)
             if isinstance(key_class, CalliopeBaseModel):
                 new_dict[key] = key_class.update(val)
+            elif (
+                isinstance(key_class, dict)
+                and isinstance(val, dict)
+                and all(isinstance(k, CalliopeBaseModel) for k in key_class.values())
+            ):
```
irm-codebase: What case triggered this?
I'm a bit concerned of this `all(... for)` loop being executed in each update. Could slow things down inadvertently.

src/calliope/schemas/attributes.py
```diff
@@ -0,0 +1,73 @@
...
+    """Full model definition (except any data table contents)"""
+
+    applied_math: MathSchema = MathSchema()
+    """Math applied when building or updating the optimisation problem."""
+
+    applied_overrides: str | None = None
+    """Overrides applied when initialising the model."""
+
+    allow_operate_mode: bool = True
+    """If False, building in `operate` mode will not be allowed."""
```
irm-codebase: I forgot about this one!
Instead of having this here, perhaps it could be a small check in `model.build` using `model.math_priority`?

src/calliope/backend/backend_model.py
brynpickering: It is a decorator, it's just a pain to add decorators to abstract methods as you need to remember the decorator in all child classes of the ABC. Anyway, it was very clunky so I've removed it.

src/calliope/model.py
brynpickering: Updated!

src/calliope/schemas/attributes.py
brynpickering: I had a play around with it and the only issue is that `attrs` becomes a little polluted with `techs`, `nodes` etc. There's benefit in having `config` at the top, but perhaps not all the others?

src/calliope/schemas/general.py
brynpickering: This is caused by not using `__pydantic_extra__` to initialise our tech/nodes/param dicts. At the point where this method reaches e.g. `calliope.Model.attrs.model_def.techs` it sees a _dict_ not a pydantic model. This means you can't update it in the pydantic way (the current approach just overrides the dict completely with the content of the new dict) and you can't use `calliope.Model.attrs.model_def.techs.model_dump()` to get a dict of dicts. 

The perhaps cleaner approach would be for e.g. calliope techs dict to be:

```
class CalliopeTechs(CalliopeBaseModel):
     __pydantic_extra__: dict[AttrStr, CalliopeTech]
     model_config = {"extra": "allow"}
```

Then, the tech dict will be a pydantic model that can be updated and dumped as any other pydantic model in our repertoire. The only issue is that accessing tech definitions in a dict-like way is a bit less neat: `tech_def.__pydantic_extra__["foo"]` (equiv to `tech_def.foo`, if you know the name of the tech you're accessing).

src/calliope/schemas/attributes.py
irm-codebase: Aha! How about leaving mutable 'dimensional' data in an isolated section such as `model.definition.data` ?
See this issue: https://github.com/calliope-project/calliope/issues/738

I planned to move these to this section as part of the params / dims stuff, but perhaps it makes sense to do this now.

src/calliope/schemas/general.py
irm-codebase: Ok... I think we should go for an option that avoids the `for` loop.
Two questions:
- Wouldn't `tech_def["foo"]` work in your example above, though?
- How risky is this? It seems to suggest that we would not define anything for these cases. Why not just use a regular `dict` at this point, and just ensure the update goes through the `Attrdict.union` method?

brynpickering: Our pydantic model `update` method works on the basis that all entries in the pydantic model are themselves pydantic models. This works nicely for updating and validating at any point in the hierarchy. The `dict[str, type[BaseModel]]` entries breaks this and finding out in a generalised fashion whether we're currently looking at one of these interm dicts is hard.

irm-codebase: Hmm. Would `RootModel` help here?
https://docs.pydantic.dev/2.11/concepts/models/#rootmodel-and-custom-root-types

If not, let's stick to your proposed method. This is something I missed in my approach...

src/calliope/model.py
```diff
@@ -82,60 +82,54 @@ class Model:
 
     _TS_OFFSET = pd.Timedelta(1, unit="nanoseconds")
 
-    def __init__(self) -> None:
+    def __init__(
+        self, inputs: xr.Dataset, attrs: dict, results: xr.Dataset | None = None
+    ) -> None:
         """Returns an empty Model.
```
irm-codebase: No longer an empty model ;)

src/calliope/model.py
```diff
@@ -82,60 +82,54 @@ class Model:
...
-            "model_data_loaded",
-            comment="Model: loaded model_data",
+            LOGGER, self._attrs.timings, "model_creation", comment="Model: initialising"
         )
-        return model
+
+    def _check_versions(self) -> None:
+        """Check the initialised and defined calliope version."""
+        version_def = self._attrs.model_def.config.init.calliope_version
+        version_init = self._attrs.calliope_version_initialised
```
irm-codebase: Hopefully we can reduce nesting by flattening this into just `model.attrs` or `model.definition`?
Also, I see no need to do `_attrs` as pydantic already protects this from overwrites thanks to `frozen`.

src/calliope/model.py
```diff
@@ -147,68 +141,88 @@ def from_dict(
...
+
+    @property
+    def timings(self) -> dict[str, float]:
+        """Get timings of steps in the model."""
+        return self._attrs.timings
+
+    @property
+    def termination_condition(self) -> str | None:
+        """Get optimisation termination condition, which indicates whether the optimisation problem solved to optimality (`optimal`) or not (e.g. `unbounded`, `infeasible`)."""
+        return self._attrs.termination_condition
```
irm-codebase: Do we really need this?
The pydantic model is already protected. This seems like extra maintenance with little added benefit... 
How about just letting users consult these directly as `model.attrs.XXX`?

Otherwise... If combining model definition and attrs is causing us to add so many attributes, perhaps we could just keep them separate? Both can just be dumped, so it would not be a problem.

src/calliope/model.py
```diff
@@ -82,60 +82,54 @@ class Model:
...
+            )
+
+        if version_def is not None and not version_init.startswith(version_def):
+            exceptions.warn(
+                f"Model configuration specifies calliope version {version_def}, "
+                f"but you are running {version_init}. Proceed with caution!"
+            )
 
     @classmethod
     def from_dict(
```
irm-codebase: Great improvement! One last comment here...
The "external" methods  (`read_yaml`) are quite empty. It might be worth-wile considering moving all of this to those functions (as in, just having `read_yaml` and no `from_dict` method).

This is already fine, but that approach would be a bit neater, since you avoid having to validate two functions for each instantiation approach.

irm-codebase (commented):
Some quick comments for `model.py` @brynpickering 

My comments mostly relate to the new properties... I'd like to avoid these, if possible.
Pydantic already offers protection, and I do not see the benefit of hiding stuff from users. 

Instead, I think placing stuff in a way that is intuitive for both us and the user is the way to go. This way we maintain less functions and the code gets easier to understand.

As for `inputs` and `results`: feel free to go ahead with that approach! We'll likely just need a way to tell calliope to move an array from one to the other, which is not the biggest deal.

src/calliope/model.py
irm-codebase: Going with the 'separation' approach: we could make `model.definition` hold stuff that users have provided, or that is added because of their configuration (data, math, config, etc).

`model.attrs` could be for internal stuff we want to keep track of (termination conditions, flags like `_is_solved`, counters such as `_start_window_idx`, timing logs).

src/calliope/model.py
brynpickering: It was based on how e.g. pandas and xarray do it. They have a `read...` function and a `from...` function. It removes the layer of inference we have to make w.r.t. the user inputs (as in, we don't have to infer whether it is a dict or a YAML file that a user is defining). Agree though that it leads to (almost) duplicate docstrings which isn't ideal.

src/calliope/model.py
brynpickering: I would then expect both the contents of `model.definition` and `model.attrs` to be saved to file _and_ to be needed if the model is re-initialised, which is a bit onerous...

irm-codebase: True, although not that hard (just another value in `model.ATTRS_SAVED`, I think?).

What worries me here are two things:

- The sheer length of our calls (e.g., `self._attrs.model_def.config.init.calliope_version`). This is going to result in more lines of code due to formatter wrapping, the creation of more 'wrapper' like properties... and is a bit of an eye sore (personal preference, but still 🫤)
- Code safety:  separating by design stuff that has been defined externally (and which we should respect) and internal values (which we might mess around with) is useful in avoiding future bugs and other dangers. We have two boxes with different purposes, rather than one big shelf with sections that 'might' be stable and others that may  not.

irm-codebase: I'm guilty of this too (the new `math` dicts could've been placed outside of `_def`, for example...)

I just do not want this to become a perennial practice in the code. Easy to read code is easier to maintain in the long run!

src/calliope/backend/backend_model.py
brynpickering: This is for additions to the math which are made directly on the backend object (e.g. `model.backend.add_constraint`). There is no way to pass that back up to `m.applied_math`. The idea here is to store it in the backend dataset and then use it to update `m.applied_math` when the results are passed back.

src/calliope/schemas/general.py
brynpickering: I've updated to using RootModel

src/calliope/schemas/attributes.py
brynpickering: This is not about whether a user has defined it or not, it's a flag that stops operate mode runs when time clustering has been used.

src/calliope/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 from rich import pretty
```
irm-codebase: Why do we need `rich`? I missed this new dependency in my initial review.

src/calliope/model.py
```diff
@@ -104,20 +114,29 @@ def __init__(
...
+        self.definition = model_def_schema.CalliopeModelDef(
+            **kwargs.get("definition", {})
+        )
+        self.config = config_schema.CalliopeConfig(**kwargs.get("config", {}))
+        self.math = math_schema.CalliopeMath(**kwargs.get("math", {}))
+        self.runtime = runtime_attrs_schema.CalliopeRuntime(**kwargs.get("runtime", {}))
+
         self._start_window_idx: int = 0
         self._is_built: bool = False
         self._is_solved: bool = False if results is None else True
```
irm-codebase: Shouldn't these be in `runtime` too? That'd protect them, and remove the need to have a `@property` for them.

src/calliope/preprocess/data_tables.py
```diff
@@ -292,7 +292,7 @@ def _df_to_ds(self, df: pd.DataFrame) -> xr.Dataset:
         else:
             tdf = df
 
-        if self.input["select"] is not None:
+        if self.input.get("select", None) is not None:
             selector = tuple(
```
irm-codebase: Up for discussion, but I think we should trust that the schema will provide the right value, and not impute `None`. The cause seems to be that the `ModelDataFactory` forces the model definition to be an `AttrDict`, so we lose context all the day down here...

If there is something breaking the use of the schema here (maybe due to `defaults`?) it might be justified, for now.

src/calliope/preprocess/data_tables.py
```diff
@@ -303,10 +303,10 @@ def _df_to_ds(self, df: pd.DataFrame) -> xr.Dataset:
             )
             tdf = tdf.loc[selector]
 
-        if self.input["drop"] is not None:
+        if self.input.get("drop", None) is not None:
             tdf = tdf.droplevel(self.input["drop"])
 
-        if self.input["add_dims"] is not None:
+        if self.input.get("add_dims", None) is not None:
             for dim_name, index_items in self.input["add_dims"].items():
```
irm-codebase: Ditto.

src/calliope/preprocess/model_definition.py
```diff
@@ -40,20 +58,26 @@ def prepare_model_definition(
...
+
+    runtime = {
+        "applied_overrides": applied_overrides,
+        "scenario": scenario,
+        "defaults": extract_from_schema(MODEL_SCHEMA, "default"),
+    }
+
+    return CalliopeInputs(
+        config=config, definition=definition, math={"init": math}, runtime=runtime
+    )
```
irm-codebase: I'm seeing `mypy` throw several warnings here:

image

There is some kind of mismatch in this new approach...

src/calliope/model.py
```diff
@@ -104,20 +114,29 @@ def __init__(
...
         self.results = xr.Dataset() if results is None else results
         self.backend: BackendModel
-        self._attrs = CalliopeAttrs(**attrs)
+        self.definition = model_def_schema.CalliopeModelDef(
+            **kwargs.get("definition", {})
+        )
+        self.config = config_schema.CalliopeConfig(**kwargs.get("config", {}))
+        self.math = math_schema.CalliopeMath(**kwargs.get("math", {}))
+        self.runtime = runtime_attrs_schema.CalliopeRuntime(**kwargs.get("runtime", {}))
+
```
irm-codebase: I am seeing several `mypy` mismatches here. Something in the type chain may be broken?

image

src/calliope/preprocess/model_definition.py
```diff
@@ -9,19 +9,37 @@
...
 
+class CalliopeInputs(general.CalliopeBaseModel):
+    """All Calliope attributes."""
+
+    definition: model_def_schema.CalliopeModelDef = model_def_schema.CalliopeModelDef()
+    config: config_schema.CalliopeConfig = config_schema.CalliopeConfig()
+    math: math_schema.CalliopeMath = math_schema.CalliopeMath()
+    runtime: runtime_attrs_schema.CalliopeRuntime = (
+        runtime_attrs_schema.CalliopeRuntime()
+    )
```
irm-codebase: This is a bit out of place. To avoid such a heavy use of `imports` (all of the `schemas` are only used here), let's place this in the `schemas` package and import only `CalliopeInputs`?

Alternatively, a named tuple that requires `prepare_model_definition` to pass individually validated entries would work too (and avoid the need for a new class).

src/calliope/preprocess/model_definition.py
irm-codebase: Seems like `mypy` does not appreciate the use of `foo=bar` here, for some reason...
Passing a dictionary with ** works though:

image

src/calliope/preprocess/model_math.py
```diff
@@ -81,7 +81,7 @@ def build_applied_math(
         ModelError: a given name was not found in the math dictionary.
 
     Returns:
-        AttrDict: constructed and validated math dataset.
+        MathSchema: constructed and validated math dataset.
     """
```
irm-codebase: This still returns an `AttrDict`

src/calliope/schemas/dimension_data_schema.py
```diff
@@ -105,3 +108,21 @@ def check_dependent_definitions(self) -> Self:
                 f"Invalid latitude/longitude definition. Types must match, found ({self.latitude}, {self.longitude})."
             )
         return self
+
+
+class CalliopeParams(CalliopeDictModel):
+    """Calliope Parameters dictionary."""
+
+    root: dict[AttrStr, DataValue | IndexedParam] = Field(default_factory=dict)
+
```
irm-codebase, line 117: Nice find! This should take care of our problems re-validation issues, then?

src/calliope/schemas/math_schema.py
```diff
@@ -63,13 +82,13 @@ class PiecewiseConstraint(MathIndexedComponent):
...
+    x_expression: str = ""
     """X variable name whose values are assigned at each breakpoint."""
-    y_expression: str
+    y_expression: str = ""
     """Y variable name whose values are assigned at each breakpoint."""
-    x_values: str
+    x_values: str = ""
     """X parameter name containing data, indexed over the `breakpoints` dimension."""
-    y_values: str
+    y_values: str = ""
```
irm-codebase: Is it optional to set any of these? I initially did not specify a default to force users to define all of them.

irm-codebase (changes requested):
Here is the review. Things look good in general, with a few exceptions (see comments).

Before merging this in, we need to be careful in evaluating the implications of the `inputs`/`results` separation, though.

src/calliope/__init__.py
brynpickering: `rich` works well with pydantic (and xarray it turns out). In particular, to reign in the model information depth that is printed in an interactive session. I was actually thinking of making it an optional dependency and trigger the pretty install if the dependency exists in the environment.
 
Without `rich`:

image

With `rich`:

image

src/calliope/preprocess/data_tables.py
brynpickering: Yeah, I started converting all of this to expect a pydantic model but it really exploded the PR 😅 

So I would leave it to another PR.

src/calliope/preprocess/model_definition.py
brynpickering: This proved to involve the _least_ amount of repetition. Named tuples requires explicit reference to each of the input attrs in `calliope.Model.from_dict` and then dumping each of them back to a dict with explicit names equivalent to dumping `CalliopeInputs`. I'll move this to `schemas` instead.

src/calliope/schemas/dimension_data_schema.py
brynpickering: It does so long as it is a dictionary of `pydantic.BaseModel` type. The `init` math is a nested dictionary which I couldn't work out a nice way to extend the rootmodel into the furthest reaches of that nesting.

src/calliope/schemas/math_schema.py
brynpickering: Sorry, yeah, I'll revert. This was from a test in which I made the init math a pydantic model from the start (instead of an attrdict) but that requires all attrs to be optional. It exposed us to too many opportunities for the final math to be malformed though, so I reverted that.

src/calliope/model.py
brynpickering: I tried this and I think the answer is no, they shouldn't be. We don't want to save the state of these properties to file as they should be set on each init. If you have it in `runtime` and then build+solve your model and do a netcdf roundtrip then "is_built" and "is_solved" will both be `True`. This might be fine for `is_solved` in this case but not for `is_built` as it is no longer built. Since we'd have to check and set these properties at init every time anyway I would keep them as private properties with public facing `@property` methods.

requirements/dev.txt
```diff
@@ -14,3 +14,4 @@ pytest >= 8
 pytest-cov >= 4
 pytest-order >= 1
 pytest-xdist >= 3 # pytest distributed testing plugin
+rich >= 14.0.0
```
irm-codebase, line 17: I guess we are assuming only devs will play with prints?
Either here or `base.txt` is fine (unless it adds too many dependencies).

irm-codebase (approved):

requirements/dev.txt
brynpickering: I think it's nice to have pretty printed functionality for users who might be scripting their model runs but it isn't a must-have for calliope to work so I'm OK with making it optional for users to avoid excessive dependencies.

PR #774, opened 2025-06-25
Update postprocessing funcs

### LINKED ISSUE #762: Capacity factors in the results do not account for time resampling
### What can be improved?

It would make sense for `capacity_factor` in the results to account for time resampling. Right now, if resampling to, e.g., 3h, CF values in the results are summed over the 3-hour steps, becoming higher than 1. 

### Version

v0.7.0dev6

### LINKED ISSUE #767: LCOE Based on met demand rather than energy produced
### What can be improved?

While implementing curtailment and export markets, we noticed that LCOE is calculated by by the formula:

LCOE = (cost of the system) / ( sum of flow_out for supply and conversion)

This means when we curtail energy, or deliver it to a market outside our system, that energy is still included in the denominator of the LCOE for the system.

I'm suggesting the option to have LCOE calculated from the sum of flow_in to demand techs in the system. This way flow_export can remove energy while still adding the associated costs or revenue. LCOE will then be representative of average cost of energy delivered to consumers inside the system.

I implemented this into the postprocess.py by redefining the postprocess_model_results and systemwide_levelised_cost to add an extra variable demand_levelised_cost.

[Modified levelised Cost.txt](https://github.com/user-attachments/files/20634600/Modified.levelised.Cost.txt)

### Version

V0.7.0dev6

brynpickering:
Fixes #767 
Fixes #762 

* Account for time resolution in timeseries capacity factor calc
* Account for export flow in levelised cost calc
* Add some docs on the postprocessing methods we use to generate the new datasets a user will see

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/docs/reference/api/postprocess.md
+++ b/docs/reference/api/postprocess.md
@@ -1,0 +1,10 @@
+---
+search:
+    boost: 0.5
+---
+
+This file contains the methods used to generate postprocessed data tables that are available by default in the model results.
+
+::: calliope.postprocess.postprocess.capacity_factor
+
+::: calliope.postprocess.postprocess.systemwide_levelised_cost
--- a/src/calliope/postprocess/postprocess.py
+++ b/src/calliope/postprocess/postprocess.py
@@ -111,27 +111,22 @@ def systemwide_levelised_cost(
     """
     # Here we scale production by timestep weight
     cost = results["cost"].sum(dim="nodes", min_count=1)
-    flow_out = (
+    generation = (
         (results["flow_out"] + results.get("flow_export", xr.DataArray(0)).fillna(0))
         * model_data.timestep_weights
     ).sum(dim=["timesteps", "nodes"], min_count=1)
 
     if total:
-        # cost is the total cost of the system
-        # flow_out is only the flow_out of supply and conversion technologies
+        # `cost` is the total cost of the system
+        # `generation`` is only the generation of supply and conversion technologies
         allowed_techs = ("supply", "conversion")
         valid_techs = model_data.base_tech.isin(allowed_techs)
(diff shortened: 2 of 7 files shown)


docs/reference/api/postprocess.md
```diff
@@ -0,0 +1,10 @@
+---
+search:
+    boost: 0.5
+---
```
irm-codebase, line 4: Can you clarify the need for this boost?

src/calliope/postprocess/postprocess.py
```diff
@@ -110,11 +110,11 @@ def systemwide_levelised_cost(
...
     # Here we scale production by timestep weight
-    flow_out = results["flow_out"] * model_data.timestep_weights
     cost = results["cost"].sum(dim="nodes", min_count=1)
-    flow_out = (results["flow_out"] * model_data.timestep_weights).sum(
-        dim=["timesteps", "nodes"], min_count=1
-    )
+    flow_out = (
+        (results["flow_out"] + results.get("flow_export", xr.DataArray(0)).fillna(0))
+        * model_data.timestep_weights
+    ).sum(dim=["timesteps", "nodes"], min_count=1)
```
irm-codebase: Using `flow_out =` is a tad confusing here... how about `total_production` or `total_out(put)`?

irm-codebase (dismissed):
seems good to me!

docs/reference/api/postprocess.md
brynpickering: It stops the API being too high up in the search results (<1 == deprioritised) since it can clutter the search results with long strings.

PR #778, opened 2025-07-23
Add `group_datetime` and `sum_next_n` helper functions

### LINKED ISSUE #777: Reproducing a rolling sum helper function with predefined helper functions in v0.7.0dev6
Hi all,

Testing out the latest v0.7.0 release (dev6), Im trying to reproduce some math that I used to implement via Python before. Im looking for some feedback on how to implement this using the new math interface.

Im trying to constraint the startup/shutdown of assets to simulate a minimum startup/shutdown period. Using the operating_units variable that keeps track of the integer dispatch of units, I added two more strictly positive integer decision variables "startup_units" and "shutdown_units" to keep track when the state of the units changes.

By constraining the number of state changes over a rolling time period, I can enforce the units to stay off or on over this window. For this, I need a rolling sum functionality, something like (e.g. 24 hour time window):

`startups_in_window = sum(roll(startup_units,timesteps=1:24),over=timesteps)`
`shutdowns_in_window = sum(roll(shutdown_units,timesteps=1:24),over=timesteps)`

Such that ultimately I can constraint these state changes with the total number of units:
`$startups_in_window  + $shutdowns_in_window  <= purchased_units`

Meaning that each unit can undergo only one state change within the time window. 

I struggle to reproduce this with available the helper functions. Is this currently possible with the existing helper functions, or would I need to define my own helper function? 

Thanks in advance!
Floris

brynpickering:
Fixes #777

I had implemented these ad-hoc for other modelling efforts but since @fvandebeek brought it up in an issue I decided to formalise it in the core code.

The helper func names and docstrings might need more work to add clarity. I'd welcome ideas!

* `group_datetime` to sum over a time dimension in a math component using pre-defined datetime periods (date, hour, weekofyear, month, etc.)
* `sum_next_n` to apply a rolling summation over a time/ordered dimension of a math component.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/backend/helper_functions.py
+++ b/src/calliope/backend/helper_functions.py
@@ -13,6 +13,7 @@ from collections.abc import Mapping
 from typing import TYPE_CHECKING, Any, Literal, overload
 
 import numpy as np
+import pandas as pd
 import xarray as xr
 
 from calliope.exceptions import BackendError
@@ -105,7 +106,11 @@ class ParsingHelperFunction(ABC):
             _registry[allowed][cls.NAME] = cls
 
     @staticmethod
-    def _add_to_iterator(instring: str, iterator_converter: dict[str, str]) -> str:
+    def _update_iterator(
+        instring: str,
+        iterator_converter: dict[str, str],
+        method: Literal["add", "replace"],
+    ) -> str:
         r"""Utility function for generating latex strings in multiple helper functions.
 
         Find an iterator in the iterator substring of the component string
@@ -117,8 +122,8 @@ class ParsingHelperFunction(ABC):
             instring (str): String in which the iterator substring can be found.
             iterator_converter (dict[str, str]):
                 key: the iterator to search for.
-                val: The new string to **append** to the iterator name.
-
+                val: The new string to **append** to the iterator name (if method = add) or **replace**.
+            method (Literal[add, replace]): Whether to add to the iterator or replace it entirely
         Returns:
             str: `instring`, but with `iterator` replaced with `iterator + new_string`
         """
@@ -128,7 +133,11 @@ class ParsingHelperFunction(ABC):
             new_iterator_list = []
             for it in iterator_list:
                 if it in iterator_converter:
-                    it += iterator_converter[it]
+                    it = (
+                        it + iterator_converter[it]
+                        if method == "add"
+                        else iterator_converter[it]
+                    )
                 new_iterator_list.append(it)
 
             return matched.group(1) + ",".join(new_iterator_list) + matched.group(3)
@@ -415,7 +424,7 @@ class SelectFromLookupArrays(ParsingHelperFunction):
             (iterator := dim.removesuffix("s")): rf"={array}[{iterator}]"
             for dim, array in lookup_arrays.items()
         }
-        array = self._add_to_iterator(array, new_strings)
+        array = self._update_iterator(array, new_strings, "add")
         return array
 
     def as_array(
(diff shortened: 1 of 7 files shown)


src/calliope/backend/helper_functions.py
```diff
@@ -105,7 +106,11 @@ def __init_subclass__(cls):
             _registry[allowed][cls.NAME] = cls
 
     @staticmethod
```
irm-codebase, line 108: Optimisation:
I'd consider keeping `_add_to_iterator` and creating a `_replace_iterator` alternative instead.
This way you avoid adding the `if` computation when solving the helper, and keep the code faster.

src/calliope/backend/helper_functions.py
```diff
@@ -707,3 +716,145 @@ def as_array(self, array: xr.DataArray, condition: xr.DataArray) -> xr.DataArray
...
+
+    def as_array(self, array: xr.DataArray, over: str, group: str) -> xr.DataArray:
+        """Sum an expression array over the given dimension(s).
+
+        Args:
+            array (xr.DataArray): expression array
+            over (str): dimension name over which to group
+            group (str): datetime grouper.
+                Any xarray/pandas datetime grouper options
+                datetime grouper options include 'date', 'day', 'dayofweek', 'dayofyear', 'hour', 'month', 'week', 'weekday', 'weekofyear', 'year'.
```
irm-codebase: Nitpick:
I'd only mention two or so. This kind of comment can become misleading over time, since it depends on how the `pandas` and `xarray` implementation evolves.

src/calliope/backend/helper_functions.py
```diff
@@ -707,3 +716,145 @@ def as_array(self, array: xr.DataArray, condition: xr.DataArray) -> xr.DataArray
...
+            a `sink_use_flexible` timeseries parameter which we will use in the DSR constraint:
+
+            ```yaml
+            constraints:
+                4hr_demand_side_response:
+                    foreach: ["nodes", "techs", "carriers", "timesteps"]
+                    where: "carrier_in AND sink_use_flexible AND timesteps<=get_val_at_index(timesteps=-24)"
+                    equations:
+                        - expression: sum_next_n(flow_in, timesteps, 4) == sum_next_n(sink_use_flexible, timesteps, 4)"
+            ```
```
irm-codebase: This seems wrong... it mentions the `sum_next_n`, but this is `group_datetime`.

src/calliope/backend/helper_functions.py
```diff
@@ -707,3 +716,145 @@ def as_array(self, array: xr.DataArray, condition: xr.DataArray) -> xr.DataArray
...
+        """
+        # We can't apply typical xarray rolling window functionality
+        grouped: dict[str | int, xr.DataArray] = {}
+        grouper = array[over].groupby(getattr(array[over].dt, group))
+        for group_name, group_items in grouper:
+            if group == "date":
+                group_name = str(group_name)
+            else:
+                group_name = int(group_name)
+            grouped[group_name] = array.sel(**{over: group_items}).sum(
```
irm-codebase: Optimisation:
Is it really necessary to check, or is there a way to avoid, this `if`/`else` check within the loop?
It seems to be related to the `over` / `group` combination. If the type can be inferred from those, this would lead to a speed up.

If it cannot be inferred, ignore the above.

src/calliope/backend/helper_functions.py
brynpickering: I think this is minor from an optimisation perspective; it only gets used for latex math which is already quite lightweight. The extra duplicated code doesn't seem worth that effort.

irm-codebase (approved):
Approved!

PR #782, opened 2025-07-24
Fix gurobi expr eval

### LINKED ISSUE #780: Gurobi backend doesn't convert variable objects in expressions on loading results
### What happened?

If defining a global expression array element as simply a decision variable, it won't be parsed correctly when loading the results as the call to resolve an expression object is different to a decision variable and we expect all to be expressions.

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Version

v0.7.0.dev6

### Relevant log output

```Shell

```

brynpickering:
Fixes #780

Also fixed our dummy test model used in spores runs so that it doesn't have a free heat transmission tech it can increase/decrease the capacity of without consequence. Ideally I'd remove this free heat transmission technology from all electricity-only versions of the dummy model, but that touched far too many places in the tests.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/tests/test_backend_gurobi.py
+++ b/tests/test_backend_gurobi.py
@@ -64,6 +64,13 @@ class TestNewBackend:
             .all()
         )
 
+    def test_from_global_expression_fails_on_str_type(self, simple_supply_gurobi):
+        """Gurobi backend fails if it finds a string in an expression array."""
+        with pytest.raises(
+            TypeError, match="Cannot convert Gurobi object of type <class 'str'>"
+        ):
+            simple_supply_gurobi.backend._from_gurobi_expr("foo")
+
     def test_new_build_get_constraint(self, simple_supply_gurobi):
         constr = simple_supply_gurobi.backend.get_constraint("system_balance")
         assert (


src/calliope/backend/gurobi_backend_model.py
```diff
@@ -537,22 +537,22 @@ def _from_gurobi_var(val: gurobipy.Var) -> Any:
...
             Any: If the input is nullable, return np.nan, otherwise a numeric value
             (eval_body=True and problem is optimised) or a string.
         """
-        return val.getValue()
+        if isinstance(val, gurobipy.LinExpr):
+            return val.getValue()
+        elif isinstance(val, gurobipy.Var):
+            return val.x  # type: ignore
+        else:
+            return val
```
irm-codebase: The `Returns:` docstring seems to be outdated (or perhaps I am missing something?).

Similarly, I am not sure why this shouldn't fail if an invalid thing is passed... silently returning the value seems like trouble.

tests/common/test_model/scenarios.yaml
```diff
@@ -380,7 +398,7 @@ overrides:
       solve.spores:
         number: 2
     parameters:
-      spores_slack: 0.1
+      spores_slack: 0.3
```
irm-codebase, line 402: not part of this PR?

tests/conftest.py
```diff
@@ -102,6 +102,16 @@ def simple_supply_and_supply_plus():
...
+@pytest.fixture(scope="session")
+def simple_supply_spores_ready():
+    m = build_model(
+        {}, "var_costs,simple_supply_spores_ready,two_hours,investment_costs"
+    )
+    m.build()
+    m.solve()
+    return m
+
+
```
irm-codebase, line 114: Same as above?
If these fix another problem, please mention it in the PR. I'm not sure what this is doing here...

irm-codebase (changes requested):
@brynpickering small comment about a docstring.

However... wouldn't it be safer and simpler to just fail if the expression is a variable?
The reason gurobi does not convert those to expressions if probably because it counts as improper usage. You basically have an expression that does nothing (i.e., an user is putting things where they shouldn't).

brynpickering:
> However... wouldn't it be safer and simpler to just fail if the expression is a variable?
The reason gurobi does not convert those to expressions if probably because it counts as improper usage. You basically have an expression that does nothing (i.e., an user is putting things where they shouldn't).

It would be added memory and time overhead to always force an array entry in a global expression to an expression object. E.g., parameters are stored as plain floats in gurobi (it doesn't have a mutable parameter container like pyomo does). If an expression ends up being a single value then it's lots of extra gurobi objects if we wrap those.

As to having plain variables/params in an expression array. It is quite common that you get them. E.g., if you imagine a conditional that e.g. sets `flow_cap * param` for some techs but just `flow_cap` for others, for instance. I realise you could also do `flow_cap * 1` to create an expression, but that is less intuitive to write and to then review later.

tests/common/test_model/scenarios.yaml
brynpickering: Sorry, have updated in changelog and PR message now.

irm-codebase (approved):

PR #783, opened 2025-07-25
Test param dim from init

brynpickering:
This is a PoC of how we might use param/dim/lookup definitions in the math config and use that information to prepare model data on instatiation and to handle setup of the optimisation problem.

## TODO

- [ ] Add lookups into latex math
- [ ] Update docs examples and overview with changes
- [x] merge in #796 and fix YAMLs
- [x] Work out how to handle `time_cluster` in a vaguely elegant way (unlike subsetting and resampling, haven't yet found a way to generalise this).
- [x] Update spores math to define the spores parameters / lookups
- [x] Update backend parsing tests to use pydantic models rather than dicts
- [x] Add a `get_lookup` method on the backend? It just passes back the input array, but it would fit with the others
- [ ] actually validate against "one_of" in lookups. This is just passed through at the moment.
- [ ] split modeldatafactory into a module for transforming a dictionary to an xarray dataset and one for cleaning up a dataset. Only resolved math paths in the former but build math from priority list in both.
- [ ] decide on how to limit init kwargs that can be applied at init when re-instantiating with arrays. E.g. math paths can't be updated since they will already be resolved but extra math list can be. Clustering param could be defined if it wasn't defined before but you can't re-cluster/unclusrer an already clustered model. 

## Thoughts / open questions while working on this

1. When using `calliope.Model.backend.add_<component>(...)`, it will pass over the component if the parameters it has referenced aren't also already in the math. E.g., adding a new piecewise constraint invokes a few new parameters. These parameters weren't yet defined in the math so the piecewise constraint simply didn't get added (as the `where` string resolved to False on not finding the parameters)
2. we are storing the per-component definition in two places in the backend: in the math pydantic model and as attributes of the dataset arrays. Attributes help us grab the information when parsing but there's always going to be a syncing question here.
3. I'm now requiring the parameter/lookup definition to be included when adding each to the backend dataset. Since parameters don't actually need anything to be explicitly defined (it can take the pydantic model defaults), perhaps this is optional?
4. We should distinguish between array types in parsing. At the moment a model array is a generic component but if we parse dimensions, params, lookups, and variables/global expressions separately then we can control which can be defined where. E.g. slices can only accept lookups, value comparison in the where string can only accept parameters/lookups/dims, and lookups maybe can't make their way into the math expressions unless they are inside a helper function.

## Summary of changes

* You define your math options and your chosen math all at `init`. This would look different if/when we move math into the main model definition structure under a `math` heading because then you would have scenarios/overrides with changes to the math. Either way. Math choices are fixed at `init`.
* The param/lookup/dim defs provided in the math config are used at init when preparing the model dataset. They define data types and, for dims, they put some constraints on what can/can't be subset/resampled.
* Any dimension can feasibly be subset/resampled. Subsetting is done with from->to slices for datetime/ordered (numeric) data. It is done by selecting the defined elements for unordered dims. Resampling is only possible with datetime dims, but could feasibly extend to ordered numeric ones, too.
* Subsetting/resampling is defined at init. I thought about defining it in the math under each dim definition but this nests it quite deeply and it is something that users are likely going to want to do on-the-fly on instantiating a model so it seems wrong to store it so deeply.
* There are `parameters` (purely numeric) and `lookups` (any dtype) sections. These are separated in the backend, too although there isn't yet the logic to keep lookups only available in `where` strings. I thought about calling `lookups` `filters` but then referencing them in the source code can be annoying (as `filter` is a core python method)

## How it works

Some examples here. It's still buggy but these kinds of simple things seem to work.

### Initial `init`

```python
import calliope

# resolve YAML & overrides -> build dataset & math -> send to `calliope.Model` -> clean dataset with "built" math
m1 = calliope.examples.national_scale()

# pass model input tables as either parameters (converted to backend objs) or lookups (kept as-is) to the backend dataset -> build math components, filling in defaults where necessary.
# Technically, a math dict could be injected here. I haven't yet looked at limiting that to e.g. just changes to constraints
m1.build()

m1.solve()
```

### re-`init`

1. with a different time sampling
```python
m2 = calliope.Model(inputs=m1.inputs, attrs=m1.dump_model_attrs(), resample={"timesteps": "2h"})
```
5. with subsetting
```python
m2 = calliope.Model(inputs=m1.inputs, attrs=m1.dump_model_attrs(), subset={"timesteps": ["2005-01-01", "2005-01-01"], "nodes": ["region1, "region1_1"]})
```
6. with a new math mode
```python
m2 = calliope.Model(inputs=m1.inputs, attrs=m1.dump_model_attrs(), mode="operate")
```
This is less simple as we don't have any info about capacities to use as parameter values, so it will set them all to `inf` which will create an infeasible model due to setting e.g. `source_use == source_use_equals * flow_cap` when flow_cap is `inf`. So, one can do:
```python
import xarray as xr
m2 = calliope.Model(inputs=xr.merge([m1.inputs, m1.results]), attrs=m1.dump_all_attrs(), mode="operate")
```
Or, be more specific:
```python
m2 = calliope.Model(inputs=m1.inputs.assign(flow_cap=m1.results.flow_cap), attrs=m1.dump_all_attrs(), mode="operate")
```
Or with SPORES:
```python
m2 = calliope.Model(inputs=m1.inputs.assign(spores_slack=0.2, spores_baseline_cost=m1.results.min_cost_optimisation), attrs=m1.dump_all_attrs(), mode="spores")
```
7. with a new math mode _and_ subsetting
E.g., to run operate mode over just a few weeks:
```python
m2 = calliope.Model(inputs=xr.merge([m1.inputs, m1.results]), attrs=m1.dump_all_attrs(), mode="operate", subset={"timesteps": ["2005-01", "2005-01"]})
```


src/calliope/math/operate.yaml
irm-codebase, line 1: Nice! We should also deactivate `flow_cap` in the variables here.
As far as I understand, this should be possible with our current code.

src/calliope/math/plan.yaml
```diff
@@ -1,3 +1,644 @@
...
+    description: If True, non-zero `flow_out` and `flow_in` cannot both occur in the same timestep.
+
+  cyclic_storage:
+    type: float
+    default: 1.
+    title: Cyclic storage switch.
+    description: >
+      If true, link storage levels in the last model timestep with the first model timestep.
+      `inter_cluster_storage` custom math must be included if using time clustering and setting this to `true`.
+      This must be set to `false` if using `operate` mode.
```
irm-codebase: Should we restrict `lookups` to be either strings or booleans in the schema?
I'm noticing mixed usage: `definition_matrix` is bool, `integer_dispatch` and this are floats, but they do the same thing...

If `.nan` is supported for bools too, this restriction would help in writing consistent math defintions.

src/calliope/preprocess/model_data.py
```diff
@@ -73,56 +75,69 @@ class ModelDataFactory:
...
             data_table_dfs: (dict[str, pd.DataFrame], None): Dataframes with model data. Defaults to None.
             attributes (dict): Attributes to attach to the model Dataset.
-            param_attributes (dict[str, dict]): Attributes to attach to the generated model DataArrays.
         """
         self.config: Init = init_config
-        self.model_definition: ModelDefinition = model_definition.copy()
-        self.dataset = xr.Dataset()
+        if isinstance(model_definition, dict):
+            self.model_definition: ModelDefinition = model_definition.copy()
+            self.dataset = xr.Dataset()
```
irm-codebase: We should strive to eliminate the use of `copy` as much as possible.

Given that we are now passing inputs and other settings between model objects, the chances of 'exploding' the memory requirements increase. We can eliminate these cases as long as we ensure these objects (`model_definition`, `inputs`, `math`, etc) are not altered unless explicitly requested to do so (i.e., resampling).

src/calliope/preprocess/model_data.py
```diff
@@ -73,56 +75,69 @@ class ModelDataFactory:
...
-        self.init_from_data_tables(tables)
-
-        flipped_attributes: dict[str, dict] = dict()
-        for key, val in param_attributes.items():
-            for subkey, subval in val.items():
-                flipped_attributes.setdefault(subkey, {})
-                flipped_attributes[subkey][key] = subval
-        self.param_attrs = flipped_attributes
+        if tables:
+            self.init_from_data_tables(tables)
```
irm-codebase: I assume this is never reached when initialising a second model using the results of a previous one, correct?

src/calliope/preprocess/model_data.py
```diff
@@ -251,20 +268,18 @@ def add_top_level_params(self):
...
-        time_subset = self.config.time_subset
-        if time_subset is not None:
-            self.dataset = time.subset_timeseries(self.dataset, time_subset)
+        self._subset_dims()
+        self._resample_dims()
+
         self.dataset = time.add_inferred_time_params(self.dataset)
 
         # By default, the model allows operate mode
         self.dataset.attrs["allow_operate_mode"] = 1
```
irm-codebase: Would it make sense to move `allow_operate_mode` to the math definition files as a global lookup?
`True` in `base.yaml` and only set to `False` in the clustered file?

src/calliope/preprocess/model_data.py
```diff
@@ -601,9 +621,9 @@ def _deactivate_item(self, **item_ref):
             self.dataset = self.dataset.drop_sel(**item_ref)
         else:
             if "carrier_in" in self.dataset:
-                self.dataset["carrier_in"].loc[item_ref] = np.nan
+                self.dataset["carrier_in"].loc[item_ref] = False
             if "carrier_out" in self.dataset:
-                self.dataset["carrier_out"].loc[item_ref] = np.nan
+                self.dataset["carrier_out"].loc[item_ref] = False
```
irm-codebase, line 623: I am a bit puzzled by this move to `False`... could you elaborate?

src/calliope/preprocess/model_data.py
```diff
@@ -737,16 +777,63 @@ def _update_numeric_dims(ds: xr.Dataset, id_: str) -> xr.Dataset:
...
+
+        for dim_name, subset in self.config.subset.root.items():
+            if subset is None or dim_name not in self.dataset.coords:
                 continue
+            is_ordered = self.math.dims[dim_name].ordered
+            dim_vals = self.dataset.coords[dim_name]
 
-        return ds
+            if dim_vals.dtype.kind == "M":
+                time._check_time_subset(dim_vals.to_index(), subset)
```
irm-codebase: Please clarify where the magic value (`"M"`) comes from...

is this some kind of `pandas` standard? it should be a global constant in one of our files, for clarity / maintainability.

src/calliope/schemas/config_schema.py
```diff
@@ -19,6 +25,20 @@
 ]
 
 
+class Subsets(CalliopeDictModel):
+    """Dimension subsets."""
+
+    root: dict[AttrStr, NonEmptyList[str | int | float] | None] = Field(
+        default_factory=dict
+    )
+
```
irm-codebase, line 34: Not a fan of this one... it can be easily confused with the subset of any dimension (e.g., a subset of fossil technologies).

If this is datetime specific, the name should reflect it.

src/calliope/schemas/math_schema.py
```diff
@@ -40,6 +40,33 @@ class MathComponent(CalliopeBaseModel):
...
+class Lookup(MathComponent):
+    """Schema for named lookup arrays."""
+
+    default: str | float | int | bool | None = None
+    type: Literal["integer", "float", "string", "bool"]
+    resample_method: Literal["mean", "sum", "first"] = "first"
+    unit: str = ""
+    one_of: list | None = None
+    dim_to_bool: str | None = None
+
```
irm-codebase: only `str` and `bool` as types?

irm-codebase (commented):
@brynpickering I like this quite a lot!

Key comments:

This new approach requires more care in avoiding memory explosions. Even slight duplications could have unintended effects. Whenever possible, we should avoid / remove the use of any `deepcopy`, and enforce immutability of inputs (e.g., the backend does not `copy()` the math, and never alters it).

src/calliope/schemas/math_schema.py
```diff
@@ -40,6 +40,33 @@ class MathComponent(CalliopeBaseModel):
...
+    type: Literal["string", "datetime", "integer", "float"]
+    ordered: bool = False
+
+
+class Parameter(MathComponent):
+    """Schema for named parameter."""
+
+    default: float | int = float("nan")
+    type: Literal["integer", "float"]
```
irm-codebase: you probably want to change `type` to `dtype`, as `type` is an internal python name

src/calliope/math/plan.yaml
brynpickering: Good point. I'll go through and ensure all boolean lookups are actually boolean. But there are lookups that are used to map between dimension items (e.g. the clustering timeseries ones, which map from the first to the last timestep of a cluster). These could feasibly have integers/floats as the values, if the dimension they're mapping between is integer/float, so we should probably keep the option for int/float dtype.

src/calliope/preprocess/model_data.py
brynpickering: correct. Unless a user has an xarray data variable called `data_tables`. It's an edge case we should probably plan for!

src/calliope/preprocess/model_data.py
brynpickering: It could be a lookup with a true/false default. The issue is that you don't have to use inter-cluster storage with time clustering so there's no way to force the switch just from the choice of math. I agree though that there doesn't seem much point passing this around so much when it is kinda niche. We should just have a simpler check for a clustering dimension/param in the input data when using operate mode.

src/calliope/preprocess/model_data.py
brynpickering: These are now boolean arrays rather than fake boolean arrays (where np.nan was a proxy for False) so we need to handle them as such.

src/calliope/preprocess/model_data.py
brynpickering: "M" is a [numpy type kind](https://numpy.org/doc/stable/reference/generated/numpy.dtype.kind.html) that points to being a datetime. We could make it a util function for clarity.

src/calliope/schemas/config_schema.py
brynpickering: It's not datetime specific (see example 2 in the main comment where I subset on `nodes` as well as `timesteps`)

src/calliope/preprocess/model_data.py
irm-codebase: Nah, I think a global with a clear name is better in this case. Functions add computation time (you always need to 'jump' to a new level in the stack).
A global is clear, has no computational impact and removes the 'magic' from these values.

src/calliope/schemas/config_schema.py
irm-codebase: Got it. We might need to think how to ensure this is not confused with actual subsets if we ever add that to the math, though.

sjpfenninger: `DimensionSubsets`? And then maybe also `DimensionResamples` for consistency?

src/calliope/schemas/math_schema.py
```diff
@@ -40,6 +40,33 @@ class MathComponent(CalliopeBaseModel):
     """If False, this component will be ignored during the build phase."""
 
 
+class Dim(MathComponent):
```
sjpfenninger: `Dimension`, or is there a large gain from shortening it?

src/calliope/example_models/national_scale/model.yaml
```diff
@@ -12,18 +12,18 @@ config:
     name: National-scale example model
     # What version of Calliope this model is intended for
     calliope_version: 0.7.0
-    time_subset: ["2005-01-01", "2005-01-05"] # Subset of timesteps
     broadcast_param_data: true  # allow single indexed parameter data entries to be broadcast across all index items, if there are multiple entries.
+    mode: base # Choices: base, operate
```
sjpfenninger: There is `init.base_math: plan` as default in the config, so to get the "base" math for plan mode from `plan.yaml` you need to specify `mode: base`. This is quite confusing. Shouldn't this still be `mode: plan`? And simplify to remove the special checks for "base" in the code then.

Basically we have two concepts: the math and the mode. Ideally the link between the two is clear and transparent, with as little magic as possible. So if there is "plan.yaml" math the accompanying mode should be called "plan".

src/calliope/preprocess/model_math.py
```diff
@@ -93,5 +103,46 @@ def build_applied_math(
     if overwrite:
         LOGGER.info("Math build | appending additional math.")
         math.union(overwrite, allow_override=True)
+    math_model = MathSchema(**math)
+    if validate:
+        _validate_math_string_parsing(math_model)
+    return math_model
+
+
+def _math_priority(config: config_schema.Init) -> list[str]:
```
sjpfenninger: This code is repeated in `model_data.py`

src/calliope/example_models/urban_scale/model.yaml
```diff
@@ -12,15 +12,14 @@ config:
     # What version of Calliope this model is intended for
     calliope_version: 0.7.0
     # Time series data path - can either be a path relative to this file, or an absolute path
-    time_subset: ["2005-07-01", "2005-07-02"] # Subset of timesteps
     broadcast_param_data: true  # allow single indexed parameter data entries to be broadcast across all index items, if there are multiple entries.
     extra_math:
       additional_math: "additional_math.yaml"
+    apply_math: ["additional_math"]
```
sjpfenninger: Is there a need why users should want to separately specify `extra_math` and `apply_math`? Why not just apply the math specified in `extra_math` in the order it's given, simplifying this aspect of the config?

src/calliope/example_models/national_scale/model.yaml
sjpfenninger:
```suggestion
    mode: base # Choices: base, operate, spores
```

src/calliope/example_models/urban_scale/model.yaml
```diff
@@ -12,15 +12,14 @@ config:
...
 
   build:
     mode: base # Choices: base, operate
```
sjpfenninger:
```suggestion
    mode: base # Choices: base, operate, spores
```

sjpfenninger:
Just to record this here: from #784, the suggestion to implement here: dims define their iterator in their metadata - to avoids the hacky string operations to add "s"

src/calliope/schemas/config_schema.py
brynpickering: That's OK in the pydantic model but I don't think we want the user config to be `config.init.dimension_subset: {...}`. It's just a bit too verbose for something they will likely use most times they work with models.

src/calliope/math/operate.yaml
brynpickering: it is deactivated? line 63

irm-codebase: github hiding code bit my butt here.
Looks good!

src/calliope/preprocess/model_data.py
brynpickering: We're copying when it is provided as a dict, not when it's provided as a xarray dataset. I think this is keeping it as limited as possible in this case.

docs/user_defined_math/customise.md
```diff
@@ -77,55 +77,38 @@ config:
...
+    title: Maximum rated flow capacity.
+    description: >-
+        Limits `flow_cap` to a maximum.
+    unit: power.
+lookups:
+  source_unit:
+    default: absolute
+    title: Source unit
+    description: The unit of a given technologies source data
+    one_of: [absolute, per_cap, per_area]
```
irm-codebase, line 109: Nitpick: this is featured here for the first time. Depending on the order of the docs, users may not have the context of what these are.

docs/user_defined_math/customise.md
```diff
@@ -77,55 +77,38 @@ config:
 !!! danger
     Modes and other pre-defined options such as `operate` and `spores` might not work as expected!
 
-## Adding your parameters to the YAML schema
+## Adding your parameters math definition
```
irm-codebase: Reads a tad off... how about

```suggestion
## Adding your own parameters to the math definition
```

src/calliope/backend/backend_model.py
```diff
@@ -357,70 +357,77 @@ def _add_component(
             parsing.ParsedBackendComponent | None: parsed component. None if the break_early condition was met.
         """
         references: set[str] = set()
```
irm-codebase, line 359: This function seems to be exploding in nested complexity.
At some point we will likely need to separate this into a function per type of component instead.
It will increase maintainability and avoid bugs.

src/calliope/backend/backend_model.py
```diff
@@ -27,74 +27,141 @@
...
+
+    Args:
+        func (Callable): The function with a definition to validate.
+    """
+
+    def validator(*args, **kwargs):
+        sig = inspect.signature(func)
+        bound_args = sig.bind_partial(*args, **kwargs)
+        bound_args.apply_defaults()
+        component_def = bound_args.arguments["definition"]
```
irm-codebase: perhaps a brief `# comment` explaining what this is doing?

src/calliope/backend/backend_model.py
irm-codebase: I think moving some of these into the `add_objective` / `add_xxx` methods above might help. currently they are abstract methods, but moving to using `super` could help.
Just a suggestion, not a requirement for merging these changes.

src/calliope/backend/backend_model.py
```diff
@@ -397,33 +462,22 @@ def _create_obj_list(self, key: str, component_type: ALL_COMPONENTS_T) -> None:
...
-            self.log(
-                "parameters", param_name, "Component not defined; using default value."
-            )
-            self.add_parameter(param_name, xr.DataArray(np.nan), default_val)
-        LOGGER.info("Optimisation Model | parameters | Generated.")
+        for name, data in self.inputs.data_vars.items():
+            if data.obj_type == "parameters":
+                self.add_parameter(name, data, self.math.parameters[name])
+            elif data.obj_type == "lookups":
+                self.add_lookup(name, data, self.math.lookups[name])
```
irm-codebase, line 481: Nitpick:
I'd add an `else:` that raises an error for future proofing. It's usually better to cover all the cases of an `if`

src/calliope/backend/backend_model.py
```diff
@@ -555,6 +598,8 @@ def _raise_error_on_preexistence(self, key: str, obj_type: ALL_COMPONENTS_T):
                 (either with the same or different type as `obj_type`).
         """
         if key in self._dataset.keys():
+            if not self._dataset[key].attrs.get("active", True):
+                return None
             if key in getattr(self, obj_type):
                 raise BackendError(
                     f"Trying to add already existing `{key}` to backend model {obj_type}."
```
irm-codebase: This function is a bit inconsistent in its behaviour. `None` is always returned, regardless of whether or not it is active (all functions without a `return` default to `None`. Having the `return` for that case can be confusing, IMO.

src/calliope/backend/expression_parser.py
```diff
@@ -190,18 +196,23 @@ class EvalOperatorOperand(EvalToArrayStr):
...
+    def __init__(self, instring: str, loc: int, tokens: pp.ParseResults) -> None:
         """Process successfully parsed expressions with operands separated by an operator.
 
         I.e.: OPERAND OPERATOR OPERAND OPERATOR OPERAND ...
 
         Args:
+            instring (str): String that was parsed (used in error message).
+            loc (int):
+                Location in parsed string where parsing error was logged.
+                This is not used, but comes with `instring` when setting the parse action.
```
irm-codebase: Could you clarify why `loc` is present in so many of these?
As in, the line of code that forces it?

src/calliope/backend/gurobi_backend_model.py
```diff
@@ -126,32 +121,30 @@ def _expression_setter(
...
-            bounds = variable_dict["bounds"]
-            lb = self._get_variable_bound(bounds["min"], name, references)
-            ub = self._get_variable_bound(bounds["max"], name, references)
+            bounds = definition.bounds
+            lb = self._get_variable_bound(bounds.min, name, references)
+            ub = self._get_variable_bound(bounds.max, name, references)
             var = self._apply_func(
                 self._instance.addVar, where, 1, lb, ub, vtype=domain_type
             )
             return var.fillna(value=np.nan)
```
irm-codebase, line 130: I feel like a lot of the complexity in `_add_component` could be alleviated by moving generic things to a `super()` call to the parent `BackendModel` definition of these functions...

The logic of this seems inverted: going from a specific of `Gurobi` / `Pyomo` / etc to a generic (`_add_component`) that handles things for every backend and every type of component.

src/calliope/backend/where_parser.py
```diff
@@ -163,47 +161,41 @@ def _preprocess(self) -> str:
...
-    def _data_var_with_default(self, source_dataset: xr.Dataset) -> xr.DataArray:
+    def _data_var_with_default(self) -> xr.DataArray:
         """Access data var and fill with default values.
 
         Return default value as an array if var does not exist.
         """
-        default = self.eval_attrs["defaults"].get(self.data_var)
-        var = source_dataset.get(self.data_var, xr.DataArray(default))
+        var = self.eval_attrs["input_data"][self.data_var]
+        default = var.attrs.get("default", None)
```
irm-codebase: If default is set in the schema, we should not do this since it imposes a silent assumption.
It's much safer to just fail if this is not there, since it implies that something in our code broke.

src/calliope/backend/where_parser.py
```diff
@@ -225,15 +217,32 @@ def as_math_string(self) -> str:  # noqa: D102, override
...
+                    f"where string | Can only check for existence of values in {data_var_type.removesuffix('s')} arrays in math `where` strings. "
+                    "These arrays cannot be used for comparison with expected values. "
+                    f"Received `{self.data_var}`."
+                )
 
-        if self.eval_attrs.get("apply_where", True):
-            return self._data_var_exists(source_dataset, data_var_type)
-        else:
-            return self._data_var_with_default(source_dataset)
+            return self._data_var_with_default()
```
irm-codebase, line 249: Nitpick: it's generally safer to have one return in a function, as it helps in identifying gaps in its logic more easily, and makes debugging (e.g., breakpoints) easier to set.

src/calliope/model.py
```diff
@@ -87,49 +83,56 @@ class Model:
...
 
         self._start_window_idx: int = 0
         self._is_built: bool = False
         self._is_solved: bool = False if results is None else True
 
+        self.config = self.config.update({"init": kwargs})
         self._check_versions()
+
+        model_data_factory = ModelDataFactory(
+            self.config.init, inputs, self.math.init, None, None
```
irm-codebase: Correct me if I'm wrong... but isn't this being called twice when you call `Model.from_dict`? First in `from_dict`, then in `__init__`?

src/calliope/preprocess/model_data.py
```diff
@@ -380,10 +388,14 @@ def add_colors(self):
 
     def assign_input_attr(self):
         """Assign the the available parameter metadata as attributes to each input parameter array."""
+        all_attrs = {
+            **self.math.parameters.model_dump(),
+            **self.math.lookups.model_dump(),
+        }
```
irm-codebase, line 394: Possible bug: what would happen if a parameter and a lookup have the same name?
We might need to include a post-validation test in the schema that ensures all elements are differently named (even across params, lookups, dims, etc), for safety.

src/calliope/preprocess/model_math.py
```diff
@@ -93,5 +104,36 @@ def build_applied_math(
     if overwrite:
         LOGGER.info("Math build | appending additional math.")
         math.union(overwrite, allow_override=True)
+    math_model = CalliopeBuildMath(**math)
+    if validate:
+        _validate_math_string_parsing(math_model)
```
irm-codebase, line 103: Is there a scenario in which we would not want to validate?
If not, I think we can remove the extra input / if.

src/calliope/preprocess/time.py
```diff
@@ -103,18 +105,23 @@ def resample(data: xr.Dataset, resolution: str) -> xr.Dataset:
...
 
     for var_name, var_data in data_ts.data_vars.items():
         resampler = var_data.resample(**resample_kwargs)
+        resample_method = var_data.attrs.get("resample_method")
         if var_name == "timestep_resolution":
             method = "sum"
-        elif var_name in resampling_methods:
-            method = resampling_methods.get(var_name, None)
+        elif resample_method is not None:
+            method = resample_method
```
irm-codebase: Some suggestions.

Given that we now have schema defaults, we should be relying on them. Some recs:
- the `else` should not impute a value. It already is `first` in the schema, this code should trust it.
- instead of forcing the values, it might be better to just check if the value matches the expected case. For example:
   - `if var_data.dtype.kind in ["f", "i"] and not method == "mean"` -> raise error and explain why this is wrong

src/calliope/preprocess/time.py
```diff
@@ -103,18 +105,23 @@ def resample(data: xr.Dataset, resolution: str) -> xr.Dataset:
...
         elif var_data.dtype.kind in ["f", "i"]:
             method = "mean"
```
irm-codebase: Just to clarify:
Looking at the schema, these should be "sting", "datetime", "integer" or "float". Where to these "f" and "i" come from? `pandas`?

src/calliope/preprocess/time.py
irm-codebase: The `if len(time_subset) != 2:` error in `_check_time_subset` below is a nice example of this approach

src/calliope/schemas/math_schema.py
```diff
@@ -40,6 +40,47 @@ class MathComponent(CalliopeBaseModel):
...
+    iterator: str
+    """The name of the iterator to use in the LaTeX math formulation for this dimension."""
+
+
+class Parameter(MathComponent):
+    """Schema for named parameter."""
+
+    default: float | int = float("nan")
+    """The default value for the parameter, if not set in the data."""
+    """The parameter data type."""
```
irm-codebase: leftover string?

src/calliope/schemas/math_schema.py
```diff
@@ -40,6 +40,47 @@ class MathComponent(CalliopeBaseModel):
...
+    resample_method: Literal["mean", "sum", "first"] = "first"
+    """If resampling is applied over any of the parameter's dimensions, the method to use to aggregate the data."""
+    unit: str = ""
+    """The unit of the parameter, e.g. 'kW', 'm', 'kg', 'energy', 'power', ..."""
+
+
+class Lookup(MathComponent):
+    """Schema for named lookup arrays."""
+
+    default: str | float | int | bool = float("nan")
```
irm-codebase: would it make sense to use `AttrStr` instead of `str` here? If we want these to be single words, it would be safer to go with the former.

src/calliope/schemas/math_schema.py
```diff
@@ -40,6 +40,47 @@ class MathComponent(CalliopeBaseModel):
...
+    """The lookup data type."""
+    resample_method: Literal["mean", "sum", "first"] = "first"
+    """If resampling is applied over any of the lookup's dimensions, the method to use to aggregate the data."""
+    one_of: list | None = None
+    """If given, the lookup values must be one of these items."""
+    pivot_values_to_dim: str | None = None
+    """If given, the lookup will be pivoted such that its values become the index of a new dimension and its new values are boolean, True where the index values match the old values.
+    For instance, if the lookup starts out indexed over `techs` with values of `[electricity, gas]` and `pivot_values_to_dim: carriers`,
+    then the lookup will be converted to a boolean array with the dimensions ['techs', 'carriers'].
+    """
```
irm-codebase, line 98: I am struggling a bit difficult to understand the use case. Could you clarify?

src/calliope/schemas/math_schema.py
```diff
@@ -139,32 +193,78 @@ class Variable(MathIndexedComponent):
...
+
+    @property
+    def sub_expressions(self) -> SubExpressions:
+        """Dummy property to satisfy type hinting."""
+        return SubExpressions()
+
+    @property
+    def slices(self) -> SubExpressions:
+        """Dummy property to satisfy type hinting."""
+        return SubExpressions()
```
irm-codebase: I've never had issues with type hints with `pydantic`. Could you clarify which tool this aims to help?

docs/user_defined_math/customise.md
brynpickering: Would a longer description help?

src/calliope/backend/backend_model.py
brynpickering: Agree it's exploded. Not sure it helps to split it out by component type since many of the steps are the same irrespective of the component being added. Need to better handle the difference between latex/non-latex backend and when a component is set as it's default value or not

src/calliope/backend/backend_model.py
brynpickering: For info: it's turning args into kwargs so that we can unambiguously select "definition" from the user input.

src/calliope/backend/expression_parser.py
brynpickering: It's a pyparsing feature. These classes are pyparsing hooks that it handles, so we have to adhere to their expected init signature which can be _just_ "tokens" or all three args as you see here. We just don't ever use "loc" in our error messages

src/calliope/backend/gurobi_backend_model.py
brynpickering: That would be good, yeah

src/calliope/model.py
brynpickering: Yes, in init it is created to run the cleaning functions. It would be better to split this into two classes or modules though, especially since I've realised it is loading the math paths again in init which will obviously break if those paths break. It _should_ build the math from the priority list both in add_dict and init since it needs them in both sets of methods (creating the dataset from YAML in add_dict and cleaning up that dataset in init). The current approach is a serious hack...

src/calliope/preprocess/model_data.py
brynpickering: Yeah, as with all of these components, everything must be unique. We could/should do this on building the math from the priority list as a pydantic model validator as it will then make other steps in the process cleaner as we _know_ only one active instance of a named component exists

src/calliope/preprocess/model_math.py
brynpickering: This is done at init and it parses the math which can take ~10 seconds which is quite a wait at this point in the process if you already know your math is well formed. We should rename it to make it clearer what validation at this stage really means.

src/calliope/preprocess/time.py
brynpickering: Yep agreed. This is one of the files in which we should expect to receive the pydantic model directly rather than a dictionary so that we can be sure about what we have access to

src/calliope/preprocess/time.py
brynpickering: They are numpy type kinds but agreed that we can just check the pydantic model for the named dtype. The only problem is cases where the input has not been defined in the math (i.e. it's just being passed through). You want to know the dtype and then only way to know is to look at the array itself. Same issue with "resample_method". Not sure how to handle that

src/calliope/schemas/math_schema.py
brynpickering: We've made it easy to define carriers as a list of strings when actually our indexed parameter definition would require it to be a list of bools that are indexed over the carriers dimension. The simplest would be to just force a list of bools but it's quite cumbersome given that every tech needs carrier_in/_out defined. This allows certain lookups to be defined as a string with the expectation they are pivoted to be indexed over a dimension of which those strings are members. Ideas of a better config name for this are very welcome!

src/calliope/schemas/math_schema.py
brynpickering: In the math parsing we could receive any math component and we can choose between saying `component.sub_expressions` and receiving an empty dict in these cases or saying `getattr(component, "sub_expressions", {})` to handle receiving a math component that doesn't allow this.

docs/user_defined_math/customise.md
irm-codebase: It's just making sure that `lookups` have been introduced before users see this.

docs/advanced/time.md
```diff
@@ -9,20 +9,30 @@ This default resolution can be adjusted by specifying time resolution adjustment
...
 In the above example, this would resample all time series data to 6-hourly timesteps.
 Any [pandas-compatible rule describing the target resolution][pandas.DataFrame.resample] can be used.
 
+!!! note
+    Other dimensions can also be resampled, if they have a `datetime` type.
+
 ## Time clustering
 
-By supplying a file linking dates in your model timeseries with representative days, it is possible to cluster your timeseries:
+By loading a file linking dates in your model timeseries with representative days, it is possible to cluster your timeseries:
```
sjpfenninger, line 24:
```suggestion
By loading a file linking dates in your model timeseries with representative days, it is possible to cluster your timeseries.
To do so you need to load a parameter from file via `data_tables` and then refer to this parameter from `config.init.time_cluster`:
```

docs/examples/modes.py
```diff
@@ -44,11 +44,13 @@
 calliope.set_log_verbosity("INFO", include_solver_output=False)
 
 # %% [markdown]
-# ## Running in 'plan' mode.
+# ## Running in 'base' mode.
 
 # %%
 # We subset to the same time range as operate mode
-model_plan = calliope.examples.national_scale(time_subset=["2005-01-01", "2005-01-10"])
+model_plan = calliope.examples.national_scale(
```
sjpfenninger:
```suggestion
model_base = calliope.examples.national_scale(
```

docs/examples/modes.py
```diff
@@ -44,11 +44,13 @@
...
+    subset={"timesteps": ["2005-01-01", "2005-01-10"]}
+)
 model_plan.build()
```
sjpfenninger:
```suggestion
model_base.build()
```

docs/examples/modes.py
```diff
@@ -44,11 +44,13 @@
...
 model_plan.solve()
```
sjpfenninger:
```suggestion
model_base.solve()
```

src/calliope/schemas/config_schema.py
```diff
@@ -65,13 +91,32 @@ class Init(CalliopeBaseModel):
...
+    """Mode in which to run the optimisation.
+    Triggers additional processing and appends additional math formulations.
+    Math order: base -> mode
     """
 
-    extra_math: dict[AttrStr, str] = Field(default={})
-    "Dictionary with the names and paths of additional math files."
+    extra_math: UniqueList[str] = Field(default_factory=list)
+    """
+    List of math entries to be applied on top of the `base` math and `mode` math.
```
sjpfenninger, line 108:
```suggestion
    List of math entries to be applied on top of the `base` math and `mode` math.
    The list items must have been defined as keys in `math_paths` (see below).
```

src/calliope/schemas/math_schema.py
```diff
@@ -40,7 +41,52 @@ class MathComponent(CalliopeBaseModel):
     """If False, this component will be ignored during the build phase."""
 
 
-class MathIndexedComponent(MathComponent):
+class Dimension(MathComponent):
+    """Schema for named dimension."""
+
+    dtype: Literal["string", "datetime", "integer", "float"]
+    """The dimension items data type"""
```
sjpfenninger:
```suggestion
    """The data type of this dimension's items."""
```

docs/user_defined_math/customise.md
brynpickering: This will need to be fixed by general docs fixes for lookups

irm-codebase (approved):

sjpfenninger (approved):

PR #698, opened 2024-10-18
Add `where` helper function to enable reproduction of fancier group constraints

brynpickering:
Fixes #604 
Fixes #679

We use `where` a lot now. So, if there is another name we can give this helper function, I'm all ears!

## Reviewer checklist

- [ ] Test(s) added to cover contribution
- [ ] Documentation updated
- [ ] Changelog updated
- [ ] Coverage maintained or improved

PR #784, opened 2025-07-27
Add `group_sum` helper function (potentially supersedes `group_datetime`)

### LINKED ISSUE #735: Memory error due to huge array introduced by `default_if_empty` in custom math
### What happened?

I am using custom math to define group constraints on groups of links. I am using the helper function `default_if_empty` as an indicator functions. When scaling up an toy example to a larger scale, I get the following error:

```
numpy.core._exceptions._ArrayMemoryError: Unable to allocate 775. GiB for an array with shape (98, 191, 1, 2920, 34, 56) and data type object
```
Indeed, 98 nodes * 191 links * 1 carriers * 2920 timesteps * 34 node_groups * 56 link_groups create an array with ~100*1e9 elements.
However, this array is in reality very sparse - most of them are zero. The actual number of constraints is only 56 link_groups * 2 (upper and lower bound) * 2920 timesteps.

This is the custom math I am adding

```
constraints:
  constraint_link_flow_out_group_ub:
    foreach: [node_groups,link_groups,carriers,timesteps]
    where: "base_tech=transmission AND carriers=electricity AND flow_out_link_group"
    equations:
      - expression: sum(flow_out * default_if_empty(lookup_link_group, 0) * default_if_empty(lookup_node_group, 0), over=[techs,nodes]) <= flow_out_link_group
```

- [ ] macOS
- [ ] Windows
- [x] Linux

### Version

v0.7.0.dev4

### Relevant log output

```Shell

```

brynpickering:
Fixes #735

On investigating #735, I realised that a generalisation of `group_datetime` introduced in #778 might be prudent. This was an option considered prior to implementing #698 but discounted as it was less intuitive. I'm bringing it back as a potential major memory saver when doing more complex multi-dimensional grouping. 

One option to consider here is to remove `group_datetime` entirely and point people to create an input array mapping their datetime dimension to relevant datetime periods (e.g. timesteps -> date) and then use `group_sum`.

* Added `group_sum` expression helper function

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/docs/user_defined_math/helper_functions.md
+++ b/docs/user_defined_math/helper_functions.md
@@ -99,11 +99,10 @@ If the `<condition>` has any dimensions not present in `<math_component>`, `<mat
 
 ## group_sum
 
-Summing over a group of one or more dimension members may be necessary when setting constraints.
+Summing over a group of one or more dimension members in a memory-efficient way may be necessary when setting constraints.
 For instance, if setting upper flow limits on groups of transmission lines into / out of a set of nodes.
-Combining the [`where`](#where) and [`sum`](#sum) helper functions could be used to achieve this but can create very undesirably large arrays in memory.
-`group_sum(<math_component>, <groupby_array>, <group_dimension>)` allows you to define a `groupby` array matching dimension members to groups (e.g. `(node, tech)` combinations to a members of a new `my_groups` dimension) and sum over those dimensions to create a new dimension over which the array is indexed (here: `my_groups`).
-Once completed, this helper function will return the math component indexed over its original dimensions - the groupby dimensions + the new grouper dimension (e.g., `[techs, nodes, carriers, timesteps]` -> `[my_groups, carriers, timesteps]`).
+`group_sum(<math_component>, <groupby_array>, <group_dimension>)` allows you to define a `groupby` array matching dimension members to groups (e.g. `(node, tech)` combinations to a members of a new `old_fossil_plants` dimension) and sum over those dimensions to create a new dimension over which the array is indexed (here: `old_fossil_plants`).
+Once completed, this helper function will return the math component indexed over its original dimensions - the groupby dimensions + the new grouper dimension (e.g., `[techs, nodes, carriers, timesteps]` → `[old_fossil_plants, carriers, timesteps]`).
 You will need to account for this accordingly in you math `foreach` and other expression components.
 
 !!! note
--- a/docs/user_defined_math/helper_functions.md
+++ b/docs/user_defined_math/helper_functions.md
@@ -100,9 +100,10 @@ If the `<condition>` has any dimensions not present in `<math_component>`, `<mat
 ## group_sum
 
 Summing over a group of one or more dimension members in a memory-efficient way may be necessary when setting constraints.
-For instance, if setting upper flow limits on groups of transmission lines into / out of a set of nodes.
-`group_sum(<math_component>, <groupby_array>, <group_dimension>)` allows you to define a `groupby` array matching dimension members to groups (e.g. `(node, tech)` combinations to a members of a new `old_fossil_plants` dimension) and sum over those dimensions to create a new dimension over which the array is indexed (here: `old_fossil_plants`).
-Once completed, this helper function will return the math component indexed over its original dimensions - the groupby dimensions + the new grouper dimension (e.g., `[techs, nodes, carriers, timesteps]` → `[old_fossil_plants, carriers, timesteps]`).
+For instance, if setting upper flow limits on groups of transmission lines into / out of a set of nodes, or limiting outflow from different types of power plants.
+`group_sum(<math_component>, <groupby_array>, <group_dimension>)` allows you to sum over groups of dimension members in a math component.
+In the `groupby_array`, you match math component dimension members to members of a new `group_dimension` (e.g. `(node, tech)` combinations to types of `polluting_power_plants`: `(GBR, ocgt): high_particulate_emissions`, `(FRA, ccgt): low_particulate_emissions`).
+Once completed, this helper function will return the math component indexed over its original dimensions _minus_ the groupby dimensions _plus_ the new grouper dimension (e.g., `[techs, nodes, carriers, timesteps]` → `[polluting_power_plants, carriers, timesteps]`).
 You will need to account for this accordingly in you math `foreach` and other expression components.
 
 !!! note
--- a/docs/user_defined_math/helper_functions.md
+++ b/docs/user_defined_math/helper_functions.md
@@ -97,6 +97,18 @@ If the `<condition>` has any dimensions not present in `<math_component>`, `<mat
     `Where` gets referred to a lot in Calliope math.
     It always means the same thing: applying [xarray.DataArray.where][].
 
+## group_sum
+
+Summing over a group of one or more dimension members may be necessary when setting constraints.
+For instance, if setting upper flow limits on groups of transmission lines into / out of a set of nodes.
+Combining the [`where`](#where) and [`sum`](#sum) helper functions could be used to achieve this but can create very undesirably large arrays in memory.
+`group_sum(<math_component>, <groupby_array>, <group_dimension>)` allows you to define a `groupby` array matching dimension members to groups (e.g. `(node, tech)` combinations to a members of a new `my_groups` dimension) and sum over those dimensions to create a new dimension over which the array is indexed (here: `my_groups`).
+Once completed, this helper function will return the math component indexed over its original dimensions - the groupby dimensions + the new grouper dimension (e.g., `[techs, nodes, carriers, timesteps]` -> `[my_groups, carriers, timesteps]`).
+You will need to account for this accordingly in you math `foreach` and other expression components.
+
+!!! note
+    If you want to sum over a time period on a datetime dimension, consider using the `group_datetime` convenience helper function.
+
 ## group_datetime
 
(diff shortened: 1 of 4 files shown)


src/calliope/backend/helper_functions.py
```diff
@@ -142,7 +142,30 @@ def _replace_in_iterator(matched):
...
+
+        Note:
+            the iterator -> dimension conversion is very simple (appends an `s`) so it won't work with math components that have already have `_update_iterator` applied.
+        """
+
+        def __extract_dims(matched) -> str:
+            iterators = matched.group(2)
+            # Split on `,`, add 's' back in to singular iterators to refer to dimension names,
+            # then rejoin `,` as we must return a string.
+            return ",".join([i + "s" for i in iterators.split(",")])
```
irm-codebase, line 165: At this point, we really should just go for either the s or without the s as the standard. This is just polluting all the code with needless operations.

If you agree, I will create an issue to track this.

docs/user_defined_math/helper_functions.md
```diff
@@ -97,6 +97,18 @@ If the `<condition>` has any dimensions not present in `<math_component>`, `<mat
...
+
+Summing over a group of one or more dimension members may be necessary when setting constraints.
+For instance, if setting upper flow limits on groups of transmission lines into / out of a set of nodes.
+Combining the [`where`](#where) and [`sum`](#sum) helper functions could be used to achieve this but can create very undesirably large arrays in memory.
+`group_sum(<math_component>, <groupby_array>, <group_dimension>)` allows you to define a `groupby` array matching dimension members to groups (e.g. `(node, tech)` combinations to a members of a new `my_groups` dimension) and sum over those dimensions to create a new dimension over which the array is indexed (here: `my_groups`).
+Once completed, this helper function will return the math component indexed over its original dimensions - the groupby dimensions + the new grouper dimension (e.g., `[techs, nodes, carriers, timesteps]` -> `[my_groups, carriers, timesteps]`).
+You will need to account for this accordingly in you math `foreach` and other expression components.
+
+!!! note
+    If you want to sum over a time period on a datetime dimension, consider using the `group_datetime` convenience helper function.
```
irm-codebase: This is a bit hard to understand... Some brief improvements:

- I'd remove the third sentence (`Combining the [...] large arrays in memory`) and instead mention memory efficiency in the first sentence (`Summing over a group of one or more dimension members in a memory efficient way.`).
- Perhaps use more practical example than `my_groups`? e.g., a sum of fossil powerplant capacity per node, calling the 'group' `fossil_plants_in_node`?

src/calliope/backend/helper_functions.py
```diff
@@ -718,6 +741,97 @@ def as_array(self, array: xr.DataArray, condition: xr.DataArray) -> xr.DataArray
...
+                dims: [grouped_node_techs]
+            ```
+            3. Define the math to link the two, using `group_sum`:
+            ```yaml
+            constraints:
+              node_tech_group_max:
+                foreach: [grouped_node_techs, carriers, timesteps]
+                where: node_tech_limits
+                equations:
+                  - expression: group_sum(flow_out, group_techs_and_nodes, grouped_node_techs) <= node_tech_limits
```
irm-codebase: This looks a bit too hacky...

- `group_techs_and_nodes` feels like a parameter pretending to be a set (or a set pretending to be a parameter)?
- `grouped_node_techs` is another set that appears only so you can iterate over the parameter pretending to be a set...

Seems like both are doing the same thing, and `grouped_node_techs` only appears because we are forced to use `parameter` instead of a 'proper' dimension due to our current lack of support for them...

irm-codebase (commented):
Some comments on readability, and a few concerns...

I'll let @jnnr decide on this one, but this feels like something we should take care of _after_ fixing the params / dims features.

src/calliope/backend/helper_functions.py
brynpickering: They're not the same. In math you need `for <set_member> in <set>` or in our case `for <iterator> in <dim>`, e.g. `for i in A`. We keep verbosity by doing e.g. `for tech in techs`. So our dimensions are by default `<iterator>s`. When all we have are iterators, this method lets us get the `dim` that created it. 

It breaks down as soon as a dimension _isn't_ `<iterator>s` (very likely to happen when users add dims). Long-term solution would be to store a `iterator->dim` mapping somewhere that we can refer to to make that translation. You could open an issue to track that!

brynpickering: dims could even define their iterator in their metadata

src/calliope/backend/helper_functions.py
brynpickering: _xarray_ doesn't support storing subsets except as "parameters" (data variables). So, we are following that convention internally. This makes it much clearer to users. Otherwise we allow subsets to be defined and then the user sees them in `model.inputs` unexpectedly. Maybe that would still be ok (like "lookups" will become a separate section).

src/calliope/backend/helper_functions.py
irm-codebase: I think that last proposal is the way to go.

brynpickering: OK, then let's pass this hacky implementation through and deal with it in #783 or a similar PR

docs/user_defined_math/helper_functions.md
```diff
@@ -97,6 +97,17 @@ If the `<condition>` has any dimensions not present in `<math_component>`, `<mat
     `Where` gets referred to a lot in Calliope math.
     It always means the same thing: applying [xarray.DataArray.where][].
 
+## group_sum
+
+Summing over a group of one or more dimension members in a memory-efficient way may be necessary when setting constraints.
+For instance, if setting upper flow limits on groups of transmission lines into / out of a set of nodes.
+`group_sum(<math_component>, <groupby_array>, <group_dimension>)` allows you to define a `groupby` array matching dimension members to groups (e.g. `(node, tech)` combinations to a members of a new `old_fossil_plants` dimension) and sum over those dimensions to create a new dimension over which the array is indexed (here: `old_fossil_plants`).
```
irm-codebase: Nit-pick (wording):
Third sentence has a typo (extra 's')... perhaps you meant this?

"allows you to define a `groupby` array matching dimension members to group combinations (e.g. `(node, tech)` to a"

src/calliope/backend/helper_functions.py
```diff
@@ -798,19 +798,19 @@ def as_array(
...
-              node_tech_group_max:
-                foreach: [grouped_node_techs, carriers, timesteps]
-                where: node_tech_limits
+              node_tech_emission_group_max:
+                foreach: [emission_groups, carriers, timesteps]
+                where: emission_limits
                 equations:
-                  - expression: group_sum(flow_out, group_techs_and_nodes, grouped_node_techs) <= node_tech_limits
+                  - expression: group_sum(flow_out, group_techs_and_nodes, emission_groups) <= emission_limits
             ```
```
irm-codebase: `group_techs_and_nodes` -> `power_plant_groups`

irm-codebase (changes requested):
Minor comments.
Test coverage is missing a case for `GroupSum`. After that is in I'll approve.

src/calliope/backend/helper_functions.py
sjpfenninger: > dims could even define their iterator in their metadata

Also think this is the way to go

sjpfenninger:
This looks fine to me once Ivan's comments are addressed

docs/user_defined_math/helper_functions.md
```diff
@@ -97,6 +97,18 @@ If the `<condition>` has any dimensions not present in `<math_component>`, `<mat
...
+For instance, if setting upper flow limits on groups of transmission lines into / out of a set of nodes, or limiting outflow from different types of power plants.
+`group_sum(<math_component>, <groupby_array>, <group_dimension>)` allows you to sum over groups of dimension members in a math component.
+In the `groupby_array`, you match math component dimension members to members of a new `group_dimension` (e.g. `(node, tech)` combinations to types of `polluting_power_plants`: `(GBR, ocgt): high_particulate_emissions`, `(FRA, ccgt): low_particulate_emissions`).
+Once completed, this helper function will return the math component indexed over its original dimensions _minus_ the groupby dimensions _plus_ the new grouper dimension (e.g., `[techs, nodes, carriers, timesteps]` → `[polluting_power_plants, carriers, timesteps]`).
+You will need to account for this accordingly in you math `foreach` and other expression components.
+
+!!! note
+    If you want to sum over a time period on a datetime dimension, consider using the `group_datetime` convenience helper function.
+
```
irm-codebase, line 111: Nit-pick: unnecessary. It's below this, and this just bloats the docs.

irm-codebase (approved):

irm-codebase:
@brynpickering seems like some tests are failing for Mac-OS? Otherwise I agree with the PR

docs/user_defined_math/helper_functions.md
brynpickering: @irm-codebase as mentioned in the PR title: do you think `group_datetime` is even needed now? It's a convenience function with added overhead on our side. The alternative is that the user defines a data table with the mapping from e.g. timesteps to date and uses `group_sum` instead. Is that asking too much of the user...?

irm-codebase: If we already have this code, I do not really see a problem in keeping it.
Particularly because it saves users files...

I'd only remove it if the reduced overhead is significant enough to merit it.

PR #790, opened 2025-08-07
Add random seed to spores tests

### LINKED ISSUE #789: Tests that rely on random seeds will randomly fail
### What happened?

The SPORES test that checks the `random` scoring algorithm will randomly fail as it can sometimes apply a score of zero to all technologies. To mitigate this, we can set a random seed for our test suite. It isn't ideal as it can be useful to check that randomness is working as expected. We can consider providing a way to isolate the random seeding in certain tests but this shouldn't be a priority right now as none of our tests need it.

- [x] macOS
- [ ] Windows
- [ ] Linux

### Version

v0.7.0.dev6

### Relevant log output

```Shell
FAILED tests/test_core_model.py::TestSporesMode::test_spores_caps[random] - AssertionError: assert <xarray.DataArray 'flow_cap' ()> Size: 1B\narray(False)
 +  where <xarray.DataArray 'flow_cap' ()> Size: 1B\narray(False) = any()
 +    where any = <xarray.DataArray 'flow_cap' (spores: 2, nodes: 2, techs: 4, carriers: 1)> Size: 128B\narray([[[[0.],\n         [0.],\n  ...iers  (carriers) object 8B 'electricity'\n  * nodes     (nodes) <U1 8B 'a' 'b'\n  * spores    (spores) object 16B '1' '2' != 0.any
= 1 failed, 2589 passed, 132 skipped, 29 xfailed, 1 xpassed, 1 warning in 421.19s (0:07:01) =
```

brynpickering:
Fixes #789 

I tried this first as a globally applied pytest fixture but it didn't fix the seed as expected as I don't think these fixtures are automatically used _inside other fixtures_. Instead, I've defined it within the specific SPORES test class. That way it is isolated. If we introduce other random sampling methods we may need to set random seeds for those but this way we can deal with it on a case-by-case basis.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/tests/test_core_model.py
+++ b/tests/test_core_model.py
@@ -369,7 +369,6 @@ class TestSporesMode:
         cap_diffs = spores_model.results.flow_cap.sel(spores=order_dim).diff(
             dim="spores"
         )
-        breakpoint()
         assert (cap_diffs != 0).any()
 
     def test_spores_algo_log(self, spores_model_and_log_algorithms):


irm-codebase (dismissed):
Seems good!

Only nitpick:
I'd consider defining this in `conftest.py`, to ensure all other tests use the same seed.

PR #793, opened 2025-08-07
Add global expression order option

### LINKED ISSUE #773: Allow global expressions to be reordered
### What can be improved?

If users define their own global expressions then they get appended to the set of global expressions. If they then reference those new global expressions in an _existing_ global expression, the ordering means the existing global expression won't understand the reference. This cropped up in #727 and could be solved by allowing an optional "order" or "zorder" attribute to be defined. This is usually defined as e.g. `-1` to elevate a global expression to the start of the running order. Or users could be requested to list the global expressions in the config in the order they want them to be defined.

### Version

v0.7.0dev6

brynpickering:
Fixes #773 

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


docs/user_defined_math/components.md
```diff
@@ -64,6 +64,13 @@ The equation expressions do _not_ have comparison operators; those are reserved
...
 1. It can take on a `default` value that will be used in math operations to avoid `NaN` values creeping in.
 The default value should be set such that it has no impact on the optimisation problem if it is included (most of the time, this means `NaN`).
+1. It can have an `order` defined to reprioritise its addition to the optimisation problem.
+   This is often necessary when adding new global expressions that you will use in other global expressions since they will need to be defined _before_ the other global expressions that refer to them.
+   For instance, if tracking a new cost as a global expression (say, `cost_operation_monthly`), you will want to add it to the overall `cost` global expression so it is tracked in the objective.
+   This means `cost_operation_monthly` has to be higher in the order of global expressions than `cost`.
+   But, because you defined it as user-defined math, it will be _appended_ to the set of global expressions by default.
+   To bring its order higher you can define `order` as a suitably small number.
+   Usually, a negative number is used (e.g. `-1`) to ensure you move it to before _all_ pre-defined global expressions.
```
irm-codebase: Nitpick:
I'd consider replacing most of this text with a brief yaml example. If you're unfamiliar with this, it might be hard to picture.

irm-codebase (approved):
Looks good, although in theory this `order` could be completely avoided with a bit of recursive magic.

A similar situation happened with the templates in `YAML` files, where a template may even re-use another template. All you need is to call the process again if you encounter a definition, and hold a register with the order of current calls. This also allows you to detect cyclic requests (ie: two global expressions referring to each other).

This does imply more code, so I'll leave the choice to you. This could be a feature added in v0.7+.

irm-codebase (approved):

PR #796, opened 2025-08-20
Fix separate milp math

### LINKED ISSUE #749: `lb` / `up` are printed incorrectly

irm-codebase:
Fixes #749

- [x] New `extra_math` file `milp.yaml`
- [x] Deactivation of 'inexistent' math components won't fail if said component is not present (an empty component will be added, but will be left unused).
- [x] Improved custom math testing (better extensibility)

## Reviewer checklist

- [x] Documentation updated
- [x] Changelog updated


docs/migrating.md
```diff
@@ -374,11 +374,13 @@ Investment costs are split out into the component caused by annual operation and
...
 ### Explicitly triggering MILP and storage decision variables/constraints
 
+For easier extensibility, the mixed-integer formulation of Calliope is now an [extra math option](./pre_defined_math/index.md).
+
 In v0.6, we inferred that a mixed-integer linear model was desired based on the user defining certain parameters.
 For example, defining `units_max` would trigger the integer `units` decision variable.
 Defining the `purchase` cost parameter would trigger the binary `purchased` decision variable.
 
-Now, you need to explicitly set the method using `cap_method`:
+Now, you need to explicitly set the method using `cap_method` and enable the mixed-integer math:
```
brynpickering, line 383: I think this is a bit dangerous since a user can easily set `cap_method` and not realise that none of it is being activated in the actual math. It's probably not a major issue since in #783 we log the inputs that are not invoked by the loaded math so the user will see that although they have defined `cap_method`, it is not being used in the math.

src/calliope/math/milp.yaml
```diff
@@ -0,0 +1,400 @@
...
+
+  flow_cap:
+    title: Technology flow (a.k.a. nominal) capacity
+    description: >-
+      A technology's flow capacity, also known as its nominal or nameplate capacity.
+    default: 0
+    unit: power
+    foreach: [nodes, techs, carriers]
+    bounds:
+      min: 0  # set in a distinct constraint to handle the integer purchase variable
```
brynpickering: You actually only need to define `flow_cap.bounds.min: 0` as the rest will come from the base math. This reduces maintenance burden and makes it clearer what is different but also obviously hides what was originally defined, forcing the user to refer to `plan.yaml`.

brynpickering: Same for all the others below

src/calliope/math/milp.yaml
```diff
@@ -0,0 +1,400 @@
...
+      with or without integer capacity purchasing.
+    foreach: [nodes, techs]
+    where: "base_tech==supply AND source_cap_min"
+    equations:
+      - where: NOT purchased_units
+        expression: source_cap >= source_cap_min
+      - where: purchased_units
+        expression: source_cap >= source_cap_min * purchased_units
+
+  # Modified
```
brynpickering, line 305: Again, you could limit the definition to just those sections that have been modified

src/calliope/math/milp.yaml
```diff
@@ -0,0 +1,400 @@
...
+    order: -1
+    equations:
+      - where: base_tech==transmission
+        expression: (cost_purchase + cost_purchase_per_distance * distance) * purchased_units * 0.5
+      - where: "NOT base_tech==transmission"
+        expression: cost_purchase * purchased_units
+
+  # Modified
+
+  cost_investment:
```
brynpickering, line 342: again, you could just limit this to just those sections that have been modified

src/calliope/math/plan.yaml
```diff
@@ -301,197 +301,7 @@ constraints:
...
-
-  async_flow_out_milp:
-    description: >-
-      Set a technology's ability to have outflow in the same timestep that it has inflow,
-      for any technology using the asynchronous flow binary switch.
-    foreach: [nodes, techs, timesteps]
-    where: "async_flow_switch"
-    equations:
-      - expression: sum(flow_out, over=carriers) <= async_flow_switch * bigM
+      - expression: flow_export <= export_max
```
brynpickering: this limit could now be set as a variable bound rather than a constraint, I think

src/calliope/schemas/math_schema.py
```diff
@@ -67,7 +67,7 @@ class SubExpressions(CalliopeDictModel):
 class Constraint(MathIndexedComponent):
     """Schema for named constraints."""
 
-    equations: Equations
+    equations: Equations = Equations()
```
brynpickering, line 70: I thought we said we didn't want to set defaults for these essential parts of the math components?

src/calliope/schemas/math_schema.py
```diff
@@ -124,9 +124,9 @@ class Bounds(CalliopeBaseModel):
     a single value that is applied across all decision variable index items.
     """
 
-    max: AttrStr | NumericVal
+    max: AttrStr | NumericVal = float("inf")
     """Decision variable upper bound, either as a reference to an input parameter or as a number."""
-    min: AttrStr | NumericVal
+    min: AttrStr | NumericVal = 0
```
brynpickering: Should be -.inf as a default, if we're putting them in.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### User-facing changes
 
+|changed| MILP math formulation now has its own file (`milp.yaml`), which can be requested via `extra_math`.
```
brynpickering: add issue number?

src/calliope/math/milp.yaml
irm-codebase: I mostly added the full definition for readability reasons since the file will be featured in the documentation...

brynpickering: When rendered, it merges together the content of `plan.yaml` anyway. You can see it [here](https://calliope--796.org.readthedocs.build/en/796/math/storage_inter_cluster/#balance_supply_with_storage) for a inter cluster storage constraint. The docs do seem to be ignoring all my functions that are supposed to flag if a component is being added to, removed from, or modified from the base math, but that's a different issue 😅

src/calliope/schemas/math_schema.py
irm-codebase: I thought to use it... but 0 seemed computationally safer since it forces a bound in one direction.
I'll modify

src/calliope/schemas/math_schema.py
irm-codebase: Unfortunately, we need to do this in order to be able to 'deactivate' non existent components without conflicts.
Pydantic will add a barebones version of the component that is deactivated.

Otherwise, we'd need a specific `operate_milp.yaml` file, and it'd be cumbersome to users.

src/calliope/math/plan.yaml
irm-codebase: For posterity: turns out this constraint was not even necessary at all: `export_balance` takes care of this.
Additionally, `flow_export_max` should've had `flow_export * operating_units <= export_max`.

docs/migrating.md
irm-codebase: Agreed. Perhaps we could throw in a warning, once the new math approach is finalised?

src/calliope/schemas/math_schema.py
brynpickering: It may be worthwhile to have a test for using `milp` with `operate` mode to confirm that it doesn't break horribly? I'm not suggesting an LP comparison, just a smoke test.

irm-codebase: Makes sense. I've added this by first running this case on the main branch, grabbing the results and then using them in the test on this branch.

Works well!

brynpickering (approved):

Issue #798, opened 2025-08-20
Store objective linked to a set of results in the model object on save


brynpickering:
### What can be improved?

The objective linked to a particular set of results is only reliably stored in `model.backend.objective`. 

We could store the latest objective by updating `model.config.solve.objective` after a successful solve. However, in iterative solves (like SPORES), the objective can change between iterations (from `min_cost_optimisation` in the baseline run to `min_spores` in all subsequent runs).  To store the objective alongside results we need it to be indexed over the iterated dimension or, like "termination_condition", to be stored as a comma separated string. 

### Version

v0.7.0.dev6

PR #799, opened 2025-08-20
Fix tracking set objective in the backend

### LINKED ISSUE #797: Updating objective expression in the gurobi backend doesn't update the objective
### What happened?

In gurobipy, the objective function isn't mutable. Updating the underlying expression that is used to define the objective does not update the objective function. Instead, you have to manually _set_ the objective as the latest state of that expression before solving.

Noticed this on running SPORES with the gurobi backend...

- [x] macOS
- [ ] Windows
- [ ] Linux

### Version

v0.7.0dev6

### Relevant log output

```Shell

```

brynpickering:
Fixes #797 

* Current objective hangs around more explicitly in the backend (need to work out a way to store the active objective linked to a set of results - #798)
* On recreating backend components completely, changes to math made with `model.backend.add_[component](...)` are passed.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/backend/backend_model.py
+++ b/src/calliope/backend/backend_model.py
@@ -65,8 +65,6 @@ class BackendModelGenerator(ABC, metaclass=ABCMeta):
     _PARAM_DESCRIPTIONS = extract_from_schema(MODEL_SCHEMA, "description")
     _PARAM_UNITS = extract_from_schema(MODEL_SCHEMA, "x-unit")
     _PARAM_TYPE = extract_from_schema(MODEL_SCHEMA, "x-type")
-    objective: str
-    """Optimisation problem objective name."""
 
     def __init__(
         self,
@@ -90,7 +88,18 @@ class BackendModelGenerator(ABC, metaclass=ABCMeta):
         self.defaults: dict = deepcopy(defaults)
         self._solve_logger = logging.getLogger(__name__ + ".<solve>")
 
+        self._objective: str = self.config.objective
+
         self._check_inputs()
+
+    @property
+    def objective(self) -> str:
+        """Optimisation problem objective name."""
+        return self._objective
+
+    @objective.setter
+    def objective(self, value: str):
+        self._objective = value
 
     @abstractmethod
     def add_parameter(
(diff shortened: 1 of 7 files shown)


src/calliope/backend/backend_model.py
```diff
@@ -90,8 +88,19 @@ def __init__(
...
 
+    @property
+    def objective(self) -> str:
+        """Optimisation problem objective name."""
+        return self._objective
+
+    @objective.setter
+    def objective(self, value: str):
+        self._objective = value
+
```
irm-codebase: Let's just expose the value. This approach just drives up complexity for little gain.
The backend is already the domain of `calliope` experts, and [exposing attributes is a python standard/common practice](https://dirtsimple.org/2004/12/python-is-not-java.html?utm_source=chatgpt.com) (check the setter/getter bit). 

I trust that the dev of `setuptools` knows what they're talking about ;)

src/calliope/backend/backend_model.py
```diff
@@ -366,7 +375,7 @@ def _add_component(
         )
         if name not in self.math[component_type]:
             self._dataset.attrs["applied_math"].union(
-                {f"{component_type}.{name}": component_dict}
+                {f"{component_type}.{name}": component_dict}, allow_override=True
```
irm-codebase, line 372: Shouldn't the name be added to the `math` object too?
This is a potential issue when saving the model: your `math` won't match what is there.

irm-codebase (changes requested):
Requesting a small change (removing unnecessary properties).

I am not sure if this is a complete fix. The update to any `math` component should be one level above, because we must also ensure that the `math` objects contain the new component, and that it is validated (`math_schema.Objective.model_validate(new_objective)`)... otherwise it won't be saved in the netCDF file and it may be malformed.

irm-codebase:
@brynpickering I'd be OK in approving if we agree that #712 should also ensure math updates keep the `math` objects in sync, and the `@property` bit is removed (since it functionally does nothing).

src/calliope/backend/backend_model.py
brynpickering: We add it back to the `math` object in `calliope.Model` when the problem is solved and the results are passed back. The idea being that the math shouldn't be edited while there are no results.

irm-codebase (approved):

Issue #638, opened 2024-07-16
Isolate postprocessing data creation methods (YAMLification)


irm-codebase:
### What can be improved?

At the moment, `model.solve` will automatically run several post processing methods.

However, this might not be desired in some circumstances, and is a [side-effect](https://luzkan.github.io/smells/side-effects). It's much more valuable to give users the option to run these commands at request instead of forcing them.

I recommend the following:
- Doing some slight re-work to `postprocessing` so that it's a module users can call and pass `Model` objects to.
- Removing the use of "optional" `postprocessing` methods in `model.py`.
- Moving "essential" postprocessing methods (`clean_results`?) into `model.py`.

This should result in leaner code and more value for users. Let me know what you think!

### Version

v0.7

brynpickering:
Sounds good. We just need a way to trigger preprocessing steps when running calliope from the command line.

brynpickering:
I've been thinking about adding a `model.postprocess(...)` step, which would be linked to a `config.postprocess` config level.

It would allow the results to be cleaned/expanded in whatever way the user has configured. A question remains as to whether this should return a post-processed result array or to update the results in-place. If updating in-place, the original results would still be available in an interactive session (they are extracted from optimisation backend objects, which we could just do again). We have had problems in the past with returning a result array that is independent of a Calliope object. Namely, it makes saving the file with relevant metadata a pain.

irm-codebase:
Hmm... could you elaborate on what those saving issues were?
In my mind, modifying the output is not a good idea. It could easily spiral out of control in terms of checks and edge cases...
We should ensure that whatever we save to the `netCDF` / `zarr` file is _only_ what is necessary to rebuild the same calliope model (config, math, inputs, results???).

If we make sure that postprocessing is only for data transformations, then returning arrays which can be saved seems best to me...
No postprocessing method should impact constraints or be directly relevant to the optimisation problem.
If that is the case, then they are not really post-processing in the way I thought for this issue...

For example, calculating capacity factors does not really alter data. It just depends on it.

In my mind, these steps should always return the same postprocess result:
- init -> build -> solve -> post-process "thing"
- init -> build -> save -> open -> solve -> post-process "thing"
- init -> build-> solve -> save -> open -> post-process "thing"

brynpickering:
[...]

PR #802, opened 2025-08-22
Add post processing registry

### LINKED ISSUE #747: results postprocessing as an option?
### What can be improved?

Would it be possible to make certain post-processed calculations, such as the capacity factor, optional?
In large models, these calculations can become very memory-intensive, and they might not always be necessary for every user. Providing the option to enable or disable them could help manage memory usage more efficiently depending on the use case.

### Version

v0.6.10, v0.7.0

irm-codebase:
Fixes #747, #638

- [x] `config.solve.enable_postprocessing` can be used to avoid running those functions (defaults to `False` to save computation costs).
- [x] zero threshold cleanup is now less memory hungry (uses xarray's built in where, and does not create extra datasets)
- [x] new thread-safe `calliope.postprocess.orchestrator.REGISTRY` for post processing functions which can be extended by users or us, and also specifies which base math files are compatible with each postprocessing function.

## Reviewer checklist

- [ ] Documentation updated
- [x] Changelog updated


src/calliope/postprocess/functions.py
```diff
@@ -0,0 +1,113 @@
...
+from calliope.postprocess.orchestrator import PostprocessContext, postprocessor
+
+
+def _get_data(name: str, context: PostprocessContext) -> xr.DataArray:
+    """Get an element from either inputs or results."""
+    if name in context.results.keys():
+        data = context.results[name]
+    else:
+        data = context.inputs[name]
+    return data
```
brynpickering:
```suggestion
def _get_data(name: str, context: PostprocessContext) -> xr.DataArray:
    """Get an element from either inputs or results."""
    data = context.results.get(name, context.inputs[name])
    return data
```

src/calliope/postprocess/orchestrator.py
```diff
@@ -0,0 +1,175 @@
...
+        self,
+        function: PostprocessFunction,
+        *,
+        name: str,
+        base_math: Iterable[str],
+        order: int,
+        enabled: bool,
+        overwrite: bool,
+    ) -> None:
+        """Add an entry to the postprocessing registry."""
```
brynpickering: Write out what the args are

src/calliope/postprocess/orchestrator.py
```diff
@@ -0,0 +1,175 @@
...
+                # Store postprocesses in the model results.
+                LOGGER.info("Postprocessing: applying %s postprocess function.", name)
+                postprocessed_data = function(ctx)
+                if postprocessed_data is not None:
+                    results[name] = postprocessed_data
+
+            except Exception as ex:
+                raise ModelError(f"Postprocess '{name}' failed.") from ex
+
+    _apply_zero_threshold(results, config.solve.zero_threshold)
```
brynpickering, line 191: should we skip zero thershold if it is set to zero (i.e. has no effect)? Then you don't have to run the `xr.where` or the `np.abs` on the data.

src/calliope/postprocess/orchestrator.py
```diff
@@ -0,0 +1,175 @@
...
+
+        for settings, function in applicable_postprocesses:
+            name = settings.name
+            if (name in results) or name in (results.dims):
+                raise ValueError(
+                    f"Postprocess '{name}' attempted to overwrite an existing variable/coord/dim."
+                )
+            try:
+                # Store postprocesses in the model results.
+                LOGGER.info("Postprocessing: applying %s postprocess function.", name)
```
brynpickering, line 184: Rogue to not choose an f-string!
```suggestion
                LOGGER.info(f"Postprocessing: applying {name} postprocess function.")
```

src/calliope/postprocess/orchestrator.py
```diff
@@ -0,0 +1,175 @@
...
+        overwrite (bool, optional): whether or not to allow overwriting functions. Defaults to False.
+    """
+
+    def decorator(func: F) -> F:
+        REGISTRY.register_function(
+            func,
+            base_math=base_math,
+            name=name or func.__name__,
+            order=order,
+            enabled=enabled,
```
brynpickering: what's the value of `enabled`? In the core code at least there's no way of anyone accessing it...

brynpickering: Call it `active` to align with our math / model def syntax?

src/calliope/schemas/config_schema.py
```diff
@@ -190,6 +190,9 @@ class Solve(CalliopeBaseModel):
     solver: str = Field(default="cbc")
     """Solver to use. Any solvers that have Pyomo interfaces can be used. Refer to the Pyomo documentation for the latest list."""
 
+    enable_postprocessing: bool = False
+    """Whether or not to run postprocessing functions."""
+
     zero_threshold: float = Field(default=1e-10)
```
brynpickering, line 199: Update docstring of zero_threshold to clarify that it is run whether or not postprocessing is enabled and that only setting it to a value of zero will disable it?

src/calliope/postprocess/orchestrator.py
```diff
@@ -0,0 +1,175 @@
...
+from calliope.exceptions import ModelError
+from calliope.schemas.config_schema import CalliopeConfig
+from calliope.schemas.general import AttrStr, NonEmptyUniqueList
+from calliope.util.tools import listify
+
+LOGGER = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, slots=True)
+class PostprocessContext:
```
brynpickering, line 26: For easier maintainability by aligning ith the rest of the core code, should this just be a pydantic model?

src/calliope/postprocess/orchestrator.py
```diff
@@ -0,0 +1,175 @@
...
+        """Initialise with an empty registry."""
+        self._entries: dict[str, tuple[PostprocessSettings, PostprocessFunction]] = {}
+        self._lock = threading.Lock()
+
+    def register_function(
+        self,
+        function: PostprocessFunction,
+        *,
+        name: str,
+        base_math: Iterable[str],
```
brynpickering, line 62: Does `base_math` _just_ look at `base`? Ideally it would look at the full list of `extra_math`/ `apply_math` and decide based on what the user has actually applied to their given use case.

brynpickering: Although there will always be the danger of wanting to postprocess slightly differently between two instances applied math and them becoming conflicted (e.g. between `operate` and `plan`, but then you have _both_ defined in your applied math).

brynpickering (changes requested):
Very neat! It makes me wonder whether we should use decorator syntax for our helper function registry (we'd then need sub-decorators to handle returning arrays vs math strings, but still might be easier to understand and for users to define their own).

It might be a bit overkill in terms of maintenance but can probably also remain relatively untouched once in.

Things it still needs (which you probably know):

1. tests for the added registry
2. documentation, including for users to add their own
3. More verbose docstrings including arg and returns descriptions. Use copilot to help you there!

src/calliope/postprocess/orchestrator.py
irm-codebase: Perhaps? The one bit I do not like is that an absolute zero threshold may be bad practice. I'll add a warning message, which should also make testing this easy.

src/calliope/postprocess/orchestrator.py
irm-codebase: Not rogue at all ;)
https://google.github.io/styleguide/pyguide.html#3101-logging

src/calliope/postprocess/orchestrator.py
irm-codebase: Not really... `pydantic`'s strength is for user inputs, and this is fully internal.
I did try doing this, but it did not really aid much, and only overcomplicated the code.

This is mostly because `pydantic` is best when evaluating to `python` native objects (`int`, `dict`, `list`, etc), and these are xarray datasets.

src/calliope/postprocess/functions.py
irm-codebase: That sadly won't work since you'll get a `KeyError` if `name` is not in `results`. I've updated this a bit so it's faster (`try`/`except`), though.

brynpickering: No, it shouldn't. It will fall back on `context.inputs[name]` if it isn't found in results. One thing we should possibly do is return the default value defined in math if it isn't found in either of those, rather than failing with a key error. There's nothing worse than your model failing in postprocessing after a long solve!

src/calliope/postprocess/orchestrator.py
brynpickering: Having no zero threshold is fine. This is mostly here for models solved with barrier without crossover (I.e. very large models where we care more about a solution than about to these tiny numbers littering our solution). For a simplex solved model there may not be any cleaning to do at all and so time and memory savings to be made by not applying a zero threshold

src/calliope/postprocess/functions.py
irm-codebase: It does fail, unfortunately. I did test it... unless I'm missing something.

Regarding the default: I am not sure if that would be safe... if the value is in neither results or inputs, it means that the function is asking for undefined data, and using the default would be misleading.

What we can do is throw out a warning if a postprocessing function failed, instead of an Error. I can solve that in the `postprocess_results` function.

brynpickering: Weird. I've definitely used that before without issue. Anyway, a warning rather than error sounds good. Using the default could make sense simply because that's already what we do in the math when some data is missing. Indeed, when combining with inputs that are often undefined (e.g. flow_out_eff) you would actively want all nan values filled with the default (1) rather than just receiving lots of nans back. Perhaps in that latter case it's up to the user to fill Nana's with the default from the math in their postprocessing function but I don't see why it shouldn't be extended to the array having _never_ been defined

src/calliope/postprocess/orchestrator.py
irm-codebase: It only looks at `base`, yeah. Mostly because the idea is to make these compatible with the base math, and if the user defines their own base math they can introduce their post processes without affecting standard models.

I could make this into a set, and then run the post process only if all the names match those of the built math (`(model.math_priority & spec.math) == spec.math`).
This way we could define:

```python

@postprocessor(math={"plan"}, order=0, enabled=True)
def some_plan_value(context: PostprocessContext) -> xr.DataArray:

@postprocessor(math={"plan", "operate"}, order=0, enabled=True)
def some_operate_value(context: PostprocessContext) -> xr.DataArray:

@postprocessor(math={"plan", "spores"}, order=0, enabled=True)
def some_spores_value(context: PostprocessContext) -> xr.DataArray:
```

src/calliope/postprocess/orchestrator.py
irm-codebase: This is `enabled` in the `@postprocessor` decorator

```python
@postprocessor(base_math={"plan"}, order=0, enabled=True)
def capacity_factor(context: PostprocessContext) -> xr.DataArray:
```

It can be set via the `set_enabled` function
`REGISTRY.set_enabled('capacity_factor', enabled=False)`.

(these will be changed to `active`, as requested in another comment)

src/calliope/schemas/config_schema.py
irm-codebase: I updated both to increase clarity. I've also changed `enable_postprocessing` to `postprocessing_active` to fit our terminology better.

irm-codebase:
Dismissed in favour of #812

PR #808, opened 2025-08-28
Test improvements

irm-codebase:

* Improved `find` method for built math objects
* Added post-validation of duplicate items in built math object (only params / variables are allowed to have duplicates)
* Improved test coverage

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/schemas/math_schema.py
+++ b/src/calliope/schemas/math_schema.py
@@ -3,6 +3,7 @@
 """Schema for Calliope mathematical definition."""
 
 from collections.abc import Iterable
+from itertools import chain
 from typing import Literal, Self
 
 from pydantic import Field, model_validator
@@ -55,11 +56,11 @@ class MathComponent(CalliopeBaseModel):
 class Dimension(MathComponent):
     """Schema for named dimension."""
 
-    dtype: Literal["string", "datetime", "date", "integer", "float"]
+    dtype: Literal["string", "datetime", "date", "integer", "float"] = "string"
     """The data type of this dimension's items."""
     ordered: bool = False
     """If True, the order of the dimension items is meaningful (e.g. chronological time)."""
-    iterator: str
+    iterator: str = "NEEDS_ITERATOR"
     """The name of the iterator to use in the LaTeX math formulation for this dimension."""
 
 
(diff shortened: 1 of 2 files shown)


src/calliope/schemas/math_schema.py
```diff
@@ -364,24 +365,59 @@ class CalliopeBuildMath(CalliopeBaseModel):
     checks: Checks = Checks()
     """Checks to apply before building the optimisation problem."""
 
+    @model_validator(mode="after")
+    def unique_component_names(self):
+        """Ensure all component names are unique."""
+        # Variables and Parameters can be swapped, so duplicates are allowed.
+        varparam = set(self.variables.root) | set(self.parameters.root)
+
```
brynpickering: so can lookups and params and variables and global expressions and params and global expressions. Ultimately, the backend will require unique keys for all components (even between constraints and variables) as they are all stored in one dataset. Therefore, I would just set this as being a comparison between all math components.

src/calliope/schemas/math_schema.py
```diff
@@ -364,24 +365,59 @@ class CalliopeBuildMath(CalliopeBaseModel):
...
+        # All other components must not share names with each other or with var/param.
+        groups = sorted(
+            chain(
+                (varparam,),
+                (
+                    set(getattr(self, n).root)
+                    for n in type(self).model_fields
+                    if n not in {"variables", "parameters"}
```
brynpickering: use `typing.get_args(ALL_COMPONENTS_T)` to get all fields that are not "checks"?

src/calliope/schemas/math_schema.py
```diff
@@ -364,24 +365,59 @@ class CalliopeBuildMath(CalliopeBaseModel):
...
             for field in to_search
             if (item := getattr(self, field).root.get(component)) is not None
             and item.active
         }
+        if not found:
+            raise KeyError(f"Component name `{component}` not found in math schema.")
         if len(found) > 1:
             raise ValueError(
                 f"Component name `{component}` found in multiple places: {found.keys()}."
             )
```
brynpickering: We shouldn't need this any more since we don't allow duplicates through earlier in the process  (can we even test for it?)

src/calliope/schemas/math_schema.py
```diff
@@ -364,24 +365,59 @@ class CalliopeBuildMath(CalliopeBaseModel):
...
         self, component: str, subset: Iterable[COMPONENTS_T] | None = None
     ) -> tuple[str, MathComponent]:
         """Find a component in the math schema."""
-        to_search = subset or self.model_fields_set - {"checks"}
+        to_search = (
+            set(subset)
+            if subset is not None
+            else set(type(self).model_fields) - {"checks"}
+        )
+
```
brynpickering: I don't see what was wrong with my less verbose equivalent?

tests/test_schemas/test_math_schema.py
```diff
@@ -26,3 +33,124 @@ def test_global_expression_order(self):
...
+        """Finding duplicate parameters should return an error."""
+        with pytest.raises(
+            ValueError, match=f"Component name `{to_search}` found in multiple"
+        ):
+            base_math_validated.find(to_search)
+
+    @pytest.mark.parametrize(
+        "to_search", ["flow_in_inc_eff", "datesteps", "non_existent", 3, True, 1.0]
+    )
+    def test_find_not_found_error(self, to_search, base_math_validated):
```
brynpickering: almost a full house! missing docstring here ;)

brynpickering (changes requested):

src/calliope/schemas/math_schema.py
irm-codebase: The problem of doing this is that it forbids you from deactivating a parameter and then setting it as a variable (or vice-versa). That would always lead to a crash, even though it's valid and we often do it.

I do not think this applies to the lookups / params case (or at least I do not see the utility...)
I could add global expressions to this group, as I think it does apply.

would that work?

src/calliope/schemas/math_schema.py
irm-codebase: `model_field_set` may omit something if it has not been directly set by users.
This is just to avoid future bugs in case some default is added to the schema which should be parsed.

src/calliope/schemas/math_schema.py
irm-codebase: ok, what I can do is remove non-active components from the check (if I'm not doing that already).

src/calliope/schemas/math_schema.py
irm-codebase: I'd say that it's ok to be defensive. This protects us in case something leaks through.
It should not occur in any case, but having it there is a nice insurance.

irm-codebase:
Review updates:

- Both the "after" validator and the find omit inactive components.
- Improved tests a bit
- Added docstrings to tests

src/calliope/schemas/math_schema.py
brynpickering: ok. I used `model_field_set` because `model_fields` is going to be deprecated. Not sure what alternative is available for `model_fields` as they don't offer one in their futurewarning

irm-codebase: I used `set(type(self).model_fields)` for the same reason. The warning message mentioned to access these from the class itself, not the instance.

PR #812, opened 2025-09-01
Add postprocess toggle

### LINKED ISSUE #747: results postprocessing as an option?

irm-codebase:
Fixes #747

* Added a simple toggle to turn off postprocessing
* Added a dataclass that mimics key `Model` attributes in `schemas`. This can be used by external or internal functions to process models in a 'duck-like' fashion, avoiding cyclic imports.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


src/calliope/model.py
```diff
@@ -372,9 +371,7 @@ def solve(self, force: bool = False, warmstart: bool = False, **kwargs) -> None:
 
         # Add additional post-processed result variables to results
         if results.attrs["termination_condition"] in ["optimal", "feasible"]:
-            results = postprocess_results.postprocess_model_results(
-                results, self.inputs, self.config.solve.zero_threshold
-            )
+            results = postprocess.postprocess_model_results(results, self)
```
sjpfenninger, line 374: Wouldn't it make more sense to have the `if model.config.solve.postprocessing_active` check here, along with the check for optimal termination condition, rather than inside `postprocess.postprocess_model_results`?

src/calliope/model.py
```diff
@@ -89,7 +88,7 @@ def read_yaml(
     )
 
 
-class Model:
+class Model(ModelStructure):
```
sjpfenninger, line 91: Why the introduction of `ModelStructure`?

sjpfenninger (changes requested):

src/calliope/model.py
irm-codebase: It's kind of a left-over from the plug-in approach I was trying to develop. I left it because I thought it was generally useful...

When expanding `model.build | solve` to a plug-in approach, knowing the structure lets you define processes that 'use' a model instance without inheriting `Model` directly (which often leads to cyclic imports).

```python
from calliope.external import builder, solver
from calliope.schemas import ModelStructure

@builder(mode="spores")
def build_spores(model: ModelStructure):
    ...

@solver(mode="spores")
def solve_spores(model: ModelStructure):
    ...

@builder(mode="operate")
def build_operate(model: ModelStructure):
   ...

```

These can later be called from `model.build` or `model.solve`, which would lead to a cyclic import issue if you do not follow this approach.

We were already seeing this problem for the MathDocs in the postprocessing side, so I kept it.

src/calliope/model.py
irm-codebase: It would be only if we are fine with not running the the `zero_threshold` stuff.
I kept it this way because I figured those should always run. Would you prefer me to change it?

sjpfenninger: Let's leave it as is for now!

PR #817, opened 2025-09-03
Postprocess: prevent exception when flow_cap lacks nodes dim

sjpfenninger:
This solves a rare edge case: when making a single-node model where operate mode math is added to the base math and `flow_cap` is defined directly in-YAML, `cap_sum` ends up having no "nodes" dimension, leading to an exception in postprocessing.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/postprocess/postprocess.py
+++ b/src/calliope/postprocess/postprocess.py
@@ -67,7 +67,9 @@ def capacity_factor(
             dim=["timesteps", "nodes"], min_count=1
         )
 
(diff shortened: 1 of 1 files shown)


irm-codebase (approved):
Looks good.
In the future we should be careful to ensure postprocesses do not fail in dimensionless models.

Either that, or we should ensure that we add at least one "dummy" entry in every dimension.

PR #819, opened 2025-09-04
fix equation eq operator

### LINKED ISSUE #818: Use `=` equality for mathematical expressions and `==` for logical comparisons
### What happened?

Leftover from #728

To neatly distinguish between what is math and what is logic, `where` should use `==` (#795) and `expression` should use `=`.

```yaml
  source_capacity_equals_flow_capacity:
    description: >-
      Set a `supply` technology's flow capacity to equal its source capacity.
    foreach: [nodes, techs, carriers]
    where: source_cap AND source_cap_equals_flow_cap==True  # correct, was '=' before #795 fix
    equations:
      - expression: source_cap == flow_cap  # this should be '='
```

- [ ] macOS
- [ ] Windows
- [x] Linux

### Version

_No response_

### Relevant log output

```Shell

```

irm-codebase:
Fixes #818

This PR completes the standardisation of where / expression strings into their respective domains

* mathematical operations ("equations") now use `=`, `<=` and `>=` operators
* logical operations ("where" strings) use `==`

## Reviewer checklist

- [ ] Documentation updated
- [x] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### User-facing changes
 
+|changed| `equations: - expression:` now uses `=` for 'equals' cases (#818)
+
 |fixed| `ModelDataFactory` is no longer run twice during startup, resulting in faster initialisation
 
 |changed| `Model.from_dict` has been removed in favor of `calliope.read_dict` (matching `read_netcdf` and `read_yaml`)
--- a/docs/user_defined_math/examples/chp_htp.yaml
+++ b/docs/user_defined_math/examples/chp_htp.yaml
@@ -114,9 +114,9 @@ constraints:
   balance_conversion:
     equations:
       - where: NOT turbine_type
-        expression: sum(flow_out_inc_eff, over=carriers) == sum(flow_in_inc_eff, over=carriers)
+        expression: sum(flow_out_inc_eff, over=carriers) = sum(flow_in_inc_eff, over=carriers)
       - where: turbine_type==backpressure AND NOT boiler_eff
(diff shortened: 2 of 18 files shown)


irm-codebase:
Here is a proposal for this functionality.
I'm not as versed as @brynpickering on backend functionalities, but for what I've seen this should do the trick?

# Proposal: `@` accessor for dimensions

`@` is a nice decorator for accessing dimensions because it's often spelled as 'at'.
It's functionality is to let a user access the current coordinate/indexes 'at' the specific `foreach` execution.

It should provide the following functionality.

## Rules

- If using `@dim`, that specific dimension should be present in `foreach:`. If not, an error should be raised.
- Only numbered dimensions should be usable in the math.
- Only ordered dimensions should have ordinal / offset accessors (see below)

## `dimension` accessors

For the cases below, assume:
- `year` is an ordered dimension with `[2000, 2005, 2010, 2015]`
- there are no `where` limits (all indexes are accessed, in order)

Accessors:
- **current coordinate** (`@year`): this will fetch the current case.
So `expression: @year + 1` should net `[2001, 2006, 2010, 2016]`
- **offset** (`@year[-1]`, `@year[+1]`): lets you fetch dimension values in relation to the current `foreach` position.
So `expression: @year[+1]` should net `[2005, 2010, 2015, ERROR]`
And `expression: @year[+1]` should net `[ERROR, 2000, 2005, 2010]`

## `variable` accessors

These let users access variables in relation to the current `foreach` position.
Similar to `roll` in most cases, but a bit more readable.

- `flow_cap@year[-1]`: give me `flow_cap` for the previous year.
- `flow_cap@year[+1]@timestep[-1]`: give me `flow_cap` at the previous year for the next timestep

PR #832, opened 2025-09-24
Remove _reentry; add runtime entries for timeseries agg/math updates

### LINKED ISSUE #824: _datetime_index() in time.py failing on ISO8601 check when reading models with clustering
### What happened?

I have successfully run dev7 calliope models with clustered days and saved these as netcdfs - no issues. However, later when attempting to read these files using calliope.read_netcdf, an exception is raised within preprocessing's time.py in the function _datetime_index(). 

I do not fully understand it but I believe there is some issue with the way in which clustering aspect is written into netcdf file, it is recording the time component twice perhaps? e.g. 00:00:00 00:00:00 instead of yyyy-mm-dd 00:00:00. I am surprised there is a time component at all considering clustering is on dates only.

I personally fixed this with by writing a wrapper for read_netcdf that applies .pop to remove 'time_cluster' from the config.init attributes, as per the function below:

import calliope
from pathlib import Path
from netCDF4 import Dataset
import  yaml

def read_clustered_netcdf(path):

    p = Path(path).resolve()
    with Dataset(p, "a") as nc:  # append mode
        g = nc.groups["attrs"]
        cfg = yaml.safe_load(g.getncattr("config"))
        cfg.get("init", {}).pop("time_cluster", None)  # remove lingering clustering
        g.setncattr("config", yaml.safe_dump(cfg))

    return calliope.read_netcdf(path)

Not ideal, as this loses the clustering information. I can provide a 'faulty' netcdf file for you to play around with in case that helps.

- [ ] macOS
- [x] Windows
- [ ] Linux

### Version

dev7

### Relevant log output

```Shell

```

brynpickering:
Fixes #824 

This should make re-init more robust by storing more things in `runtime` so we can compare against them when cleaning up the input dataset. This allows us to subset/resample/cluster only when there's a need to (runtime settings have changed).

* Removed `_reentry` and replaced with more config stored in `runtime`. This is more robust for cases of re-init where clustering/subsetting/resampling are defined and/or are being changed.
* Allowed updates to init config on loading from file. This would allow someone to store and share a model at full resolution and with lots of techs/nodes but for others to then easily subset/resample/cluster that model for their needs.
* Split `ModelDataFactory` into a builder and cleaner class. They share one method (dtype updater) which I have placed in a shared abstract base class.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/model.py
+++ b/src/calliope/model.py
@@ -15,7 +15,8 @@ import xarray as xr
 
 from calliope import _version, backend, exceptions, io, postprocess, preprocess
 from calliope.attrdict import AttrDict
-from calliope.preprocess.model_data import ModelDataFactory
+from calliope.preprocess import model_math
+from calliope.preprocess.model_data import ModelDataBuilder, ModelDataCleaner
 from calliope.schemas import CalliopeAttrs, ModelStructure, config_schema
 from calliope.util.logging import log_time
 
@@ -25,11 +26,12 @@ if TYPE_CHECKING:
 LOGGER = logging.getLogger(__name__)
 
 
-def read_netcdf(path: str | Path) -> Model:
+def read_netcdf(path: str | Path, **kwargs) -> Model:
     """Return a Model object reconstructed from model data in a NetCDF file.
 
     Args:
         path (str | Path): Path to Calliope model NetCDF file.
+        **kwargs: initialisation overrides.
 
     Returns:
         Model: Calliope Model instance.
@@ -39,6 +41,7 @@ def read_netcdf(path: str | Path) -> Model:
         model_data["inputs"],
         CalliopeAttrs(**model_data["attrs"].attrs),
         model_data["results"],
+        **kwargs,
     )
 
 
@@ -120,17 +123,25 @@ def read_dict(
         "preprocess_start",
         comment="Model: preprocessing data",
     )
-    model_data_factory = ModelDataFactory(
+    # TODO: build math on re-init, too.
+    math_priority = model_math.get_math_priority(model_def.config.init)
+    math = model_math.build_math(
+        math_priority,
+        model_def.math.init.model_dump(),
+        validate=model_def.config.init.pre_validate_math_strings,
+    )
+    model_data_factory = ModelDataBuilder(
         model_def.config.init,
         AttrDict(model_def.definition.model_dump(exclude_defaults=True)),
-        model_def.math.init,
+        math,
         definition_path,
         data_table_dfs,
     )
     model_data_factory.build()
-    model_data_factory.clean()
-    model_def = model_def.update({"math.build": model_data_factory.math.model_dump()})
-    return Model(inputs=model_data_factory.dataset, attrs=model_def, _reentry=False)
+    model_def = model_def.update(
+        {"math.build": math.model_dump(), "runtime.math_priority": math_priority}
+    )
+    return Model(inputs=model_data_factory.dataset, attrs=model_def)
 
 
 class Model(ModelStructure):
@@ -143,7 +154,6 @@ class Model(ModelStructure):
         inputs: xr.Dataset,
         attrs: CalliopeAttrs,
         results: xr.Dataset | None = None,
-        _reentry: bool = True,
         **kwargs,
     ) -> None:
         """Returns a instantiated Calliope Model.
@@ -154,10 +164,6 @@ class Model(ModelStructure):
             results (xr.Dataset | None, optional):
                 Results dataset from another Calliope Model with compatible math formulation.
                 Defaults to None.
-            _reentry (bool, optional):
-                Specifies model math and configuration must be reinitialised.
-                Should only be set to `False` if this is the first time the model has been instantiated.
-                Defaults to True.
             **kwargs:
                 initialisation keyword arguments
         """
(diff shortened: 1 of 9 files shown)


src/calliope/preprocess/model_data.py
```diff
@@ -733,62 +627,247 @@ def _update_one_way_links(node_from_data: dict, node_to_data: dict):
...
+    def add_link_distances(self):
+        """If latitude/longitude are provided but distances between nodes have not been computed, compute them now.
+
+        The schema will have already handled the fact that if one of lat/lon is provided, the other must also be provided.
+        """
+        # If no distance was given, we calculate it from coordinates
+        if (
+            "latitude" in self.dataset.data_vars
+            and "longitude" in self.dataset.data_vars
+            and (self.dataset.base_tech == "transmission").any()
```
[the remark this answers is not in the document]
brynpickering: This method is only called once so I think it's easier to leave there than to abstract it

brynpickering:
Failing tests are some randomly occurring infinite recursion error on Linux. Any ideas @irm-codebase ?

src/calliope/model.py
```diff
@@ -120,17 +123,30 @@ def read_dict(
         "preprocess_start",
         comment="Model: preprocessing data",
     )
-    model_data_factory = ModelDataFactory(
+    # TODO: build math on re-init, too.
```
irm-codebase: Left over TODO?
Math looks complete after function is done.

src/calliope/model.py
```diff
@@ -167,34 +178,38 @@ def __init__(
...
+                    "runtime.math_priority": math_priority,
+                    "math.build": math.model_dump(),
+                }
             )
-            self.config = self.config.update({"init": kwargs})
-            model_data_factory = ModelDataFactory(
-                self.config.init, inputs, self.math.init
+        if inputs:
+            model_data_factory = ModelDataCleaner(
+                self.config.init, inputs, attrs.math.build, attrs.runtime
```
irm-codebase, line 205: Ha. Splitting `ModelDataFactory` into a cleaning portion was an idea during the `_reentry` PR...
Shows I should not have been lazy, I suppose.

irm-codebase: @brynpickering can you describe in which instance `if inputs` will result to `False`?

All `read_` functions deliver a non-empty object, and creating a new model from another (direct `Model` call by the user) should also result in a non-empty `inputs`.

To me it looks like we should throw an error if `inputs` is not an `xarray` dataset, and remove this `if`...

src/calliope/preprocess/data_tables.py
```diff
@@ -67,21 +53,18 @@ def __init__(
         self.model_definition_path = model_definition_path
```
irm-codebase, line 53: Nice to have: consider defining the typing of `self.` class attributes here (i.e., `self.name: str = table_name`). That way `mypy` will throw errors if your logic does not comply with that typing.

src/calliope/preprocess/__init__.py
irm-codebase, line 1: Nitpick:
I'd suggest giving access to `ModelDataFactory` and `ModelDatacleaner` here.
It's a way of 'suggesting' that other functionality is not meant for external access, and it makes our imports shorter.

src/calliope/preprocess/model_data.py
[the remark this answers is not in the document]
irm-codebase: meh, ignore the bot

src/calliope/preprocess/model_data.py
```diff
@@ -733,62 +615,261 @@ def _update_one_way_links(node_from_data: dict, node_to_data: dict):
...
+    def assign_input_attr(self):
+        """Assign metadata as attributes to each input array."""
+        all_attrs = {
+            **self.math.parameters.model_dump(),
+            **self.math.lookups.model_dump(),
+        }
+        for var_name, var_data in self.dataset.data_vars.items():
+            self.dataset[var_name] = var_data.assign_attrs(all_attrs.get(var_name, {}))
+            # Remove this redundant attribute
+            self.dataset[var_name].attrs.pop("active", None)
```
irm-codebase, line 818: Perhaps a topic for a future PR, but it's best to ask now.

Do we need to store these in the resulting dataset? This data is always a repeat of what is in `math`, which is stable, safer, easier to read in code, and requires no duplication (which avoids future bugs).

irm-codebase:
Ok, after running the tests several times I did manage to reproduce the failed `test_backend_pyomo` case @brynpickering.
This is very hard to debug, but my best guess at the moment is this:

Something in Linux does not enjoy that the `stdout` and `stderr` are being redirected in the `pyomo` backend. Looks like this is the sequence:

- something in the logging chain becomes inaccessible
- a `stdout` / `stderr` message tries to go through
- cannot due to lack of access
- a `stdout` / `stderr` message tries to go through to specify this access issue
- cannot due to lack of access
- etc etc

This is a separate bug, and would be solved by not redirecting `std_err` / `std_out` in the pyomo backend.
Gut guess is that something in the test setup is polluting the model instance.

src/calliope/model.py
brynpickering: Inputs may be an empty dataset. This happens for instance in SPORES where results are saved but inputs are not. When loaded back in, this resolves to an empty dataset.

src/calliope/preprocess/data_tables.py
brynpickering: It will match the type of the associated method arguments so I don't think there is a need to re-iterate the type for mypy to pick that up, right?

src/calliope/preprocess/model_data.py
brynpickering: I tend to agree due to avoiding future bugs and repetition. The only reason to keep the information is for users to be able to quickly see the metadata associated with each array (be it a parameter, lookup, variable, or global expression)....

src/calliope/model.py
irm-codebase: So basically a `model` instance that cannot run `build`/`solve`?
I guess that is fine, although I personally find it odd... my not just keep a `netCDF` file in that case?

src/calliope/preprocess/model_data.py
irm-codebase: I think it'd be better to only have the `math` definition, then. We should slowly get users familiarised with the new approach, and not having this stuff duplicated is part of that. It also prevents duplication bugs (e.g., the backend using these values instead of `math`) from happening in the future

irm-codebase: That's a topic for another PR though.

irm-codebase (approved):
Changes look good to me.
The failing test we are seeing should be benign on the user side. I've created #833 to track it.

src/calliope/preprocess/data_tables.py
irm-codebase: https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#classes
You are correct!

PR #834, opened 2025-09-30
Update parsing to allow math and dims in where strings

### LINKED ISSUE #733: Conducting math operations using dimension data
### What can be improved?

Numeric dims should be accessible from inside math (and `where` arrays) to e.g. get the difference between two dimensions. E.g. getting the difference between `investsteps` and `vintagesteps`.

### Version

v0.7.0.dev5

brynpickering:
Fixes #733

* where strings can now accept arithmetic in helper functions and when making comparisons.
* valid component names are split by type (dims, inputs, vars&exprs) so that parsing can handle them separately. This allows us to limit what gets included in different parts of where strings (e.g. no vars/exprs in comparisons) and simplifies evaluation (no need to check what type it is as the parsing has already handled that). This is only applied in `where` strings. There is scope to extend it to `expression` strings but I haven't done so here.
* helper functions updated as now everything is passed around as an xarray dataarray, never a simple string. It's a slight overcomplication when all you're doing is referring to a string (e.g. `base_tech==transmission`) but it's the only way of handling situations when there is legitimate crossover between strings and components (`base_tech==storage` since storage is both a base tech and a decision variable). To handle this ambiguity, simple strings/numbers have both the value and array name assigned to the string/number (e.g. `da = xr.DataArray(2, name=2)`) so we can use it both in arithmetic (`flow_cap_max+2`) and in comparisons (`flow_cap_max==2`) without it acting weird (I hope...).
* Removed `default_if_empty` rather than fix it to match the changes made since we handle default values differently now.
* 
## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/backend/backend_model.py
+++ b/src/calliope/backend/backend_model.py
@@ -154,6 +154,9 @@ class BackendModelGenerator(ABC, metaclass=SelectiveWrappingMeta):
 
         self._add_to_dataset(name, values, "lookups", definition.model_dump())
 
+        if name not in self.math["lookups"]:
+            self.math = self.math.update({f"lookups.{name}": definition.model_dump()})
+
     @abstractmethod
     def add_parameter(
         self, name: str, values: xr.DataArray, definition: math_schema.Parameter
@@ -287,7 +290,9 @@ class BackendModelGenerator(ABC, metaclass=SelectiveWrappingMeta):
     def _check_inputs(self):
         data_checks = self.math.checks
         check_results = {"raise": [], "warn": []}
-        parser_ = parsing.where_parser.generate_where_string_parser()
+        parser_ = parsing.where_parser.generate_where_string_parser(
+            **self.valid_component_names
+        )
         eval_kwargs = {
             "backend_interface": self,
             "math": self.math,
@@ -372,7 +377,7 @@ class BackendModelGenerator(ABC, metaclass=SelectiveWrappingMeta):
         else:
             self._raise_error_on_preexistence(name, component_type)
             parsed_component = parsing.ParsedBackendComponent(
-                component_type, name, component_def
+                component_type, name, component_def, self.valid_component_names
             )
 
             top_level_where = parsed_component.generate_top_level_where_array(
@@ -386,7 +391,7 @@ class BackendModelGenerator(ABC, metaclass=SelectiveWrappingMeta):
 
             else:
                 self._create_obj_list(name, component_type)
-                equations = parsed_component.parse_equations(self.valid_component_names)
+                equations = parsed_component.parse_equations()
 
                 if not equations:
                     component_da = component_setter(
@@ -648,23 +653,34 @@ class BackendModelGenerator(ABC, metaclass=SelectiveWrappingMeta):
         return self._dataset.filter_by_attrs(obj_type="objectives")
 
     @property
-    def valid_component_names(self) -> set:
+    def valid_component_names(self) -> dict[str, set[str]]:
         """Return a set of valid component names in the model data.
 
         Returns:
             set: set of valid names.
         """
 
-        def _filter(val):
-            return val in ["variables", "parameters", "global_expressions", "lookups"]
+        def _active_names(group) -> set[str]:
+            return set(k for k, v in self.math[group].root.items() if v.active)
 
-        in_data = set(self._dataset.filter_by_attrs(obj_type=_filter).data_vars.keys())
-        in_math = set(
-            name
-            for component in ["variables", "global_expressions"]
-            for name in self.math[component].root
+        names = {
+            "dimension_names": _active_names("dimensions"),
+            "input_names": _active_names("parameters").union(_active_names("lookups")),
+            "var_expr_names": _active_names("variables").union(
+                _active_names("global_expressions")
+            ),
+        }
+        all_input_names = set(self.math.parameters.root).union(self.math.lookups.root)
+        all_var_expr_names = set(self.math.variables.root).union(
+            self.math.global_expressions.root
         )
-        return in_data.union(in_math)
+        inputs_to_add_back = all_input_names - names["input_names"]
+        var_exprs_to_add_back = (
+            all_var_expr_names - names["var_expr_names"] - all_input_names
+        )
+        names["input_names"].update(inputs_to_add_back)
+        names["var_expr_names"].update(var_exprs_to_add_back)
+        return names
 
 
 class BackendModel(BackendModelGenerator, Generic[T]):
@@ -711,8 +727,9 @@ class BackendModel(BackendModelGenerator, Generic[T]):
                     "piecewise_constraints",
                     name,
                     math_schema.GlobalExpression.model_validate(dummy_expression_dict),
+                    self.valid_component_names,
                 )
-                eq = parsed_component.parse_equations(self.valid_component_names)
+                eq = parsed_component.parse_equations()
                 expression_da = eq[0].evaluate_expression(
                     self, where=where, references=references
                 )
--- a/src/calliope/backend/expression_parser.py
+++ b/src/calliope/backend/expression_parser.py
@@ -102,7 +102,7 @@ class EvalToArrayStr(EvalString):
         """Evaluate and return expression as LaTeX."""
 
     @abstractmethod
-    def as_array(self) -> xr.DataArray | list[str | float]:
+    def as_array(self) -> xr.DataArray | list[xr.DataArray]:
         """Evaluate and return expression as a DataArray or list."""
 
     # Math strings evaluate to strings.
@@ -113,11 +113,11 @@ class EvalToArrayStr(EvalString):
     @overload
     def eval(
         self, return_type: Literal["array"], **eval_kwargs
-    ) -> xr.DataArray | list[str | float]: ...
+    ) -> xr.DataArray | list[xr.DataArray]: ...
 
     def eval(
         self, return_type: RETURN_T, **eval_kwargs
-    ) -> str | list[str | float] | xr.DataArray:
+    ) -> str | xr.DataArray | list[xr.DataArray]:
         """Evaluate math string expression.
 
         Args:
@@ -142,6 +142,7 @@ class EvalToArrayStr(EvalString):
                 If `array` is desired, returns xarray DataArray or a list of strings/numbers (if the expression represents a list).
         """
         self.eval_attrs = eval_kwargs
+        evaluated: str | list[str | float] | xr.DataArray | list[xr.DataArray]
         if return_type == "array":
             evaluated = self.as_array()
         elif return_type == "math_string":
@@ -240,7 +241,7 @@ class EvalOperatorOperand(EvalToArrayStr):
 
     def _apply_where_array(self, evaluated: xr.DataArray) -> xr.DataArray:
         """Util function to apply where arrays to non-latex strings."""
-        where_array = self.eval_attrs["where_array"]
+        where_array = self.eval_attrs.get("where_array", xr.DataArray(True))
         try:
             evaluated = evaluated.where(where_array)
         except AttributeError:
@@ -605,7 +606,10 @@ class EvalSlicedComponent(EvalToArrayStr):
     def _eval(self, return_type: RETURN_T) -> tuple[str | xr.DataArray, dict]:
         """Evaluate the slice dim and vals of each slice element."""
         slices: dict[str, Any] = {
-            k: v.eval(return_type, **self.eval_attrs) for k, v in self.slices.items()
+            k: xr.concat(slice_, dim=k)
+            if isinstance(slice_ := v.eval(return_type, **self.eval_attrs), list)
+            else slice_
+            for k, v in self.slices.items()
         }
 
         evaluated = self.obj_name.eval(return_type, **self.eval_attrs)
@@ -661,11 +665,11 @@ class EvalIndexSlice(EvalToArrayStr):
     @overload
     def _eval(
         self, return_type: Literal["array"], as_values: bool
-    ) -> xr.DataArray | list[str | float]: ...
+    ) -> xr.DataArray | list[xr.DataArray]: ...
 
     def _eval(
         self, return_type: RETURN_T, as_values: bool
-    ) -> str | xr.DataArray | list[str | float]:
+    ) -> str | xr.DataArray | list[xr.DataArray]:
         """Evaluate the referenced `slice`."""
         self.eval_attrs["as_values"] = as_values
         return self.eval_attrs["slice_dict"][self.name][0].eval(
@@ -675,7 +679,7 @@ class EvalIndexSlice(EvalToArrayStr):
     def as_math_string(self) -> str:  # noqa: D102, override
         return self._eval("math_string", False)
 
-    def as_array(self) -> xr.DataArray | list[str | float]:  # noqa: D102, override
+    def as_array(self) -> xr.DataArray | list[xr.DataArray]:  # noqa: D102, override
         evaluated = self._eval("array", True)
         return evaluated
 
@@ -754,7 +758,7 @@ class EvalNumber(EvalToArrayStr):
         )
 
     def as_array(self) -> xr.DataArray:  # noqa: D102, override
-        return xr.DataArray(float(self.value), attrs={"obj_type": "number"})
+        return xr.DataArray(float(self.value), name=float(self.value))
 
 
 class ListParser(EvalToArrayStr):
@@ -782,13 +786,13 @@ class ListParser(EvalToArrayStr):
 
     def as_math_string(self) -> str:  # noqa: D102, override
         input_list = self.as_array()
-        return "[" + ",".join(str(i) for i in input_list) + "]"
+        return "[" + ",".join(str(i.name) for i in input_list) + "]"
 
-    def as_array(self) -> list[str | float]:  # noqa: D102, override
+    def as_array(self) -> list[xr.DataArray]:  # noqa: D102, override
         values = [val.eval("array", **self.eval_attrs) for val in self.val]
         # strings and numbers are returned as xarray arrays of size 1,
         # so we extract those values.
-        return [da.item() if da.size == 1 else da.name for da in values]
+        return values
 
 
 class EvalUnslicedComponent(EvalToArrayStr):
@@ -821,10 +825,10 @@ class EvalUnslicedComponent(EvalToArrayStr):
         evaluated = self.as_array()
         self.eval_attrs["references"].add(self.name)
 
-        if evaluated.attrs["obj_type"] != "string":
+        if "math_repr" in evaluated.attrs:
             data_var_string = evaluated.attrs["math_repr"]
         else:
-            data_var_string = rf"\text{{{self.name}}}"
+            data_var_string = self.name
 
         return data_var_string
 
@@ -845,7 +849,7 @@ class EvalUnslicedComponent(EvalToArrayStr):
                 f"Trying to access a math component that is not yet defined: {self.name}. "
                 "If the referenced component is a global expression, set its `order` to have it defined first."
             )
-        if pd.notna(evaluated.attrs["default"]):
+        if evaluated.isnull().any() and pd.notna(evaluated.attrs["default"]):
             evaluated = evaluated.fillna(evaluated.attrs["default"])
 
         self.eval_attrs["references"].add(self.name)
@@ -879,7 +883,7 @@ class EvalGenericString(EvalToArrayStr):
         return str(self.val)
 
     def as_array(self) -> xr.DataArray:  # noqa: D102, override
-        return xr.DataArray(str(self.val), attrs={"obj_type": "string"})
+        return xr.DataArray(str(self.val), name=str(self.val))
 
 
 def helper_function_parser(
@@ -953,9 +957,8 @@ def helper_function_parser(
 
 
 def sliced_param_or_var_parser(
-    number: pp.ParserElement,
+    slicer: Iterable[pp.ParserElement],
     generic_identifier: pp.ParserElement,
-    evaluatable_identifier: pp.ParserElement,
     unsliced_object: pp.ParserElement,
     allow_slice_references: bool = True,
 ) -> pp.ParserElement:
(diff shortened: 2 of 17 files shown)


src/calliope/backend/backend_model.py
```diff
@@ -154,6 +154,9 @@ def add_lookup(
 
         self._add_to_dataset(name, values, "lookups", definition.model_dump())
 
+        if name not in self.math["lookups"]:
+            self.math = self.math.update({f"lookups.{name}": definition.model_dump()})
+
```
brynpickering, line 159: new math definitions should be added to the build math no matter whether they're related to inputs or math components. This was an oversight from the previous PR

src/calliope/backend/backend_model.py
```diff
@@ -648,23 +653,34 @@ def objectives(self):
...
+                _active_names("global_expressions")
+            ),
+        }
+        all_input_names = set(self.math.parameters.root).union(self.math.lookups.root)
+        all_var_expr_names = set(self.math.variables.root).union(
+            self.math.global_expressions.root
+        )
+        inputs_to_add_back = all_input_names - names["input_names"]
+        var_exprs_to_add_back = (
+            all_var_expr_names - names["var_expr_names"] - all_input_names
```
brynpickering: Process is now as follows:

1. get all active components
2. get all inactive inputs that are not then reactivated in vars/expressions. These are also included as "valid" so that the `where` string parsing will pass (and will resolve to `False` in the `where` string and to its default value in the `expression` string)
3. get all inactive vars/exprs that are not then reactivated as inputs, for the same reason as (2).

src/calliope/backend/expression_parser.py
```diff
@@ -240,7 +241,7 @@ def _operator_operands(
 
     def _apply_where_array(self, evaluated: xr.DataArray) -> xr.DataArray:
         """Util function to apply where arrays to non-latex strings."""
-        where_array = self.eval_attrs["where_array"]
+        where_array = self.eval_attrs.get("where_array", xr.DataArray(True))
```
brynpickering: Now we have `where` parsers using this evaluator but we don't define `where_array` for them (they are busy trying to generate it!), so this is a default for it. An alternative would be to set that as the value for all `where` parsers to have in their `eval_attrs` so that we can revert to `self.eval_attrs["where_array"]` here.

src/calliope/backend/expression_parser.py
```diff
@@ -605,7 +606,10 @@ def _eval(self, return_type: Literal["array"]) -> tuple[xr.DataArray, dict]: ...
     def _eval(self, return_type: RETURN_T) -> tuple[str | xr.DataArray, dict]:
         """Evaluate the slice dim and vals of each slice element."""
         slices: dict[str, Any] = {
-            k: v.eval(return_type, **self.eval_attrs) for k, v in self.slices.items()
+            k: xr.concat(slice_, dim=k)
+            if isinstance(slice_ := v.eval(return_type, **self.eval_attrs), list)
+            else slice_
+            for k, v in self.slices.items()
```
brynpickering, line 573: This one was fun. `nodes=[a, b]` now resolves to [`xr.DataArray("a"), xr.DataArray("b")]` but we can't slice with that. What we _can_ slice with is `xr.DataArray(["a", "b"], name="nodes")`, which is what I now do here if it receives a list.

src/calliope/backend/expression_parser.py
```diff
@@ -821,10 +825,10 @@ def as_math_string(self) -> str:  # noqa: D102, override
         evaluated = self.as_array()
         self.eval_attrs["references"].add(self.name)
 
-        if evaluated.attrs["obj_type"] != "string":
+        if "math_repr" in evaluated.attrs:
             data_var_string = evaluated.attrs["math_repr"]
         else:
-            data_var_string = rf"\text{{{self.name}}}"
+            data_var_string = self.name
```
brynpickering, line 792: dimensions have no math representation and we don't want them to be `\text{...}` as we need their plain string values to do fun things with them in the helper functions (like get their associated iterator from the math).

src/calliope/backend/expression_parser.py
```diff
@@ -953,9 +957,8 @@ def helper_function_parser(
 
 
 def sliced_param_or_var_parser(
-    number: pp.ParserElement,
+    slicer: Iterable[pp.ParserElement],
```
brynpickering, line 922: Generally trying to move to this more generalised approach of defining iterables that go into `pp.MatchFirst` rather than requiring them all to be defined explicitly and then doing `option_1 | option_2 | ...`.

src/calliope/backend/expression_parser.py
```diff
@@ -1200,12 +1197,12 @@ def generate_slice_parser(valid_component_names: Iterable) -> pp.ParserElement:
     evaluatable_identifier = evaluatable_identifier_parser(
         identifier, valid_component_names
     )
-    id_list = list_parser(number, evaluatable_identifier)
     unsliced_param = unsliced_object_parser(valid_component_names)
+    helper_func_list = list_parser(number, evaluatable_identifier, unsliced_param)
```
brynpickering, line 1164: helper functions can accept lists with all sorts in, plain strings, plain numbers, dimensions, and even entire input/math arrays. It is up to the helper function itself to deal with these as necessary so we let them all through. Subsetting lists `flow_cap[techs=[A, B]]` however, cannot be so free, so we have two different list parsers now.

src/calliope/backend/helper_functions.py
```diff
@@ -188,7 +186,9 @@ def _instr(self, dim: str) -> str:
         iterator = self._dim_iterator(dim)
         return rf"\text{{{iterator}}} \in \text{{{dim}}}"
 
-    def _listify(self, vals: list[str] | str) -> list[str]:
+    def _listify(
+        self, vals: list[str | xr.DataArray] | str | xr.DataArray
+    ) -> list[str]:
```
brynpickering: To all intents and purposes, `xr.DataArray("foo")` will be considered by Python as `"foo"`. [`xr.DataArray("foo")`, `xr.DataArray("bar")`] will not be considered as `["foo", "bar"]`, though, so we handle that here.

brynpickering: Maybe the func should be `_stringify_and_listify` or there should be a separate `_stringify` that accepts a list of mixed types and returns a string list. I just don't like `_stringify(_listify(...))`

src/calliope/backend/helper_functions.py
```diff
@@ -237,26 +240,15 @@ def as_array(self, input_component: str, *, over: str | list[str]) -> xr.DataArr
         Returns:
             xr.DataArray: resulting array.
         """
-        if input_component in self._input_data.data_vars:
-            component_da = self._input_data[input_component]
-            bool_component_da = (
-                component_da.notnull()
-                & (component_da != np.inf)
-                & (component_da != -np.inf)
+        if isinstance(input_component, str):
```
brynpickering: This is all now dealt with before we reach here, by the `where` parsing evaluator.

src/calliope/backend/backend_model.py
irm-codebase: Good catch. This also should add it to the `model.math` object since we no longer do deep copies of it, so no sync issues there.

src/calliope/backend/backend_model.py
```diff
@@ -648,23 +653,34 @@ def objectives(self):
         return self._dataset.filter_by_attrs(obj_type="objectives")
 
     @property
-    def valid_component_names(self) -> set:
+    def valid_component_names(self) -> dict[str, set[str]]:
         """Return a set of valid component names in the model data.
 
         Returns:
             set: set of valid names.
         """
```
irm-codebase: docstring needs updating
```suggestion
       Returns:
           dict[str, set[str]]: dictionary with set of valid names per component grouping.
       """
```

src/calliope/backend/backend_model.py
```diff
@@ -648,23 +653,34 @@ def objectives(self):
...
 
-        def _filter(val):
-            return val in ["variables", "parameters", "global_expressions", "lookups"]
+        def _active_names(group) -> set[str]:
+            return set(k for k, v in self.math[group].root.items() if v.active)
```
irm-codebase: Nitpick:
It's fine to have this here, but this strikes me as something that might be worth having in the `CalliopeBuildMath` level. This function is only here as a crutch for something that class should be providing.

src/calliope/backend/backend_model.py
irm-codebase: I think this process is fine... except for resolving to `default` when inactive. Wouldn't it make more sense to resolve to `null` in those cases, as a standard? My main worry here is non-zero defaults creeping in unexpectedly: something that is inactive should not have variant behaviour in the logic.

If the default is always `null`, `where` checks should still resolve to `False` (I think...), and references to an inactive component without conditionals would fail (which _is good_, as it indicates faulty logic).

This means that some conditionals (such as the `config.ensure_feasibility` checks in this PR https://github.com/calliope-project/calliope/pull/836) would need to stay, but it is ultimately a more stable approach IMO.

src/calliope/backend/expression_parser.py
irm-codebase: I'm partial to the second option as long as it does not cause trouble.
For what I can see, we sort of do that already in other sections (see `TRUE_ARRAY = xr.DataArray(True)` in `parsing.py`, although that's a default optional function input).

The closer this kind of thing is to the `__init__` of these classes, the safer and easier to maintain it will be since it's harder to miss specifics like this.

src/calliope/backend/expression_parser.py
```diff
@@ -754,7 +758,7 @@ def as_math_string(self) -> str:  # noqa: D102, override
         )
 
     def as_array(self) -> xr.DataArray:  # noqa: D102, override
-        return xr.DataArray(float(self.value), attrs={"obj_type": "number"})
+        return xr.DataArray(float(self.value), name=float(self.value))
```
irm-codebase, line 723: I assume this is the `da = xr.DataArray(2, name=2)` case.
Perhaps describe this case a bit in the docstring? It's quite specific, so it merits a reminder for the future.

src/calliope/backend/expression_parser.py
irm-codebase: Can you clarify further? As far as I understand, this will only take relevance in LaTeX backend stuff, right? How or why would a helper function need anything beyond the `\text{...}` case?

src/calliope/backend/expression_parser.py
```diff
@@ -1080,23 +1081,19 @@ def evaluatable_identifier_parser(
...
     """Parse strings which define a list of other strings or numbers.
 
     Lists are defined as anything wrapped in square brackets (`[]`).
 
-    Strings that could be evaluated as model component arrays will still be evaluated as strings (using the model component name).
-
-    Args:
-        number (pp.ParserElement): Parser for numbers (integer, float, scientific notation, "inf"/".inf").
-        evaluatable_identifier (pp.ParserElement): Parser for valid python variables without leading underscore and not called "inf".
+    Args (pp.ParserElement):
```
irm-codebase: Nitpick:
variable length should be specified in the docstring in the following way in the Google standard:

```
    Args:
        foo (int): something.
        bar (:obj:`str`, optional): Something else. Defaults to None.
        *args: Variable length argument list.
        **kwargs: Arbitrary keyword arguments.
```

src/calliope/backend/expression_parser.py
irm-codebase: Nice to see that helper functions now use a duck typing approach!

src/calliope/backend/helper_functions.py
irm-codebase: Eh, it's fine as-is I think. 
If you wish to change it, I have to admit I am not a fan of super long names, so how about `_to_str_list`?

src/calliope/backend/parsing.py
```diff
@@ -513,7 +514,7 @@ def parse_equations(
         slice_dict = {
             idx_name: self.generate_expression_list(
                 expression_parser=expression_parser.generate_slice_parser(
-                    valid_component_names
+                    set().union(*self._valid_component_names.values())
                 ),
```
irm-codebase, line 567: Nitpick on efficiency:
I suspect that our current use of `_valid_component_names` might be having some slight detrimental effects since it's constructed on-call at each component addition (`_add_component)... This object is generally unchanged in most instances, so we could save some computing power with better re-arranging and a single call for construction.

src/calliope/backend/where_parser.py
```diff
@@ -134,11 +138,11 @@ def as_array(self) -> xr.DataArray:  # noqa: D102, override
             return xr.DataArray(config_val)
 
 
-class DataVarParser(EvalWhere):
-    """Data variable processing."""
+class VarExprArrayParser(EvalWhere):
+    """Variable/Expression array processing."""
```
irm-codebase: Would `ResultArrayParser` make sense?
It's basically what the combination of variables and expressions is, and it would match the `inputs` / `results` setup at the `Model` level

src/calliope/backend/where_parser.py
```diff
@@ -148,104 +152,119 @@ def __init__(self, instring: str, loc: int, tokens: pp.ParseResults) -> None:
...
+        return self.array_name
+
+    def as_array(self) -> xr.DataArray:  # noqa: D102, override
+        if (name := self.array_name) in self.eval_attrs["input_data"]:
+            da = self.eval_attrs["input_data"][name]
+        else:
+            # We want the where string to evaluate successfully even if a dimension hasn't been defined.
+            da = xr.DataArray().astype(
+                DTYPE_OPTIONS[self.eval_attrs["math"].dimensions[name].dtype]
+            )
```
irm-codebase: Any particular reason for not returning something like `xr.DataArray(np.nan)`?

Running this command gives quite different values for the `DTYPES` we have, with `bool` being `True`, for example. I just want to ensure this will result in stable behaviour.

Image

irm-codebase (changes requested):
A few docstring comments, and some questions regarding decisions in the logic.

Overall it looks great. I particularly appreciate the improved friendliness of some of the parsing functionality.

src/calliope/backend/expression_parser.py
brynpickering: Yes, only for the LaTeX backend. It needs the plain string so it can look it up in the math to find the iterator for the dimension. `\text{...}` will have wrapped the string so we'd then have to know to search inside the `\text` for the corresponding dimension name which I tend to think is more error prone

src/calliope/backend/expression_parser.py
brynpickering: Updated to set default dict values initially and then to expect those entries to exist in the eval attrs.

src/calliope/backend/parsing.py
brynpickering: I'm not sure there's a way to re-arrange since adding a new component might be updating the math and hence changing the valid names available. This is a set union that's applied ~30 times so I doubt we'd see more than a few microseconds of impact

src/calliope/backend/where_parser.py
brynpickering: hmm, fair. I might need to limit it to switching between np.nan and NaT as this was caused by datetime dtypes not being happy with seeing a float dtype default array.

brynpickering: OK, on re-evaluating, I think this is a non-issue and I've simplified to np.nan throughout. I think it's a legacy from a prior iteration.

src/calliope/backend/expression_parser.py
brynpickering: Updated main `as_array` docstring

src/calliope/backend/expression_parser.py
brynpickering: Updated

src/calliope/backend/parsing.py
brynpickering: Moved into the math schema instead. I'm not sure it will have improved efficiency as it will still be evaluated on being called but it seems a neater approach.

src/calliope/backend/backend_model.py
brynpickering: I've updated it to fail on inactive. It doesn't resolve to `null` but rather the `expression` string parser fails as we don't pass any inactive math components to it as valid names.

brynpickering: It's not perfect as it's not the only opportunity for it to fail in such a case. E.g., an inactive component could be set in a variable bound or as one of the entries to piecewise constraints. This now is a null/completely missing array and both will fail for non-parsing reasons (because we don't use the parser for either). It's probably fine for now. The key thing is that the backend build fails if a user refers to a component that they've explicitly deactivated. The error message just isn't tidy as we can't catch all references to that case in one place.

src/calliope/backend/backend_model.py
brynpickering: Moved to math as `_active` property

src/calliope/backend/backend_model.py
irm-codebase: This seems ok to me.

Later, one thing we could try is finding a spot where all these possible issues 'converge',
At this spot we could add a `try` / `except` with an error that warns users about this and returns the causing error too.

src/calliope/backend/backend_model.py
```diff
@@ -150,7 +150,7 @@ def add_lookup(
 
         if values.isnull().all():
             self.log("lookups", name, "Component not added; no data found in array.")
-            values = values.astype(float)
+            values = xr.DataArray(np.nan, attrs=values.attrs)
```
irm-codebase, line 154: Correct me if I'm confused, but I thought another PR removed adding attributes to the `xarray` bits.
Just something to keep in mind when merging things.

src/calliope/schemas/math_schema.py
irm-codebase, line 1: The `@property` approach is fine, but it is slow (they are still functions, so they generate a call stack, which wastes time / memory, even if it's a tiny bit).

For these, we want to skip validation, and short code. We can use [`ClassVar`](https://docs.pydantic.dev/latest/concepts/models/#class-variables). Consider using `_group: ClassVar[Literal["lookups"]] = "lookups"` instead (`Literal` is just here to help `mypy` find errors).

src/calliope/schemas/math_schema.py
```diff
@@ -390,30 +438,69 @@ def unique_component_names(self):
...
+        """
+        names: dict[
+            Literal["dimension_names", "input_names", "result_names"], set[str]
+        ] = {
+            "dimension_names": set(self.dimensions._active),
+            "input_names": set(self.parameters._active).union(self.lookups._active),
+            "result_names": set(self.variables._active).union(
+                self.global_expressions._active
+            ),
+        }
```
irm-codebase: Some nitpicks.
- We should avoid the repeated `Literal`
- Consider keeping the component names and letting the caller decide on how to combine them. This file is generally agnostic to what these names mean (for now it's a model / backend thing). Alternatively, define what these mean here to enable re-use.
- `_name` seems to only be here to keep `generate_where_string_parser` happy (due to its argument names). I'd stick to the naming used at this level, and tweak the arguments of that function to fit it instead.
- I suggest the returned type `dict[str, set[str]]`. `Literal` forces `mypy` to check for exact strings, leading to false negatives (see attached image).

### Suggested code

```python
INPUT_COMPONENTS_T = Literal["lookups", "parameters"]
RESULT_COMPONENTS_T = Literal["global_expressions", "variables"]
WHERE_COMPONENTS_T = Literal["dimensions", INPUT_COMPONENTS_T, RESULT_COMPONENTS_T]
# Above could also be Literal["dimensions", "inputs", "results"]
COMPONENTS_T = Literal["constraints", "objectives", "piecewise_constraints", WHERE_COMPONENTS_T]
# Above could also be Literal["constraints", "objectives", "piecewise_constraints", INPUT_COMPONENTS_, RESULT_COMPONENTS_T]
```


### Suggested code

```suggestion
    @property
    def where_components(
        self,
    ) -> dict[str, set[str]]:
        """Return a set of valid component names in the model to use in `where` string parsing.

        Returns:
            dict[str, set[str]]:
                Set of valid names grouped by location in the math in which they are defined.
        """
        # if you wish to return 'inputs' and 'results', you'd need to split this a bit
        names: dict[set, set[str]] = { k: set(getattr(self, k)._active) for k in typing.get_args(WHERE_COMPONENTS_T)}
```

### Typing bug due to `-> Literal`

Image

src/calliope/backend/expression_parser.py
```diff
@@ -141,6 +146,7 @@ def eval(
                 If `math_string` is desired, returns a valid LaTex math string.
                 If `array` is desired, returns xarray DataArray or a list of strings/numbers (if the expression represents a list).
         """
+        eval_kwargs.setdefault("as_values", False)
         self.eval_attrs = eval_kwargs
```
irm-codebase: Nitpick:
Seems like we are reaching the limits of `TypedDict` here.
Later down the line, it'd be safer to change `EvalAttrs` to a [`dataclass`](https://docs.python.org/3/library/dataclasses.html) with defaults (mostly `None`).

```python
@dataclass
class EvalAttrs:
    ...
    as_values: bool = False

...

self.eval_attrs = EvalKwargs(**kwargs)
```

src/calliope/schemas/math_schema.py
```diff
@@ -390,30 +438,69 @@ def unique_component_names(self):
...
+                    "dimensions",
+                    "parameters",
+                    "lookups",
+                    "variables",
+                    "global_expressions",
+                ]
+            ]
+        )
+
+        return names
```
irm-codebase: Nitpick:
The returned objects of `where_components` and `expression_components` differ even though they have similar names and access... I understand why (forcing the caller to 'merge' all keys on their end is verbose and perhaps a bit inefficient), but it is a tad unintuitive.

src/calliope/backend/backend_model.py
brynpickering: An extra layer could be added even to these other inputs that requires them to go via the parser (after all, parsing `flow_cap_max` will return the `flow_cap_max` array). I don't know how much overhead there would be there, but at least we'd be funnelling everything through the same step that we can then catch exceptions in.

src/calliope/backend/backend_model.py
brynpickering: They keep some attrs internal to the backend (their math representation and any cross-references to other components used in recreating components on updates). Attaching them to the array attrs still seems the easiest way to keep everything together.

src/calliope/backend/expression_parser.py
brynpickering: updated to a dataclass

src/calliope/schemas/math_schema.py
brynpickering: Updated to a single output. They are both the same structure now. We can potentially leverage this in future to also parse them differently in the equation expression parser but for now they are just collapsed into a single set at the point of being passed to the parser.

src/calliope/schemas/math_schema.py
brynpickering: Updated. Also updated to `@cached_property` where possible, which effectively adds an attribute to the class, which will therefore not be run on each call. I can't do it for placeholder attrs inside math components as that gets picked up by pydantic as a field, which then causes issues with extra fields being forbidden.

irm-codebase (approved):
Looking great!

PR #835, opened 2025-10-03
Add cost source use & reorganise annual investment cost expression

### LINKED ISSUE #804: cost_operation_fixed is ignored if no investment costs are set
### What happened?

`cost_operation_fixed` is not triggered for technologies without investment costs. `cost_om_annual` should be applied regardless of the investment costs (e.x. already-built technologies).
```
  cost_operation_fixed:
    title: Total fixed operation costs
    description: >-
      The fixed, annual operation costs of a technology, which are calculated relative to investment costs.
      If the modeling period is not equal to one full year, this will be scaled accordingly.
    default: 0
    unit: cost
    foreach: [nodes, techs, costs]
    where: cost_investment AND (cost_om_annual OR cost_om_annual_investment_fraction)
    equations:
      - expression: >-
          $annualisation_weight * (
            sum(cost_om_annual * flow_cap, over=carriers) +
            cost_investment * cost_om_annual_investment_fraction
          )
    sub_expressions:
      annualisation_weight: *annualisation_weight
```
The issue seems to be in the `where` clause which requires `cost_investment` set.
`where: cost_om_annual OR (cost_investment AND cost_om_annual_investment_fraction)`
should enable fixed O&M costs without investment costs.

- [ ] macOS
- [x] Windows
- [x] Linux

### Version

v0.0.7.dev6

### Relevant log output

```Shell

```

### LINKED ISSUE #831: cost_flow_in doesn't work with supply
### What happened?

`cost_flow_in` fails when used with supply techs if the parameter is in use elsewhere in a non-supply tech. The other techs index the parameter over `carriers` which causes an expression parsing error when supply tries to use it without a `carrier` dimension.

```
calliope.exceptions.BackendError: global_expressions:cost_operation_variable:0-cost_flow_out:0-cost_flow_in:0-cost_export:1 | The linear expression array is indexed over dimensions not present in `foreach`: {'carriers'}
```

The specific issue is in the `cost_flow_in` sub-expression in `cost_operation_variable`

```
cost_operation_variable:
    title: Variable operating costs
    description: >-
      The operating costs per timestep of a technology.
    default: 0
    unit: $\frac{\text{cost}}{\text{hour}}$
    foreach: [nodes, techs, costs, timesteps]
    where: "cost_export OR cost_flow_in OR cost_flow_out"
    equations:
      - expression: timestep_weights * ($cost_export + $cost_flow_out + $cost_flow_in)
    sub_expressions:
      cost_export:
        - where: any(carrier_export, over=carriers) AND any(cost_export, over=carriers)
          expression: sum(cost_export * flow_export, over=carriers)
        - where: NOT (any(carrier_export, over=carriers) AND any(cost_export, over=carriers))
          expression: sum(0, over=carriers)
      cost_flow_in:
        - where: "base_tech==supply"
          expression: cost_flow_in * source_use
        - where: "NOT base_tech==supply"
          expression: sum(cost_flow_in * flow_in, over=carriers)
      cost_flow_out:
        - expression: sum(cost_flow_out * flow_out, over=carriers)
```
The error appears if `cost_flow_in` is set on any other tech, even if it isn't set on any supply techs. Setting `cost_flow_out` or `cost_export` on a supply will trigger the expression parsing and encounter the dimension issue.

It seems like we should either assume `cost_flow_in` will be indexed over carrier on supply techs and adjust the math to sum over carriers, or more likely add a `cost_source_use` that won't be used by other base_techs and can safely be set without a carrier index in supply. 

- [ ] macOS
- [x] Windows
- [x] Linux

### Version

v0.7.0.dev7

### Relevant log output

```Shell

```

brynpickering:
Fixes #831 
Fixes #804 

* use a carrier-agnostic cost for source use of supply technologies, to avoid clashes when indexing `cost_flow_in` over carriers _and_ assigning a value for it to supply technologies
* update where string of `cost_operation_fixed` to build the expression for either instance of it being needed
* fix units referring to cost from `\{cost}` to `\text{cost}`

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/math/base.yaml
+++ b/src/calliope/math/base.yaml
@@ -84,7 +84,7 @@ parameters:
     description: >-
       Cost per unit of the decision variable `flow_cap` and per unit distance of a transmission link.
       Applied to transmission links only.
-    unit: $\frac{\{cost}}{\text{power}\times\text{distance}}$
+    unit: $\frac{\text{cost}}{\text{power}\times\text{distance}}$
 
   cost_purchase_per_distance:
     default: 0
@@ -92,14 +92,14 @@ parameters:
     description: >-
       Cost applied if the binary variable `purchased` is 1 or per unit of the integer variable `units`.
       Requires the parameter `cap_method` to be `integer`.
-    unit: $\frac{\{cost}}{\text{unit}\times\text{distance}}$
+    unit: $\frac{\text{cost}}{\text{unit}\times\text{distance}}$
 
   cost_flow_cap:
     default: 0
     title: Cost of flow capacity.
     description: >-
       Cost per unit of the decision variable `flow_cap`.
-    unit: $\frac{\{cost}}{\text{power}}$
+    unit: $\frac{\text{cost}}{\text{power}}$
 
   cost_export:
     default: 0
@@ -108,7 +108,7 @@ parameters:
     description: >-
       Cost per unit of `flow_export` in each timestep.
       Usually used in the negative sense, as a subsidy.
-    unit: $\frac{\{cost}}{\text{energy}}$
+    unit: $\frac{\text{cost}}{\text{energy}}$
 
   cost_interest_rate:
     default: 0
@@ -123,7 +123,7 @@ parameters:
     description: >-
       Annual costs applied per unit `flow_cap`.
       These costs are not subject to being recalculated relative to technology lifetime, only scaled to reflect the fraction of one year that the model represents (e.g., 7 days ~= 0.02 of a year).
-    unit: $\frac{\{cost}}{\text{power}}$
+    unit: $\frac{\text{cost}}{\text{power}}$
 
   cost_om_annual_investment_fraction:
     default: 0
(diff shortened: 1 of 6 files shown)


src/calliope/math/base.yaml
irm-codebase, line 1220: It's an edge case of an edge case, but I'm wondering if `base_tech==sink` won't have a similar issue when it comes to `cost_flow_out`. It's the counterpart to `supply`, no?

irm-codebase (commented):
Sensible update. I like the separation of `cost_flow_in` and `cost_source_use`.
I've left a comment on a similar case I think might lead to similar trouble. After a reply I'll approve 👍

src/calliope/math/base.yaml
brynpickering: There isn't an equivalent for demand as we don't track `sink_use` in the same we do with `source_use`. Still, it is an edge case worth catching. I've added a check.

PR #836, opened 2025-10-03
Simplify math conditionals

brynpickering:
Now that we have default values for all math components, we can do away with a number of conditionals in our math expressions. Instead, the missing values will be replaced with their default values. This will lead to a minor speed up for math parsing and should reduce the complexity of the latex representation of our base math.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,18 @@
-## 0.7.0.dev8 (TBD)
+## 0.7.0.dev8
 
 ### User-facing changes
+
+|changed| Init config can be updated on re-init _and_ on loading a model from NetCDF, e.g. to apply new subsetting, resampling, clustering, math.
+
+|fixed| Reloading a clustered model from file was attempting to re-cluster (#824).
 
 |fixed| Example notebook outputs no longer clutter search results in docs and search result example notebook pages are grouped under the top-level page header (#675).
 
 ### Internal changes
 
 |changed| Math conditionals removed where math component defaults will achieve the same result.
+
+|fixed| Remove `stderr` redirect at pyomo backend model solve stage to avoid random infinite recursion errors in CI tests (#833).
 
 ## 0.7.0.dev7 (2025-09-05)
 
--- a/src/calliope/backend/pyomo_backend_model.py
+++ b/src/calliope/backend/pyomo_backend_model.py
@@ -8,7 +8,7 @@ import logging
 import re
 from abc import ABC
 from collections.abc import Iterable
-from contextlib import redirect_stderr, redirect_stdout
+from contextlib import redirect_stdout
 from pathlib import Path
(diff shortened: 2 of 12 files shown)


docs/user_defined_math/examples/fuel_dist.yaml
```diff
@@ -133,27 +130,12 @@ objectives:
...
-          expression: "0"
+      - expression: >-
+          sum(
+            (
+              sum(cost, over=[nodes, techs]) +
+              sum(cost_var_fuel_distribution, over=[nodes, carriers, timesteps])
+            ) *
+            objective_cost_weights,
+            over=costs
+          ) + $unmet_demand
```
irm-codebase, line 141: Quick glance at this suggests that a failure would occur if a cost with a `null` default is defined, right?

I think that's fine, in any case.

src/calliope/math/spores.yaml
irm-codebase, line 1: Could you walk me through how `unmet_demand` and `unused_supply` will act in the new case?

I know these variables have a `where: config.ensure_fesibility == True` ... but I'd expect this to mean that they only exist _if_ that is true (and I suspect whoever wrote these `sub_expressions` did the same).
What's preventing a crash here?

irm-codebase (commented):
Looks good, but I have some questions before approving.

docs/user_defined_math/examples/fuel_dist.yaml
brynpickering: A `null` default and _no_ costs defined anywhere in the model. The summations ignore nans so even a single value in the model will resolve to that value rather than nan at the end

src/calliope/math/spores.yaml
brynpickering: I wrote that, but before we had a robust mechanism for defining and using default values. Now, so long as a math component has been defined in the math it will be replaced with its default value when referred to but not active in the model. So, here both will be replaced with a non-dimensional zero value when `ensure_feasibility` is false. It helps with maintainability as we only have to set that conditional once (when defining the variable) and not every time it gets used

irm-codebase: Understood.
Basically, default is 'the value that will be put whenever referenced and not active', as well as 'the value that will be pre-filled in when active and not redefined'.

I guess my mind saw both as different, with the first being a 'skip' and the second leading to actual data, if that makes sense.

irm-codebase: The only risk I see in this approach is that setting default to `1` and then `active: False` will lead to data still being put in, which is quite counter intuitive...

brynpickering: Updated in #834 to fail on an explicitly deactivated component being referred to in an equation expression that has a `where` string which resolves to `True` at some point.

PR #837, opened 2025-10-03
Remove stderr redirect in pyomo backend

### LINKED ISSUE #833: Logger randomly gets into a racing condition.
### What happened?

When running tests, you may randomly encounter the following error:

```
FAILED tests/test_backend_pyomo.py::TestLogging::test_no_duplicate_log_message - assert 0 == 1
 +  where 0 = sum([False, False, False, False, False])
= 1 failed, 2660 passed, 113 skipped, 28 xfailed, 6 warnings in 500.58s (0:08:20) =
```

Which is caused by a recursive logging error, seen in this text file [recursive error.txt](https://github.com/user-attachments/files/22573940/recursive.error.txt).

## The culprit

`pyomo_backend_model.py` redirects regular `stderr` and `stdout`.

```python
        with redirect_stdout(LogWriter(self._solve_logger, "debug", strip=True)):  # type: ignore
            with redirect_stderr(LogWriter(self._solve_logger, "error", strip=True)):  # type: ignore
                # Ignore most of gurobipy's logging, as it's output is
                # already captured through STDOUT
                logging.getLogger("gurobipy").setLevel(logging.ERROR)
                results = opt.solve(self._instance, tee=True, **solve_kwargs)
```

In some cases (e.g., trying to use `caplog` in `pytest`), this may result in a racing condition because the logger is no longer accessible, but `stdout` is still re-directed to it using `LogWritter.write`:

```python
    def write(self, message):
        """Save a message to the logger."""
        if message != "\n":
            if self.strip:
                message = message.strip()
            getattr(self.logger, self.level)(message)  # <<- this line right here
```

When this occurs, `Logger` tries to write an error to `stderr`... which is sent right back to the logger, creating an infinite loop!

## Suggested fix

Not redirecting `stdout` and `stderr` at all in the `pyomo` backend?

- [ ] macOS
- [ ] Windows
- [x] Linux

### Version

v0.7

brynpickering:
Fixes #833 

Currently testing a fix. We need the redirect as otherwise we have no control over the solver outputs (the log to stdout). I'm hoping removing the double context will help.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,16 @@
-## 0.7.0.dev8 (TBD)
+## 0.7.0.dev8
+
+### User-facing changes
+
+|changed| Init config can be updated on re-init _and_ on loading a model from NetCDF, e.g. to apply new subsetting, resampling, clustering, math.
+
+|fixed| Reloading a clustered model from file was attempting to re-cluster (#824).
+
+|fixed| Example notebook outputs no longer clutter search results in docs and search result example notebook pages are grouped under the top-level page header (#675).
 
 ### Internal changes
 
 |fixed| Remove `stderr` redirect at pyomo backend model solve stage to avoid random infinite recursion errors in CI tests (#833).
-
 
 ## 0.7.0.dev7 (2025-09-05)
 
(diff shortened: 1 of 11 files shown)


irm-codebase (approved):
I also think this should solve the problem.

PR #838, opened 2025-10-03
Remove math config contents from array attrs

brynpickering:
As discussed in #834 

- Removes all use of `attrs` to access information that is already stored in the math.
- Removes math attrs from being attached to input data in the first place

>[!NOTE]
>To be merged in after #834 

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/backend/expression_parser.py
+++ b/src/calliope/backend/expression_parser.py
@@ -849,8 +849,10 @@ class EvalUnslicedComponent(EvalToArrayStr):
                 f"Trying to access a math component that is not yet defined: {self.name}. "
                 "If the referenced component is a global expression, set its `order` to have it defined first."
             )
-        if evaluated.isnull().any() and pd.notna(evaluated.attrs["default"]):
-            evaluated = evaluated.fillna(evaluated.attrs["default"])
+        if evaluated.isnull().any() and pd.notna(
+            default := self.eval_attrs["math"].find(self.name)[1].default
+        ):
+            evaluated = evaluated.fillna(default)
 
         self.eval_attrs["references"].add(self.name)
         return evaluated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
-## 0.7.0.dev7 (2025-09-05)
+## 0.7.0.dev8
 
 ### User-facing changes
+
+|removed| Duplication of math attributes in array attributes.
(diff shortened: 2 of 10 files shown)


src/calliope/backend/expression_parser.py
```diff
@@ -849,8 +849,10 @@ def as_array(self) -> xr.DataArray:  # noqa: D102, override
...
             )
-        if evaluated.isnull().any() and pd.notna(evaluated.attrs["default"]):
-            evaluated = evaluated.fillna(evaluated.attrs["default"])
+        if evaluated.isnull().any() and pd.notna(
+            default := self.eval_attrs["math"].find(self.name)[1].default
+        ):
+            evaluated = evaluated.fillna(default)
 
         self.eval_attrs["references"].add(self.name)
         return evaluated
```
irm-codebase, line 864: Just to clarify on my side: why is the `[1]` necessary here?

Shouldn't `find` return the math object, with `default` as a directly accessible value?

irm-codebase (approved):
PRs with net-negative lines of code are the best ones, and fully standardising how math is accessed will save a lot of headaches down the line.

I only have a small question, this looks great otherwise.

src/calliope/backend/expression_parser.py
irm-codebase: I also find the position a bit odd... I'd expect `[0]` in most cases.

brynpickering: Because `find` returns a tuple. The first item is the group it belongs to. Perhaps we don't need that group information and can just return the pydantic model object. I'd need to check where we use it currently

irm-codebase: A quick glance suggests we do use this name, mostly for debug logger scripts and when calling specific functions in the backend. We could just do something like `type(math.find(name)).__name__` in these cases, though.

It could even remove the need for the 's' suffix removal in some areas. Perhaps something to consider for later, or for this PR as an extra.

In any case, not a problem.

brynpickering: I've updated this in #834 to just return the pydantic model and that pydantic model has a `_group` property to return the group it's a part of.

irm-codebase: Sounds good. Feel free to merge!

PR #839, opened 2025-10-03
Postprocess yaml

### LINKED ISSUE #638: Isolate postprocessing data creation methods (YAMLification)

brynpickering:
Fixes #638 

This is very much a proof of concept. It's obviously quite hacky to create a new dummy backend just to be able to parse and evaluate the postprocessed array math. The difficulty is that parsing expects to receive `backend_interface._dataset` containing all the backend arrays. One option would be to change what parsing expects, to be able to switch between passing it `backend_interface._dataset` and `results`, but this would need to filter through quite a few methods. 

~~Another thought I had was to have some kind of context manager that briefly replaces `_dataset` with the contents of `results` on the backend interface while running the postprocessing. I imagine this will also look quite hacky...~~ I gave this a go and it _seems_ cleaner. You can check the difference between https://github.com/calliope-project/calliope/pull/839/commits/a2d2071307cfb4a17dbab7014891ff55c63bc5f1 and https://github.com/calliope-project/calliope/pull/839/commits/cb17af90fe35eaf74fa8c8c4e5e91fb30b05c419

* Moved creation of postprocessed arrays to YAML
* Added `curtailment` postprocessed arrays
* Allow variables/global expressions in where string comparisons _if_ in postprocessing (because then they are just numbers and it's good to be able to apply conditionals to their values). 

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/tests/test_backend_gurobi.py
+++ b/tests/test_backend_gurobi.py
@@ -478,3 +478,18 @@ class TestPiecewiseConstraints:
             excinfo,
             "Gurobi backend can only build piecewise constraints using decision variables.",
         )
+
+
+class TestGurobiImportHandling:
+    """Test handling of Gurobi import failures."""
+
+    def test_gurobi_backend_requires_gurobi(self, monkeypatch):
+        """Test that building with gurobi backend fails gracefully without gurobipy."""
+        import sys
+
+        # Mock gurobipy as unavailable
+        monkeypatch.setitem(sys.modules, "gurobipy", None)
+
+        m = build_model({}, "simple_supply,two_hours")
+        with pytest.raises(ImportError, match="Install the `gurobipy` package"):
+            m.build(backend="gurobi")
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -247,3 +247,28 @@ class TestCLI:
             ],
         )
         assert result.exit_code == 0, result.output
+
+    def test_run_with_profiling(self, tmp_path):
+        """Test profiling output."""
+        runner = CliRunner()
+        profile_file = tmp_path / "profile.stats"
+        result = runner.invoke(
+            cli.run,
+            [
+                _MODEL_NATIONAL,
+                f"--save_netcdf={tmp_path / 'output.nc'}",
+                "--override_dict={'config.solve.solver': 'glpk'}",
+                "--profile",
+                f"--profile_filename={profile_file}",
+            ],
+        )
+        assert result.exit_code == 0, result.output
+        assert profile_file.exists()
+
+    def test_debug_mode_with_exception(self):
+        """Test debug mode shows full traceback."""
+        runner = CliRunner()
+        result = runner.invoke(cli.run, ["nonexistent_model.yaml", "--debug"])
+        assert result.exit_code == 1
+        # Debug mode should show traceback
+        assert "Traceback" in result.output or "Error" in result.output
--- a/tests/test_preprocess_time.py
+++ b/tests/test_preprocess_time.py
@@ -323,3 +323,53 @@ class TestResampling:
             model.inputs.sink_use_equals.sel(nodes="a").fillna(0)
             == model.inputs.sink_use_equals.sel(nodes="b").fillna(0) / 4
         ).all()
+
+
+class TestMissingTimeData:
+    """Test detection of missing timeseries data."""
+
+    def test_warn_missing_timeseries_data(self):
+        """Test warning when timeseries has gaps."""
+        import numpy as np
+        import xarray as xr
+
+        ds = xr.Dataset(
+            {"demand": (["timesteps", "nodes"], [[1, 2], [np.nan, 4], [5, 6]])},
+            coords={"timesteps": pd.date_range("2005-01-01", periods=3, freq="h")},
+        )
+
+        with pytest.warns(match="Possibly missing data on the timesteps dimension"):
+            time._check_missing_data(ds, "timesteps")
+
+    def test_no_warn_all_data_present(self):
+        """Test no warning when all data present."""
+        import xarray as xr
+
+        ds = xr.Dataset(
+            {"demand": (["timesteps", "nodes"], [[1, 2], [3, 4], [5, 6]])},
+            coords={"timesteps": pd.date_range("2005-01-01", periods=3, freq="h")},
+        )
+
+        # Should not raise warning - use context manager to verify
+        import warnings
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            time._check_missing_data(ds, "timesteps")
+
+
+class TestDatetimeConversion:
+    """Test datetime format conversion."""
+
+    @pytest.mark.parametrize(
+        ("time_format", "expected"),
+        [
+            ("ISO8601", pd.Timestamp("2005-01-01")),
+            ("%Y-%m-%d", pd.Timestamp("2005-01-01")),
+        ],
+    )
+    def test_datetime_index_formats(self, time_format, expected):
+        """Test _datetime_index handles different formats."""
+        index = pd.Index(["2005-01-01"])
+        result = time._datetime_index(index, time_format)
+        assert result[0] == expected
--- a/tests/util/test_generate_runs.py
+++ b/tests/util/test_generate_runs.py
@@ -1,0 +1,110 @@
+"""Tests for util.generate_runs module."""
+
+import pytest
+
+from calliope.util import generate_runs
+
+
+@pytest.fixture
+def simple_model_file(tmp_path):
+    """Create a simple model file with scenarios."""
+    model = tmp_path / "model.yaml"
+    model.write_text("""
+scenarios:
+  scenario1: [override1]
+  scenario2: [override2]
+overrides:
+  override1:
+    config.init.name: test1
+  override2:
+    config.init.name: test2
+""")
+    return str(model)
+
+
+@pytest.fixture
+def override_file(tmp_path):
+    """Create an override file."""
+    override = tmp_path / "override.yaml"
+    override.write_text("""
+config:
+  solve:
+    solver: glpk
+""")
+    return str(override)
+
+
+class TestGenerateRuns:
+    def test_generate_from_scenarios(self, simple_model_file):
+        """Test command generation from scenarios in model file."""
+        commands = generate_runs.generate_runs(simple_model_file)
+        assert len(commands) == 2
+        assert (
+            commands[0]
+            == f"calliope run {simple_model_file} --scenario scenario1 --save_netcdf out_1_scenario1.nc"
+        )
+        assert (
+            commands[1]
+            == f"calliope run {simple_model_file} --scenario scenario2 --save_netcdf out_2_scenario2.nc"
+        )
+
+    def test_generate_with_explicit_scenarios(self, simple_model_file):
+        """Test command generation with explicitly specified scenarios."""
+        commands = generate_runs.generate_runs(simple_model_file, scenarios="scenario1")
+        assert len(commands) == 1
+        assert (
+            commands[0]
+            == f"calliope run {simple_model_file} --scenario scenario1 --save_netcdf out_1_scenario1.nc"
+        )
+
+    def test_generate_with_override_dict(self, simple_model_file, override_file):
+        """Test command generation with override dictionary."""
+        commands = generate_runs.generate_runs(
+            simple_model_file, override_dict=override_file
+        )
+        assert len(commands) == 2
+        assert all(f'--override_dict="{override_file}"' in cmd for cmd in commands)
+
+    def test_generate_with_additional_args(self, simple_model_file):
+        """Test command generation with additional arguments."""
+        commands = generate_runs.generate_runs(
+            simple_model_file, additional_args="--debug"
+        )
+        assert len(commands) == 2
+        assert all("--debug" in cmd for cmd in commands)
+
+    def test_generate_padding(self, tmp_path):
+        """Test that output filenames are zero-padded correctly."""
+        model = tmp_path / "model.yaml"
+        overrides = "\n".join(
+            [f"  override{i}:\n    config.init.name: test{i}" for i in range(12)]
+        )
+        model.write_text(f"overrides:\n{overrides}")
+
+        commands = generate_runs.generate_runs(str(model))
+        assert "out_01_" in commands[0]
+        assert "out_12_" in commands[11]
+
+
+class TestGenerateBashScript:
+    def test_generate_script(self, simple_model_file, tmp_path):
+        """Test bash script generation."""
+        output = tmp_path / "run.sh"
+        generate_runs.generate_bash_script(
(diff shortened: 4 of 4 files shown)


src/calliope/schemas/math_schema.py
```diff
@@ -268,6 +268,16 @@ def where(self) -> str:
...
 
 
+class PostprocessArray(GlobalExpression):
+    """Schema for postprocessed expressions.
+
+    Can be used to combine parameters, variables, and global expressions into a single expression solving the model.
+
+    NOTE: If expecting to use postprocessed array `A` in postprocessed array `B`, `A` must
+    be defined above `B`.
+    """
```
irm-codebase: No need for this class, since it's a duplicate. This can just live in the docstring of `PostprocessArrays` below.

src/calliope/schemas/math_schema.py
```diff
@@ -329,6 +339,12 @@ class Objectives(CalliopeDictModel):
     root: dict[AttrStr, Objective] = Field(default_factory=dict)
 
 
+class PostprocessArrays(CalliopeDictModel):
+    """Calliope model postprocessed expressions dictionary."""
```
irm-codebase, line 380: The name is a bit unintuitive. Why not `PostprocessExpression` or `Postprocessed`?

src/calliope/postprocess/postprocess.py
```diff
@@ -26,120 +26,11 @@ def postprocess_model_results(results: xr.Dataset, model: ModelStructure) -> xr.
...
-            results, model.inputs
-        )
-        results["total_levelised_cost"] = systemwide_levelised_cost(
-            results, model.inputs, total=True
-        )
-        results["unmet_sum"] = unmet_sum(results)
-
     results = apply_zero_threshold(results, model.config.solve.zero_threshold)
 
     return results
```
irm-codebase: If this is all this calls, we can simplify by just calling `apply_zero_threshold` at the `Model` level.

src/calliope/backend/where_parser.py
```diff
@@ -19,7 +19,7 @@
 from calliope.util import DTYPE_OPTIONS, tools
 
 if TYPE_CHECKING:
-    from calliope.backend.backend_model import BackendModel
+    pass
```
irm-codebase: Not needed any more?

src/calliope/backend/parsing.py
```diff
@@ -394,6 +394,7 @@ class ParsedBackendComponent(ParsedBackendEquation):
     PARSERS: dict[str, Callable] = {
         "constraints": expression_parser.generate_equation_parser,
         "global_expressions": expression_parser.generate_arithmetic_parser,
+        "postprocessed": expression_parser.generate_arithmetic_parser,
         "objectives": expression_parser.generate_arithmetic_parser,
         "variables": lambda x: None,
         "piecewise_constraints": expression_parser.generate_arithmetic_parser,
```
irm-codebase, line 406: A quick glance at the code suggests `variables` could be removed here.
I could not find any case where `PARSER['variables'] was called using `expression_groups`.

src/calliope/backend/helper_functions.py
```diff
@@ -21,7 +21,7 @@
 from calliope.util import DTYPE_OPTIONS
 
 if TYPE_CHECKING:
-    from calliope.backend.backend_model import BackendModel
+    pass
```
irm-codebase: delete?

src/calliope/backend/expression_parser.py
```diff
@@ -129,7 +129,7 @@ def eval(
             equation_name (str): Name of math component in which expression is defined.
             slice_dict (dict): Dictionary mapping the index slice name to a parsed equation expression.
             sub_expression_dict (dict): Dictionary mapping the sub-expression name to a parsed equation expression.
-            backend_interface (backend_model.BackendModel): Interface to optimisation backend.
+            backend_data (backend_model.BackendModel): Interface to optimisation backend.
             input_data (xr.Dataset): Input data arrays.
```
irm-codebase: docstring update needed for `backend_data` (ditto for other functions with this input)

src/calliope/backend/backend_model.py
```diff
@@ -376,8 +397,13 @@ def _add_component(
                 parsed_component = None
         else:
             self._raise_error_on_preexistence(name, component_type)
+            valid_names = self.valid_component_names
+            if component_type == "postprocessed":
+                valid_names["var_expr_names"] = valid_names["var_expr_names"] | set(
+                    self.math.postprocessed.root
+                )
```
irm-codebase: Potential bug:

I belive this will modify `'var_expr_names'` in self.valid_component_names.
This is likely unintended.

A safer approach would be to build the dictionary here


```suggestion
            valid_components = OPTIM_COMPONENTS # tuple or set, ideally defined in an agreed place
            if component_type == "postprocessed":
                valid_components |= {"postprocessed"}
            valid_components = {k: set(getattr(self.math, k).root)}  # or use a function given by the math schema
```

src/calliope/backend/backend_model.py
```diff
@@ -1212,7 +1295,16 @@ def _drop_attrs(da):
             attrs=self._dataset.attrs,
         ).astype(float)
 
-        return results
+        results = self.add_postprocessed_arrays(results)
```
irm-codebase: This should obey `config.solve.postprocessing_active`

src/calliope/backend/backend_model.py
```diff
@@ -492,6 +518,63 @@ def _clean_arrays(*args) -> None:
...
+        try:
+            self._dataset = temp_dataset
+            for method in methods_to_mock:
+                setattr(self, method, lambda *args, **kwargs: None)
+            yield None
+        finally:
+            # Always restore the original attributes, even if an exception occurs
+            self._dataset = original_dataset
+            for method in methods_to_mock:
+                setattr(self, method, original_methods[method])
```
irm-codebase: This looks dangerous 😬 ... some questions / comments:

- Why are you always yielding `None`?
- replacing methods with `setattr` is likely going to break stuff in odd ways. [descriptors](https://docs.python.org/3/howto/descriptor.html) (class-bound functions) are not equal to plain functions. Basically, when you re-assign them through `setattr`, some of their characteristics are irrevocably lost (for example, calling `self.__name__` is not going to work as intended in those functions after this is called)
- `with` heavily modifies the xarray dataset, which might not be thread safe if two functions are handling it. It's an edge case, though.

In general, this seems to try to get around some of the hard-coded bits of `_add_component`.
It'd be better to instead try to improve the logic of that function (mostly  that all the component additions are funneled through it, meaning it has to deal with specifics of every component type and backend).

irm-codebase: It might be time to re-think how `add_parameter` / `add_dimension` / etc functions operate.

- `_add_component` as it stands should be split, moving specifics to the `add_xxxx` component functions instead. Generic things can become methods these functions call (defined at the `BackendModelGenerator` level).
- The preparation of a component `DataArray` should be separated from its assignation to `self._dataset` to enable more flexibility.

### Logic proposal

Two types of function.

**add_xxxx**
- Generic functions, only defined at the `BackendModelGenerator` level.
- In charge of modification of core shared variables (i.e., `self._dataset` and `self.math`) if, and only if, the given inputs are valid.
- They handle component specific input requirements.
- parser or DataArray logic here.
- These are the functions we expect 'advanced users' to call.

**prepare_xxxx**
- Called by `add_xxxx` functions.
- A combination of `_expression_setter` and component-specific functionality in `_add_component`.
- Their job is to build the respective `DataArray` for each component _without_ modifying `self`. They also raise exceptions, deal with backend particularities, etc.
- For backend specific cases, these functions may rely on their parent up via `self.super().add_xxxxx(...)` or completely replace functionality necessary.

    
```python
class BackendModelGenerator:


    def add_foobar(self, name: str, definition: math_schema.Foobar, force: bool = False):
        """I'm in charge of validity and storage, and nothing else"""
        # Generic logic to handle inactive components, duplicates or invalid setups.
        if not definition.active:
            self.log("Not active. I'll do nothing.")
            return
        self.raise_error_on_preexistence(name, "foobar")
        
        # Call the worker, and tell it what is valid or not
        valid_names = set().union(*self._valid_components.values(), self.math.foobars.root)
        foo_da = self.prepare_foobar(name, definition, valid_names)

        # Only add if preparation was successful
        # In some cases, you might want to save to self.results instead!
        self._add_to_dataset(name, foo_da, "foobar")
        if force or name not in self.math["foobar"]:
            self.math = self.math.update(...)

    def prepare_foobar(
        self, name, definition: math_schema.Foobar, valid_names: set[str]
    ) -> xr.DataArray:
        """I do the actual work... without altering anything!

        I might call generics or self.super() in many cases!
        """
        parsed_foobar = parsing.ParsedFoobar(name, definition, valid_names)

        top_where = parsed_foobar.generate_top_level_where_array(
            self, align_to_foreach_sets=False, break_early=True
        )
        if not top_where.any():
            return xr.DataArray(np.nan)

        # setters live here
        def _foobar_setter(
            element: parsing.ParsedFoobar, where: xr.DataArray, references: set
        ) -> xr.DataArray:
            ...  # imagine setter code here
            return foobar_da


        equations = parsed_foobar.parse_equations()
        elements_in_foreach_da = parsed_foobar.drop_dims_not_in_foreach(top_where)
        if not equations:
            foobar_da = _foobar_setter(elements_in_foreach_da)
        else:
            xr.DataArray().where(elements_in_foreach_da).astype(np.dtype("O"))
            for e in equations:
                where = e.evaluate_where(self._data, top_were=top_where)
                if not where.any():
                    continue
                where = parsed_foobar.drop_dims_not_in_foreach(where)
                if foobar_da.where(where).notnull().any():
                    raise ValueError("something something equations bad")
                foobar_da = foobar_da.fillna(_foobar_setter(e, where))

        return foobar_da


```

src/calliope/backend/parsing.py
brynpickering: It comes up in `_add_component` ;)

src/calliope/backend/backend_model.py
```diff
@@ -34,7 +35,7 @@
 from calliope.schemas import config_schema, math_schema
 
 if TYPE_CHECKING:
-    from calliope.backend.parsing import T as Tp
+    pass
```
irm-codebase: Can be removed.

src/calliope/backend/backend_model.py
```diff
@@ -173,8 +162,108 @@ def add_parameter(
             values (xr.DataArray): Array of parameter values.
             definition (math_schema.Parameter): Parameter math definition.
         """
+        self._raise_error_on_preexistence(name, "parameters")
+
```
irm-codebase, line 160: I find it a bit odd that this is checked for parameters, but not for other components...
Is there a particular reason for that?

src/calliope/backend/backend_model.py
```diff
@@ -173,8 +162,108 @@ def add_parameter(
...
+                "Component deactivated; only metadata will be stored if no other component with the same name is defined.",
+            )
+            if (
+                name in self.math.parsing_components["where"]["results"]
+                # FIXME: won't work if we later add a global expression with the same name
+                and name not in self.math.parsing_components["expression"]["results"]
+            ):
+                component_da = default_empty
+            else:
+                return
```
irm-codebase: I suspect this and the else below can be removed with a bit of logic re-arrangement.

src/calliope/backend/backend_model.py
```diff
@@ -562,25 +694,25 @@ def add_postprocessed_arrays(self, results: xr.Dataset) -> xr.Dataset:
         ordered_items = sorted(
             postprocessed_math.items(), key=lambda item: item[1].order
         )
```
irm-codebase: There is another section where we sort things in this file...
Seems like they are equivalent, so a bit of standardisation could help stay consistent.

src/calliope/backend/expression_parser.py
```diff
@@ -34,41 +34,25 @@
...
 import xarray as xr
-from typing_extensions import NotRequired, TypedDict, Unpack
 
 from calliope.backend.helper_functions import ParsingHelperFunction
 from calliope.exceptions import BackendError
-from calliope.schemas.math_schema import CalliopeBuildMath
 
 if TYPE_CHECKING:
-    pass
+    from calliope.backend.parsing import EvalAttrs
```
irm-codebase: Hm.
This circular import suggests EvalAttrs should not live in `parsing`...
Not the biggest thing, but perhaps it should live in its own file?

src/calliope/backend/pyomo_backend_model.py
```diff
@@ -45,10 +46,106 @@
...
+        component_dict = getattr(instance, component_type)
+        if key in component_dict:
+            raise BackendError(
+                f"Trying to add already existing `{key}` to backend model {component_type}."
+            )
+        else:
+            singular_component = component_type.removesuffix("s")
+            component_dict[key] = getattr(
+                pmo, f"{COMPONENT_TRANSLATOR[singular_component]}_list"
+            )()
```
irm-codebase: How about adding the 's' to the translator, to skip the need to remove it in execution?

src/calliope/backend/pyomo_backend_model.py
```diff
@@ -45,10 +46,106 @@
...
+            if hasattr(cls, method_name):
+                original = getattr(cls, method_name)
+                setattr(
+                    cls,
+                    method_name,
+                    _create_backend_obj_adding_component(
+                        original, method_name.removeprefix("add_") + "s"
+                    ),
+                )
+        return cls
```
irm-codebase, line 140: I *think* I get what this achieves, but let's confirm...
It looks like this renames the add_xxxx methods to -> variables / prameters / etc?

If so, why? If not, perhaps add a bit more explanation on what it achieves?

src/calliope/backend/backend_model.py
irm-codebase: Fantastic changes! I think this is solved :+1:

src/calliope/backend/pyomo_backend_model.py
brynpickering: It's a wrapper to create the empty pyomo component list/dict objects to be filled but delete those objects if they ultimately aren't needed. I'll add more explanation.

irm-codebase (dismissed):
And it's done!

PR #845, opened 2025-11-21
Add all test math lps

### LINKED ISSUE #433: Compare LPs to improve testing of constraints

brynpickering:
Fixes #433 

This was written with the support of Claude Sonnet 4.5 as it is such a mind-numbing task. I have spot checked the generated LPs and they seem fine. These LPs are generated by taking the dumped LPs from building the models, so they're valid so long as our math is valid 😅.

One change to the pre-defined math I've made is to move all milp-related params and lookups to the `milp` math file. This means that `cap_method` is only recognised as a lookup if the `milp` math is included. I could move it back if preferred.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/math/base.yaml
+++ b/src/calliope/math/base.yaml
@@ -1416,8 +1416,8 @@ checks:
   no_storage_discharge_depth_when_clustering:
     where: storage_discharge_depth AND lookup_cluster_last_timestep
     message: >-
-      storage_discharge_depth is currently not allowed when time clustering is active.
-    errors: raise
+      storage_discharge_depth will currently have no effect when time clustering is active.
+    errors: warn
 
   cyclic_storage_needs_inter_cluster:
     where: cyclic_storage AND lookup_cluster_last_timestep
--- a/docs/hooks/dummy_model/model.yaml
+++ b/docs/hooks/dummy_model/model.yaml
@@ -24,6 +24,7 @@ overrides:
       spores_slack: 0.1
 
 config.init.name: base
+config.build.ensure_feasibility: true
 
 nodes:
   A.techs: { demand_tech, conversion_tech, supply_tech, storage_tech }
--- a/docs/hooks/generate_math_docs.py
+++ b/docs/hooks/generate_math_docs.py
@@ -200,36 +200,34 @@ def generate_custom_math_documentation(
     full_del = []
     expr_del = []
     for component_group, component_group_dict in model.math.build.model_dump().items():
-        if component_group == "checks":
+        if component_group in ["checks", "dimensions", "parameters", "lookups"]:
             continue
         for name, component_dict in component_group_dict.items():
-            if name in base_documentation.math[component_group]:
+            if name in base_documentation.math[component_group]._active:
                 if not component_dict.get("active", True):
-                    expr_del.append(name)
+                    expr_del.append((component_group, name))
                     component_dict["description"] = "|REMOVED|"
                     component_dict["active"] = True
-                elif (
-                    base_documentation.math[component_group].get(name, {})
-                    != component_dict
-                ):
+                elif base_documentation.math.find(name).model_dump() != component_dict:
                     _add_to_description(component_dict, "|UPDATED|")
                 else:
-                    full_del.append(name)
+                    full_del.append((component_group, name))
             else:
                 _add_to_description(component_dict, "|NEW|")
-
+        model.math = model.math.update(
+            {f"build.{component_group}": component_group_dict}
+        )
     math_documentation = MathDocumentation(model)
-    for key in expr_del:
-        math_documentation.backend._dataset[key].attrs["math_string"] = ""
-    for key in full_del:
+
+    for grp, key in expr_del:
+        math_documentation.backend.math_strings[grp][key] = ""
+    for grp, key in full_del:
         del math_documentation.backend._dataset[key]
     for var in math_documentation.backend._dataset.values():
         var.attrs["references"] = var.attrs["references"].intersection(
             math_documentation.backend._dataset.keys()
         )
         var.attrs["references"] = var.attrs["references"].difference(expr_del)
-
-    logger.info(math_documentation.backend._dataset["carrier_in"].attrs["references"])
 
     return math_documentation
 


src/calliope/math/base.yaml
```diff
@@ -156,14 +149,6 @@ parameters:
...
     unit: $\frac{\text{cost}}{\text{energy}}$
 
-  cost_purchase:
-    default: 0
-    title: Purchase cost.
-    description: >-
-      Cost applied to the variable `purchased_units`.
-      Requires the parameter `cap_method` to be `integer`.
-    unit: $\frac{\text{cost}}{\text{unit}}$
-
```
irm-codebase, line 166: I'm surprised (and a bit worried) about the amount of bits I missed in the MILP separation PR.
Thanks for the corrections!

src/calliope/math/base.yaml
```diff
@@ -1285,7 +1200,7 @@ global_expressions:
       depreciation_rate:
         - where: cost_depreciation_rate
           expression: cost_depreciation_rate
-        - where: NOT cost_depreciation_rate AND cost_interest_rate==0
+        - where: NOT cost_depreciation_rate AND (NOT cost_interest_rate OR cost_interest_rate==0)
```
irm-codebase, line 1203: I assume we now enforce capitalisation of conditionals (e.g., `NOT`) ?

src/calliope/math/base.yaml
```diff
@@ -1469,35 +1384,20 @@ checks:
     errors: raise
 
   all_or_nothing_lat_lon:
-    where: (any(latitude, over=nodes) or any(longitude, over=nodes)) and not (latitude or longitude)
+    where: (any(latitude, over=nodes) or any(longitude, over=nodes)) AND NOT (latitude or longitude)
```
irm-codebase, line 1387: `or` -> `OR`?
(if capitalisation is not a requirement, feel free to ignore this)

src/calliope/math/milp.yaml
```diff
@@ -355,3 +467,24 @@ global_expressions:
           cost_investment_source_cap +
           cost_investment_area_use +
           cost_investment_purchase
+
+checks:
+  distance_only_for_transmission_milp:
+    where: not base_tech==transmission and cost_purchase_per_distance
+    message: Only transmission technologies can define `distance` or `..._per_distance`.
```
irm-codebase, line 474: Nitpick: `not` -> `NOT`?

src/calliope/math/storage_inter_cluster.yaml
```diff
@@ -159,3 +159,7 @@ variables:
     bounds:
       min: -.inf
       max: .inf
+
+checks:
+  cyclic_storage_needs_inter_cluster:
+    active: false
```
irm-codebase, line 165: Nitpick:
For what I remember, this check will pass regardless. Leaving this check should be mostly benign.
I suggest to not deactivate checks unless needed to avoid breakage to help with maintainability (less divergence between math options)

tests/common/lp_files/storage_max_inter_cluster_override.lp
irm-codebase, line 1: This one is... odd.
I guess it checks that the constraint is not there?

src/calliope/math/base.yaml
brynpickering: It's not necessary for the parser but I've gone for all caps in pre-defined math generally to make it as clear as possible. If we miss some, it isn't critical

tests/common/lp_files/storage_max_inter_cluster_override.lp
brynpickering: Indeed. Our final test per class checks that all defined keys in the math file have been registered. So we could register this one elsewhere just to get that test to pass or, as I've done here, force check that the constraint is ignored. I don't mind which we do tbh.

src/calliope/math/storage_inter_cluster.yaml
brynpickering: This would raise an error if we don't deactivate it. It's a check that says if inter cluster storage isn't activated then there's a problem. Since this file loads inter-cluster storage, I thought it was neatest to just switch off that check (since all is now good). The other approach would be to mention one of the inter-cluster storage variables/params in the base math check, but this then pollutes the base math with something defined in the the inter-cluster storage math. Which is more intuitive...?

irm-codebase: Got it, in that case it's fine to leave as-is. I was thinking of the old case, before `checks:` was introduced, I guess?

A quick glance suggests that we have a bit of 'pollution' in `base.yaml` for the `balance_supply_with_storage` constraint, though. Similar to `milp`, it looks like that constraint has a `cluster_first_timestep` 'leftover' from the more mixed-up version of the math... Perhaps it's possible to even remove the check if that's fixed.

tests/common/lp_files/storage_max_inter_cluster_override.lp
irm-codebase: It's fine to leave this as-is, IMO.

src/calliope/math/storage_inter_cluster.yaml
irm-codebase: That's for a different PR, though.

src/calliope/math/base.yaml
brynpickering: I think this could have been because you split out before we added params/lookups?

src/calliope/math/storage_inter_cluster.yaml
brynpickering: > it looks like that constraint has a cluster_first_timestep 'leftover' from the more mixed-up version of the math

No, that's fine. You can have time clustering mixed with `base` math (which is where `cluster_first_timestep` is used). Inter-cluster storage adds more math on top of that. Although I suppose we should have a check that time clustering is even being used when inter-cluster storage math is added 😅

irm-codebase (approved):

src/calliope/math/base.yaml
irm-codebase: True. What matters is that these are in the right place now, I suppose.

PR #846, opened 2025-11-21
Make gurobi warmstart test compatible with gurobi>=13

brynpickering:
Fixes issues coming up in the CI.

Gurobi's LPWarmstart has a different default now (compare [now](https://docs.gurobi.com/projects/optimizer/en/current/reference/parameters.html#lpwarmstart) to [previous](https://docs.gurobi.com/projects/optimizer/en/12.0/reference/parameters.html#lpwarmstart)).

Since `1` means a type of warmstart is activated in either version, and `-1` doesn't even exist in the older version, I think this simple fix is sufficient (rather than checking gurobi version and changing test accordingly).

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/tests/test_backend_gurobi.py
+++ b/tests/test_backend_gurobi.py
@@ -288,6 +288,7 @@ class TestNewBackend:
         simple_supply_gurobi.solve(force=True, warmstart=True)
 
         # default warmstart = 1 (<v13), = -1 (>=v13)
+        # TODO: Remove `1` check when dropping support for gurobi<13
         assert simple_supply_gurobi.backend._instance.Params.LPWarmStart in [-1, 1]
         assert (
             "LPWarmStart"


tests/test_backend_gurobi.py
```diff
@@ -286,9 +286,9 @@ def test_set_solver_option(self, simple_supply_gurobi):
 
     def test_set_warmstart(self, simple_supply_gurobi):
         simple_supply_gurobi.solve(force=True, warmstart=True)
-        assert simple_supply_gurobi.backend._instance.Params.LPWarmStart == 1
 
-        # warmstart = 1 is the Gurobi default, so no change if warmstart=True
+        # default warmstart = 1 (<v13), = -1 (>=v13)
+        assert simple_supply_gurobi.backend._instance.Params.LPWarmStart in [-1, 1]
```
irm-codebase, line 292: I'd keep the check to -1.
This is test code, so we should follow the latest version, and not care about previous ones for maintainability.

irm-codebase (changes requested):

tests/test_backend_gurobi.py
brynpickering: As nice an idea as that is, gurobipy versions are hard linked to license versions. You have to request a new license version to use an updated version of gurobi successfully. This is quite a headache for anyone doing dev, if they've locally pinned to an older gurobipy version to match their local license!

irm-codebase: True...
`LPWarmStart` is limited to a max of 2.
My main worry is that this will now pass in 50% of cases, making it a weaker test overall.

I guess it's fine as long as we make note of removing `1` after a couple of years.

irm-codebase (dismissed):

tests/test_backend_gurobi.py
brynpickering: OK, I'll make a note. `1` and `2` are both warmstart cases. In the older version, `1` is the default. In the newer version, `-1` is the default means `1` or `2` will be chosen automatically dependending on other parameters. Still, it would be nice if they hadn't made such a breaking change 😅

irm-codebase: They kind of had to, for security reasons. From a QA point of view the previous approach was quite weak.

irm-codebase: The busywork does stink, though

PR #847, opened 2025-11-21
Add tests for pre-defined math checks

brynpickering:
Separate set of tests of our YAML checks, to replace all those xfailing/skipped tests (alongside the additional tests added in #845). 

To be reviewed after #845

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/math/base.yaml
+++ b/src/calliope/math/base.yaml
@@ -368,7 +368,7 @@ parameters:
     resample_method: sum
     title: Maximum bound on source.
     description: >-
-      Maximum sink use to add a carrier to the system (e.g., biofuel, coal, rainfall, wind flow).
+      Maximum source use to add a carrier to the system (e.g., biofuel, coal, rainfall, wind flow).
       Unit dictated by `source_unit`.
     unit: energy | $\frac{\text{energy}}{\text{power}}$ | $\frac{\text{energy}}{\text{area}}$
 


src/calliope/math/base.yaml
```diff
@@ -366,9 +366,9 @@ parameters:
   source_use_max:
     default: .inf
     resample_method: sum
-    title: Maximum bound on sink.
+    title: Maximum bound on source.
     description: >-
-      Maximum sink use to remove a carrier from the system (e.g., biofuel, coal, rainfall, wind flow).
+      Maximum sink use to add a carrier to the system (e.g., biofuel, coal, rainfall, wind flow).
```
irm-codebase: sink->source?
```suggestion
      Maximum source use to add a carrier to the system (e.g., biofuel, coal, rainfall, wind flow).
```

tests/test_math_checks.py
irm-codebase, line 1: Hmm. Even though it saves code, I worry that this new approach will make testing failures harder, since you need to run all tests in a bundle. IMO, it's better to have these inside their own test case.

irm-codebase: This _IS_ a matter of style, but I'd generally advice for specificity when it comes to testing.

irm-codebase: See this example: even trying to call a specific case is hard, since you lose the relation between the name in the YAML and the pytest case.

Image

irm-codebase: Suggesting  #852 to deal with this

irm-codebase (approved):
Suggested changes were implemented.
Ready to merge!

PR #848, opened 2025-11-26
Fix/differing nodes techs in tables

### LINKED ISSUE #806: Can't use data tables to set timeseries for at different levels with the same parameter
### What happened?

Ran into an issue where using a data table to set a parameter timeseries on a tech (`Utility_PV_2`) doesn't make it into the model if that parameter is set at a node somewhere else in the model **but not at any of the nodes with the original tech** (`Utility_PV_2`).

The attached model ([tech_timeseries_issue.zip](https://github.com/user-attachments/files/21997888/tech_timeseries_issue.zip)) demonstrates the issue. It has two PV technologies with `source_use_max` set at different levels. `Utility_PV` is instantiated at the `PV` node and has it set at the node level.

Image

`Utility_PV_2` has it set at the tech level and is instantiated at both the `PV` and `Kona` nodes.

Image

The model correctly sets `source_use_max` for both technologies at `PV` but fails to set `source_use_max` for `Utility_PV_2` at `Kona`. 

Image

CSV dump of the post-run model NetCDF

Any other technologies that try and set `source_use_max` at the tech level will also not set the parameter at any node that doesn't have `source_use_max` defined at the node level for some tech there. Setting a flat value within the YAML still works but assigning timeseries via data tables does not. This has been tested with several timeseries parameters, but I haven't tested it for using data tables with non-timeseries parameters.

- [ ] macOS
- [x] Windows
- [x] Linux

### Version

v0.7.0.dev6

### Relevant log output

```Shell

```

brynpickering:
Fixes #806 

* Pre-load the model definition techs and nodes for each data table in the model builder.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


src/calliope/preprocess/model_data.py
```diff
@@ -180,7 +180,11 @@ def init_from_data_tables(self, data_tables: Iterable[data_tables.DataTable]):
                     lookup_dict = data_table.lookup_dict_from_param(param, lookup_dim)
                     self.tech_data_from_tables.union(lookup_dict)
                     data_table.drop(param)
-
+        # Pre-populate the dataset with model nodes and techs
+        self.dataset = self.dataset.assign_coords(
+            nodes=list(self.model_definition["nodes"]),
+            techs=list(self.model_definition["techs"]),
+        )
```
irm-codebase, line 187: Ok, I think I see it.
Basically, this ensures the dataset is updated even in cases where definitions are 'spread out' in different files/levels.

@brynpickering just to confirm: the new approach _should_ be insensitive to the order in which data files are loaded, since the magic happens during the `build` of the data factory. In the failing example, node-specific data was loaded first and then the tech generic one second. Reversing the order should have no effect, correct?

brynpickering: Resolved this by explicitly checking it in the test

brynpickering: > Basically, this ensures the dataset is updated even in cases where definitions are 'spread out' in different files/levels.

It ensures that all tables being loaded in are "aware" of all the techs and nodes in the model, so if a data table defines something for node A, it knows that it is _not_ defining it for node B since it knows that node B exists (the data table itself makes no reference to node B, after all)

PR #849, opened 2025-11-26
Restructure test directory; remove unused files and fixtures

brynpickering:
Fixes #240 

Using the same directory structure in `tests` as in `src` and moving test configs into those subdirs where they are only used by those subdir tests.

NOTE: to be merged in after #847 

## Reviewer checklist

- [ ] Test(s) added to cover contribution
- [ ] Documentation updated
- [ ] Changelog updated
- [ ] Coverage maintained or improved

PR #850, opened 2025-11-28
Add monthly peak charging example; Fix example math tests

### LINKED ISSUE #727: Building support for monthly charges into Calliope
### What can be improved?

Idea/Goal:
Define custom lookup arrays on specific indices to enable more advanced constraint/expression `where` filtering.

Background:
We have been running into models/scenarios where we want to have technologies in Calliope that have monthly bills including monthly peak energy usage charges. Examples could include certain third-party power purchase agreements or even utility tariffs and the monthly power bills that a customer sees. We have modeled this in Calliope using multiple capacity expansion transmission/conversion techs/nodes, having one enabled (via efficiency) each month which is a pretty clunky workaround.

We have been looking into the custom math options in 0.7 to see if we can simplify building this behavior/make it more efficient and have been considering adding new expressions or decision variables to track the peak demand each month so we can apply costs to it. We're struggling to work out a way to tie values on the `timesteps` index to months in the expression `where` clauses. `select_from_lookup_arrays` looks like it would be the right helper function if we could get a lookup array with the month for each timestep, but we don't see a way to add a lookup array to the problem formulation.

If there is a way to add a lookup array to the problem formulation, we could potentially selectively enable constraints based on month, weekday/weekend, or any other categorization of timesteps/index values. A work around could be to define a timeseries parameter at the tech (or a master template) level with the month of each timestep that we can then access in a `where` clause.

Adding more decision variables might not even improve the performance of the model compared to multiple techs, but wanted to flag a potential use for the ability to add custom lookup arrays or find another way to categorize index values to enable more flexible constraint definition.

### Version

v0.7.0dev4

brynpickering:
Fixes #727 

Also adds a minor fix into how user-defined math examples are tested.

NOTE: to be merged in after #849 

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


docs/user_defined_math/examples/monthly_peak_flow_charge.yaml
```diff
@@ -0,0 +1,71 @@
+# title: Monthly peak flow charge
+#
+# description: |
+#   Apply a cost in each month of the year linked to technology the peak outflow in that month.
```
irm-codebase: Wording is a bit off

" linked to technology the peak outflow in that month." -> " linked to the peak technology outflow in that month."?

tests/math/math_test.py
```diff
@@ -996,3 +1016,43 @@ def test_net_annual_import_share_max_node_group(self, build_and_compare):
                 "parameters": ["net_import_share"],
             },
         )
+
+
+class TestPeakMonthCost(CustomMathExamples):
+    YAML_FILEPATH = "monthly_peak_flow_charge.yaml"
+    shared_overrides = {
+        "config.init.subset.timesteps": ["2005-01-01 00:00:00", "2005-01-01 03:00:00"],
```
irm-codebase, line 1032: Consider adding a brief docstring explaining this (after all, you're linking January timesteps to Feb, due to computational reasons, I suppose?).

It helps avoid future confusion

PR #856, opened 2025-12-16
Add highs backend

### LINKED ISSUE #822: Support HiGHS solver either via new backend interface or by relying on linopy
### What can be improved?

The open-source solver HiGHS is applicable out of the box to Pyomo. However, when trying to use it in Calliope, an issue arises either due to a failure to recognise HiGHS (options 1 and 2 below) or because of `solver_io`, which Calliope always tries to pass to the solver, although (apparently) HiGHS doesn't want it (option 2).

Here is an example of how I defined HiGHS in my `model.yaml` file, after having installed it in my Calliope env.

### Option 1
```
config:
    solve:
        solver: highs
```

### Option 2
```
config:
    solve:
        solver: highs
        solver_io: lp
```

### Option 3
```
config:
    solve:
        solver: appsi_highs
```

If I specify `highs` as the solver (**options 1 and 2**), I get an error stating that `highs` is a missing solver:

```
WARNING: Failed to create solver with name 'highs': Still working on this
Traceback (most recent call last):
  File "/home/fl/miniconda3/envs/calliope_v07/lib/python3.12/site-packages/pyomo/opt/base/solvers.py", line 148, in __call__
    opt = self._cls[_name](**kwds)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fl/miniconda3/envs/calliope_v07/lib/python3.12/site-packages/pyomo/contrib/solver/common/base.py", line 389, in __init__
    raise NotImplementedError('Still working on this')
NotImplementedError: Still working on this
```

The above is unexpected, because, in the same env, I can call `highs` without any problem for a random test Pyomo model, as in:

```
solver = SolverFactory("highs")
result = solver.solve(m, tee=True)
print(result.solver.status, result.solver.termination_condition)
```

If, instead, I specify `appsi_highs` as the solver (**option 3**) HiGHS is recognised, but there's a complaing about the `solver_io` parameter:

```
WARNING: Failed to create solver with name 'appsi_highs': Highs.__init__() got
an unexpected keyword argument 'solver_io'
Traceback (most recent call last):
  File "/home/fl/miniconda3/envs/calliope_v07/lib/python3.12/site-packages/pyomo/opt/base/solvers.py", line 148, in __call__
    opt = self._cls[_name](**kwds)
          ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Highs.__init__() got an unexpected keyword argument 'solver_io'
```

### Version

v0.7.0.dev4

brynpickering:
Fixes #822 

## TODO
- [x] create verbose string representations programmatically (highs does not use standard math expressions by default)
- [x] fix infeasibilities in operate mode, likely caused by incorrect col/row deletion when updating params, but could also be due to some presolve/warmstarting issues

* New Highs backend interface

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,616 @@
...
+                self._apply_func(
+                    __renamer,
+                    da.notnull(),
+                    1,
+                    da,
+                    *[da.coords[i].str.replace(" ", "_") for i in da.dims],
+                    name=da.name,
+                )
+                da.attrs["coords_in_name"] = True
+
```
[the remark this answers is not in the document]
brynpickering: Still need to investigate validity of this

src/calliope/backend/expression_parser.py
```diff
@@ -254,7 +254,7 @@ def _operate(
             case "*":
                 val = val * evaluated_operand
             case "/":
-                val = val / evaluated_operand
+                val = val * (1 / evaluated_operand)
             case "+":
```
irm-codebase, line 261: Quite an odd fix! Out of curiosity: what was going on here?

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+        """Highs solver interface class.
+
+        Args:
+            inputs (xr.Dataset): Calliope model data.
+            math (math_schema.CalliopeBuildMath): Calliope math.
+            build_config (config_schema.Build): Build configuration options.
+        """
+        if importlib.util.find_spec("highspy") is None:
+            raise ImportError(
+                "Install the `highspy` package to build the optimisation problem with the Highs backend."
```
irm-codebase: 🤓 ☝️ 

```suggestion
            raise ImportError(
                "Install the `highspy` package to build the optimisation problem with the HiGHS backend."
```

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+        self, name: str, values: xr.DataArray, definition: math_schema.Parameter
+    ) -> None:
+        self._raise_error_on_preexistence(name, "parameters")
+
+        if values.isnull().all():
+            self.log("parameters", name, "Component not added; no data found in array.")
+            values = xr.DataArray(np.nan, attrs=values.attrs)
+
+        self._add_to_dataset(name, values, "parameters", definition.model_dump())
+
```
irm-codebase: Unless I am missing something, we can just call `super` to avoid duplicate code here.
```suggestion
        super().add_parameter(name, values, definition)
```

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+        ub = self._get_variable_bound(bounds.max, name, references)
+        var = self._apply_func(
+            self._instance.addVariable, where, 1, lb, ub, type=domain_type
+        )
+
+        return var.fillna(value=np.nan)
+
+    def _add_global_expression(  # noqa: D102, override
+        self, name: str, where: xr.DataArray, expression: xr.DataArray
+    ) -> xr.DataArray:
```
irm-codebase, line 104: Nice-to-haves:
- seems like `name` is never used by any backend. Could be deleted from all of these calls to save a bit of memory in the stack.
- This function is very similar across all backends, with the exception of the `pyomo` one. The only difference is that `self.clean_arrays(expression)` is not run for that one. Is there a reason for it?

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+        expression = expression.squeeze(drop=True)
+        self._clean_arrays(expression)
+        to_fill = expression.where(where)
+
+        return to_fill
+
+    def _add_constraint(  # noqa: D102, override
+        self, name: str, where: xr.DataArray, expression: xr.DataArray
+    ) -> xr.DataArray:
+        to_fill = self._apply_func(self._instance.addConstr, where, 1, expression)
```
irm-codebase: Nitpick: `name` is also unused here. Oddly enough, `pyomo` is the only backend using the name for this one, although it's just for an error message that might be depreciated (as we now handle preexistence catches at the `add_constraint` level).

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+
+        return to_fill
+
+    def _add_objective(  # noqa: D102, override
+        self, name: str, where: xr.DataArray, expression: xr.DataArray, sense: int
+    ) -> xr.DataArray:
+        if name == self.objective:
```
irm-codebase, line 139: Nitpick: in this case, `where` is never used.

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+                else:
+                    df = pd.Series(
+                        self._instance.vals(variable.to_series().dropna().to_dict())
+                    ).reindex(variable.coords.to_index())
+                    da = df.to_xarray()
+                return da.rename(variable.name).assign_attrs(variable.attrs)
+            except AttributeError:
+                return self._apply_func(
+                    self._expr_to_str, variable.notnull(), 1, variable
+                )
```
irm-codebase, line 213: I'd re-work to have only one `return` statement. Makes the logic clearer, and is easier to debug for humans.

```suggestion
    def get_variable(  # noqa: D102, override
        self, name: str, as_backend_objs: bool = True
    ) -> xr.DataArray:
        variable = self.variables.get(name, None)
        if variable is None:
            raise KeyError(f"Unknown variable: {name}")
        if not as_backend_objs and not variable.isnull().all():
            try:
                if not variable.dims:
                    da = xr.DataArray(self._instance.val(variable.item()))
                else:
                    df = pd.Series(
                        self._instance.vals(variable.to_series().dropna().to_dict())
                    ).reindex(variable.coords.to_index())
                    da = df.to_xarray()
                variable = da.rename(variable.name).assign_attrs(variable.attrs)
            except AttributeError:
                variable = self._apply_func(
                    self._expr_to_str, variable.notnull(), 1, variable
                )
        return variable
```

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+
+    def get_variable_bounds(self, name: str) -> xr.Dataset:  # noqa: D102, override
+        variable = self.get_variable(name, as_backend_objs=True)
+
+        lb, ub = self._apply_func(
+            self._from_highs_variable_bounds, variable.notnull(), 2, variable
+        )
+        return xr.Dataset({"lb": lb, "ub": ub}, attrs=variable.attrs)
```
irm-codebase, line 222: Nitpick: repeated code across all backends, with the exception of the name of the called function (`_from_highs_variable_bounds`, `_from_gurobi_variable_bounds`, etc).

If that function was renamed to `_from_backend_variable_bounds` or similar, and set to be the `@abstractmethod` all these functions could be removed and we could just rely on the parent class from`backend_model.py`.

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+        self._instance.resetOptions()
+        self._instance.clearSolver()
+        if solve_config.solver_options is not None:
+            for k, v in solve_config.solver_options.items():
+                self._instance.setOptionValue(k, v)
+
+        if warmstart:
+            model_warn(
+                "The chosen solver, highs, does not support warmstart, which may impact performance."
+            )
```
irm-codebase: Does not seem to be true?
https://github.com/ERGO-Code/HiGHS/issues/617#issuecomment-982148533
From their comments, it seems like this is a 'just trust the backend' situation.

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+            results = xr.Dataset()
+
+        termination = self._instance.modelStatusToString(termination).lower()
+        results.attrs["termination_condition"] = str(termination)
+        return results
+
+    def verbose_strings(self) -> None:  # noqa: D102, override
+        def __renamer(val, *idx, name: str):
+            new_obj_name = f"{name}[{'__'.join(idx)}]"
+            val.name = new_obj_name
```
irm-codebase, line 283: Nitpick: this is correct, but `__renamer` appears to have a redundant `pd.notna(val)` in other backends that could be removed

src/calliope/backend/highs_backend_model.py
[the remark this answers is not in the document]
irm-codebase: The bot seems to be correct here.

Both the `highs` and `gurobi` backends break this.
The cause is that we have a lot of repeated code in these functions. We should try to have them be generic, and just modify a couple of backend dependent hooks here and there.
In this case it'd be the code inside the `for`, and perhaps a class-constant for `attribute_names`.

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+
+    @staticmethod
+    def _collect_indices(da: xr.DataArray) -> list[int]:
+        """Gather the `.index` attribute of every non-null backend object in a DataArray."""
+        indices: list[int] = []
+        for val in da.values.ravel():
+            if val is None or (isinstance(val, float) and np.isnan(val)):
+                continue
+            indices.append(int(val.index))
+        return indices
```
irm-codebase, line 337: This and the `_shift_stale_indices` function below are quite dangerous if we get them wrong, and I do not see a highs-specific test file.

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+                        continue
+                    val.index = val.index - bisect.bisect_left(
+                        sorted_deleted, val.index
+                    )
+            elif obj_type in expr_targets:
+                for val in da.values.ravel():
+                    if not isinstance(val, highspy.highs.highs_linear_expression):
+                        continue
+                    val.idxs = [
+                        i - bisect.bisect_left(sorted_deleted, i) for i in val.idxs
```
irm-codebase, line 374: An O(n^2) run for every deletion is quite nasty.
We might need to open an issue to revisit this later.

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+                )
+
+        self._apply_func(
+            self._update_highs_variable,
+            variable_da.notnull() & xr.DataArray(new_bounds).notnull(),
+            1,
+            variable_da,
+            bound_das["min"],
+            bound_das["max"],
+        )
```
irm-codebase, line 428: This is another case where all backends have the same or equivalent, except for the very last piece of code. We should at least open an issue to harmonise these (either in our codebase or by moving to linopy).

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+        self,
+        x_var: highspy.highs.highs_var,
+        y_var: highspy.highs.highs_var,
+        *vals: float,
+        name: str,
+        n_breakpoints: int,
+    ) -> None:
+        raise NotImplementedError(
+            "Piecewise constraints are not yet implemented for the Highs backend."
+        )
```
irm-codebase, line 472: We should open a feature issue for v0.7+

src/calliope/backend/highs_backend_model.py
```diff
@@ -0,0 +1,623 @@
...
+            return f"{body} == {lb:g}"
+        if np.isneginf(lb):
+            return f"{body} <= {ub:g}"
+        if np.isposinf(ub):
+            return f"{body} >= {lb:g}"
+        return f"{lb:g} <= {body} <= {ub:g}"
+
+
+class HighsShadowPrices(backend_model.ShadowPrices):
+    """Highs shadow price functionality."""
```
irm-codebase, line 617: Add a brief explanation mentioning that HiGHS cannot 'turn off' the duals here?

irm-codebase (changes requested):
Looks good, but we need more coverage on the tests. There is no HiGHS-specific test file or documentation mention. I'm assuming those will come later, and that they'll likely catch issues, so I mostly focused on architectural stuff.

Seems like we have a lot of duplicate code across the backends, which is resulting is slight divergences in behaviour. It might not be worth fixing at this time if we plan on moving to `linopy`, but I've still highlighted most of these cases. The core cause is that we have the logic reversed: instead of having known 'hooks' for backend-specific functionality, we have functions with unique names (e.g., `do_something_highs`). If we agreed on hook naming, most of this could be removed and the backends would be significantly shorter in code.

As always, nitpicks and nice-to-haves are just that, not implementation requests for this PR.

src/calliope/backend/expression_parser.py
sjpfenninger: It seems highspy backend objects don't all implement division, so this is a workaround (will be commented)

src/calliope/backend/highs_backend_model.py
sjpfenninger: Fixed (also other instances of the string)

src/calliope/backend/highs_backend_model.py
sjpfenninger: Applied

src/calliope/backend/highs_backend_model.py
sjpfenninger: Leaving this aside for now as part of a later backend cleanup

src/calliope/backend/highs_backend_model.py
sjpfenninger: Also leaving out for now; needs wider backend cleanup

src/calliope/backend/highs_backend_model.py
sjpfenninger: Also leaving out for now; needs wider backend cleanup

src/calliope/backend/highs_backend_model.py
sjpfenninger: Applied

src/calliope/backend/highs_backend_model.py
sjpfenninger: #890

src/calliope/backend/highs_backend_model.py
sjpfenninger: The bot did find another issue which is addressed:

`_from_highs_variable_bounds` as well as `_from_gurobi_variable_bounds` were returning a `pd.Series`. However, `np.frompyfunc(func, n, 2)` which is used by `_apply_func` expects a tuple, so this code was returning all NaNs. Testing didn't raise, because the tests didn't check for non-NaN values.

src/calliope/backend/highs_backend_model.py
sjpfenninger: Yes, this is now addressed similar to how the gurobi backend deals with it. Covered by a HiGHS specific test asserting `solve(warmstart=True)` re-solves optimally.

src/calliope/backend/highs_backend_model.py
sjpfenninger: Left as is, would be a backend harmonisation issue.

src/calliope/backend/highs_backend_model.py
sjpfenninger: Fixed and tests are added

src/calliope/backend/highs_backend_model.py
sjpfenninger: Fixed with Claude: `_shift_stale_indices` now pre-sorts the deleted indices once and uses `bisect` per object, so a deletion is O(n log d) rather than O(n·d).

src/calliope/backend/highs_backend_model.py
sjpfenninger: Docstrings added

src/calliope/backend/highs_backend_model.py
irm-codebase: Nice catch!

irm-codebase (dismissed):
Looks good!
Nice to see that deletions are not less expensive to run 👍 

The one thing missing is ~0.5% of coverage. The highs test file is only reaching ~86%.
i'd be OK in merging as is if we plan to shift to a harmonised backend (or `linopy`) in the near future.

src/calliope/backend/gurobi_backend_model.py
```diff
@@ -82,9 +82,6 @@ def add_parameter(  # noqa: D102, override
 
         self._add_to_dataset(name, values, "parameters", definition.model_dump())
 
-        if name not in self.math["parameters"]:
-            self.math = self.math.update({f"parameters.{name}": definition})
-
     def _add_variable(  # noqa: D102, override
```
irm-codebase, line 85: Can you explain why this was removed?

tests/backend/highs_backend_model_test.py
```diff
@@ -72,6 +72,13 @@ def test_new_build_get_parameter(self, simple_supply_highs):
         param = simple_supply_highs.backend.get_parameter("flow_in_eff")
         assert isinstance(param.item(), float)
 
+    def test_add_parameter_updates_math(self, simple_supply_highs_func):
+        """Parameters unknown to the math must be registered in the math on addition."""
+        backend = simple_supply_highs_func.backend
+        assert "foo" not in backend.math["parameters"].root
+        backend.add_parameter("foo", xr.DataArray(1), {})
+        assert "foo" in backend.math["parameters"].root
+
```
irm-codebase, line 81: Aha, I see. Makes sense, then.

irm-codebase (approved):
Looking good!
Feel free to merge this in 👍

PR #860, opened 2026-03-09
Allow making a tech inactive while it still has defined data

sjpfenninger:

There was an issue when disabling a tech with `active: False` while it still has parameter data, e.g. from tabular data. In this case the model fails with a KeyError. This makes `active: False` less useful as one cannot just mark something inactive without wiping all its data from the model too.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/preprocess/data_tables.py
+++ b/src/calliope/preprocess/data_tables.py
@@ -125,6 +125,8 @@ class DataTable:
             except KeyError:
                 continue
             for tech in techs_this_node:
+                if tech not in techs_incl_inheritance:
+                    continue
                 if (
                     techs_incl_inheritance[tech].get("base_tech", None)
                     == "transmission"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ For example, `model.inputs.flow_cap_max.attrs["default"]` is now only available 
 
 |fixed| Example notebook outputs no longer clutter search results in docs and search result example notebook pages are grouped under the top-level page header (#675).
 
+|fixed| Allow a tech to be set `active: False` even if data for it is defined (e.g. through a CSV file).
+
 ### Internal changes
 
 |fixed| example user-defined math tests that were missing global expressions in the resulting LP files.
--- a/tests/preprocess/data_sources_test.py
+++ b/tests/preprocess/data_sources_test.py
@@ -646,6 +646,14 @@ class TestDataTableNodeDict:
 
         assert node_dict == {}
 
+    def test_node_dict_skips_inactive_tech(self, table_obj):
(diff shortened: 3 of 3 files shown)


src/calliope/preprocess/data_tables.py
```diff
@@ -125,6 +125,8 @@ def node_dict(self, techs_incl_inheritance: AttrDict) -> AttrDict:
             except KeyError:
                 continue
             for tech in techs_this_node:
+                if tech not in techs_incl_inheritance:
+                    continue
```
irm-codebase, line 129: This is rather risky, as we unknowingly let any and all cases silently pass. This will likely lead to future issues.

In reality we should already know whether or not this specific tech is inactive. The real reason this fails is because we are losing context.

In `model.py`, we do:

```python
    model_data_factory = ModelDataBuilder(
        model_def.config.init,
        AttrDict(model_def.definition.model_dump(exclude_defaults=True)),  # The culprit
        math,
        tables,
    )
```

Since we exclude defaults, `active` is lost, and we are unable to check if something invalid 'sneaked by' within the code.

The long term fix is to finally remove the `TypedDict` redefinition of model schema in `model_data.py`. Our comments even say as much:

```python
# TODO: remove in favor of using the model def schema.  <<----- lookie here, lookie here
class ModelDefinition(TypedDict):
    """Uniform dictionary for model definition."""

    techs: AttrDict
    nodes: AttrDict
    data_definitions: NotRequired[AttrDict]
```

sjpfenninger: Fair enough, this is an ugly hack. But the issue is real, and one bumps into it fairly quickly when using new-style data tables, so it ought to be resolved for 0.7.0. I'm not sure if it's viable (effort-wise) to address the underlying cause right now, so a solution might be to keep the hack, or a slightly improved hack, and document this + the underlying problem in an issue to track the intended medium-term improvement?

irm-codebase: Agreed on the pragmatism.

I think for this PR it'd be enough to modify the code a bit (via a brief comment or improved naming) to make the goal of this like explicit, so that in the future we do not forget why it was added.

How about this?

```suggestion
                if tech not in techs_incl_inheritance:
                    # tech has likely been deactivated, do not process further
                    # FIXME: should check `active` in the pydantic class directly
                    continue
```

brynpickering: @sjpfenninger I'd be ok of pushing this fix through with the suggestion by @irm-codebase in the knowledge that we need to ultimately pass the pydantic model through without the dictificaiton.

brynpickering:
Closing in favour of #868

PR #864, opened 2026-04-17
Update array default filling timing

### LINKED ISSUE #862: Operate mode errors with `source_availability_supply` for `per_cap` technologies in a multicarrier model
### What happened?

Operate mode writes nan/inf `source_availability_supply` constraints for technologies with a `per_cap` source unit when there are multiple carriers in a model. This seems to be due to the `flow_cap` parameter being set to inf by default instead of the 0 value used for the variable in planning mode, but it seems to be defaulting the inf value for all model carriers, even carriers that the technology doesn't consume/produce.

```
source_availability_supply:
    description: >-
      Set the upper bound on, or a fixed total of,
      a `supply` technology's ability to consume its available resource.
    foreach: [nodes, techs, timesteps]
    where: "source_use AND (source_use_equals OR source_use_max)"
    equations:
      - where: "source_use_equals"
        expression: "source_use == source_use_equals * $source_scaler"
      - where: "NOT source_use_equals AND source_use_max"
        expression: "source_use <= source_use_max * $source_scaler"
    sub_expressions:
      source_scaler: &source_scaler
        - where: "source_unit==per_area"
          expression: "area_use"
        - where: "source_unit==per_cap"
          expression: "sum(flow_cap, over=carriers)"
        - where: "source_unit==absolute"
          expression: "1"
```
The error is specifically with the `per_cap` `source_scaler` sub-expression, where it seems that the `sum(flow_cap, over=carriers)` takes some default inf values for unused carriers and makes the sum come out to infinity. 

Below is a simple example model that encounters the issue:
```
templates:
  Transmission:
    base_tech: transmission
    carrier_in: Power
    carrier_out: Power
    name: Transmission
techs:
  PV:
    base_tech: supply
    carrier_out: Power
    color: '#f7ff00'
    name: PV
    source_unit: per_cap
  Test:
    base_tech: supply
    carrier_out: test
    color: '#f7ff00'
    name: Test
    source_unit: per_cap
  Demand:
    base_tech: demand
    carrier_in: Power
    name: Demand
    tag: null
 Transmission:
    flow_cap: 170018.09
    link_from: Supply
    link_to: Demand
    template: Transmission
nodes:
  Supply:
    latitude: 19.45721
    longitude: -154.94125
    techs:
      PV:
        flow_cap: 
          data: [144960.18]
          dims: [carriers]
          index: [Power]
        cost_flow_out:
          data: [0.01]
          dims: [costs, carriers]
          index: [[monetary, Power]]
      Test: null
  Demand:
    latitude: 19.53865
    longitude: -154.9312
    techs:
      Demand: null
```
With a `source_use_equals/max` timeseries on the PV tech, the `source_availability_supply` constraints will all have either inf or nan right hand sides (the Pyomo backend actually lets these constraints be built while Gurobi errors).

I've encountered this issue before with supplies that output multiple carriers, and manually setting the capacities for other carriers to avoids the error, but it seems to trigger even for single-carrier supplies if there is any other carrier in the model. Turning on the second supply with a second "test" carrier causes the errors on the PV tech even if that carrier is never used anywhere else in the model.

I've quickly tried setting a default_if_empty to 0 in that sum expression but it doesn't seem to stick. The workaround I would put in place for now would be to manually set a capacity on every `per_cap` technology for every carrier, defaulting it to 0, but this doesn't seem like expected behavior if there is some updates to the math for `source_availability_supply` that would resolve the issue.

- [ ] macOS
- [x] Windows
- [x] Linux

### Version

v0.7.0.dev7

### Relevant log output

```Shell

```

brynpickering:
Fixes #862 

I ended up trying to fill with default values as late as possible, just before applying arithmetic or comparisons. Since we probably want default values even when we don't go through these two processes (e.g. you just sum an array in a global expression, no other arithmetic), I apply the default value right at the end as well, once out of the expression parser.

I don't think there's one solution to this that will be super clean, but this works with current models and fixes #862 so is possibly the best right now.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/docs/user_defined_math/helper_functions.md
+++ b/docs/user_defined_math/helper_functions.md
@@ -46,6 +46,11 @@ nodes:
 
 Using `sum(..., over=)` in an expression allows you to sum over one or more dimensions of your component array (be it a parameter, decision variable, or global expression).
 
+!!! note
+    Array default values may enter into the summation if there are NaN values in the array that are not masked by the `where` string.
+    For instance, if you sum `flow_cap`, which has `inf` as the default value, you risk having a summation returning `inf`.
+    If this occurs, consider masking these values by applying the [`where`](#where) helper function to first strip out unwanted entries before the `sum` function is applied.
+
 ## select_from_lookup_arrays
 
 Some of our arrays in [`model.inputs`][calliope.Model.inputs] are not data arrays, but "lookup" arrays.
--- a/docs/user_defined_math/helper_functions.md
+++ b/docs/user_defined_math/helper_functions.md
@@ -47,8 +47,7 @@ nodes:
 Using `sum(..., over=)` in an expression allows you to sum over one or more dimensions of your component array (be it a parameter, decision variable, or global expression).
 
 !!! note
-    Array default values may enter into the summation if there are NaN values in the array that are not masked by the `where` string.
-    For instance, if you sum `flow_cap`, which has `inf` as the default value, you risk having a summation returning `inf`.
+    Invalid variable default values such as `.inf` or `.nan` can enter into the summation if they are not properly masked in the `where` string.
     If this occurs, consider masking these values by applying the [`where`](#where) helper function to first strip out unwanted entries before the `sum` function is applied.
 
 ## select_from_lookup_arrays
--- a/src/calliope/backend/expression_parser.py
+++ b/src/calliope/backend/expression_parser.py
@@ -120,9 +120,12 @@ class EvalArrayOrMath(EvalString):
                 `array` with NaNs filled with defaults, where array applied,
                 and dimensions broadcast against the where array (to ensure consistent array shapes).
         """
-        where_array = self.eval_attrs.where_array
-        masked_array = array.broadcast_like(where_array).where(where_array)
-        return masked_array
+        if not self.eval_attrs.apply_where:
+            return array
+        elif (where_array := self.eval_attrs.where_array).equals(xr.DataArray(True)):
+            return array
+        else:
+            return array.broadcast_like(where_array).where(where_array)
 
     # Math strings evaluate to strings.
     @overload
--- a/src/calliope/backend/expression_parser.py
+++ b/src/calliope/backend/expression_parser.py
@@ -120,12 +120,11 @@ class EvalArrayOrMath(EvalString):
                 `array` with NaNs filled with defaults, where array applied,
                 and dimensions broadcast against the where array (to ensure consistent array shapes).
         """
-        if not self.eval_attrs.apply_where:
-            return array
-        elif (where_array := self.eval_attrs.where_array).equals(xr.DataArray(True)):
-            return array
-        else:
-            return array.broadcast_like(where_array).where(where_array)
+        if self.eval_attrs.apply_where:
+            where_array = self.eval_attrs.where_array
+            if not where_array.equals(xr.DataArray(True)):
+                array = array.broadcast_like(where_array).where(where_array)
+        return array
 
     # Math strings evaluate to strings.
     @overload
--- a/src/calliope/backend/expression_parser.py
+++ b/src/calliope/backend/expression_parser.py
@@ -121,8 +121,8 @@ class EvalArrayOrMath(EvalString):
                 and dimensions broadcast against the where array (to ensure consistent array shapes).
         """
         where_array = self.eval_attrs.where_array
-        arr = self._apply_default(array).broadcast_like(where_array).where(where_array)
-        return arr
+        masked_array = array.broadcast_like(where_array).where(where_array)
+        return masked_array
 
     # Math strings evaluate to strings.
     @overload
@@ -328,7 +328,7 @@ class EvalSignOp(EvalArrayOrMath):
         return self.sign + self._eval("math_string")
 
     def as_array(self) -> xr.DataArray:  # noqa: D102, override
-        evaluated = self._apply_where_array(self._eval("array"))
+        evaluated = self._eval("array")
         if self.sign == "-":
             evaluated = -1 * evaluated
         return evaluated
@@ -830,9 +830,9 @@ class EvalUnslicedComponent(EvalArrayOrMath):
                     f"Trying to access a math component that is not yet defined: {self.name}. "
                     "If the referenced component is a global expression, set its `order` to have it defined first."
                 )
-
+        evaluated_filled = self._apply_default(evaluated)
         self.eval_attrs.references.add(self.name)
-        return evaluated
+        return evaluated_filled
 
 
 class EvalGenericString(EvalArrayOrMath):
--- a/src/calliope/backend/expression_parser.py
+++ b/src/calliope/backend/expression_parser.py
@@ -92,6 +92,40 @@ class EvalArrayOrMath(EvalString):
         The purpose of this is to be able to access the string/number value whether we query the array name or its data.
         """
 
+    def _apply_default(self, array: xr.DataArray) -> xr.DataArray:
+        """Apply default values to an array where relevant.
+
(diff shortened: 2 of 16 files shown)


docs/user_defined_math/helper_functions.md
```diff
@@ -46,6 +46,11 @@ nodes:
 
 Using `sum(..., over=)` in an expression allows you to sum over one or more dimensions of your component array (be it a parameter, decision variable, or global expression).
 
+!!! note
+    Array default values may enter into the summation if there are NaN values in the array that are not masked by the `where` string.
+    For instance, if you sum `flow_cap`, which has `inf` as the default value, you risk having a summation returning `inf`.
+    If this occurs, consider masking these values by applying the [`where`](#where) helper function to first strip out unwanted entries before the `sum` function is applied.
```
irm-codebase, line 51: Let's be kind to our future selves and avoid references that may become outdated.
Describing behaviour should be enough.

```suggestion
!!! note
    Invalid variable default values such as `.inf` or `.nan` can enter into the summation if they are not properly masked in the `where` string.
    If this occurs, consider masking these values by applying the [`where`](#where) helper function to first strip out unwanted entries before the `sum` function is applied.
```

src/calliope/backend/expression_parser.py
```diff
@@ -92,6 +92,40 @@ def as_array(self) -> xr.DataArray | list[xr.DataArray]:
...
+        Returns:
+            xr.DataArray: Array with default values applied, if the array name is defined in the math schema.
+        """
+        try:
+            default = self.eval_attrs.math.find(array.name)["default"]
+            if array.isnull().any() and pd.notna(default):
+                array = array.fillna(default)
+        except KeyError:
+            pass
+        return array
```
irm-codebase, line 110: This is a clear anti-pattern, as we leave the door open to silent errors.
Based on some debugging, this is hit only on these cases:
- `array.name` is empty
- `array.name` contains a number (e.g., 8760), for some reason.

both of these seem to come from the following call in `EvalOperatorOperand`:

```python
evaluated_operand = self._apply_where_array(
                operand.eval("array", self.eval_attrs)  # this guy breaks our array.name sometimes
            )
```

Below is an alternative that works without the `try / except`.
Ultimately, though, the real questions here are:
- when is this called for an invalid case, and can we avoid the call at all?
- why is `array.name` filled with junk sometimes?


```suggestion
        """
        if isinstance(array.name, str):
            default = self.eval_attrs.math.find(array.name)["default"]
            if array.isnull().any() and pd.notna(default):
                array = array.fillna(default)
        return array
```

src/calliope/backend/helper_functions.py
irm-codebase, line 1: There is a lot of  'attribute' protection added here.
Should this be a standard instead? When would this break our backend parsing?

irm-codebase (changes requested):
@brynpickering tried my best with the review, but I must admit that following parsing processes is quite difficult.

Ultimately, I think the biggest smell here is the clear reliance on `array.name`. Would this be something that might break user provided helper functions, if they are unaware of this 'quirck'?

src/calliope/backend/helper_functions.py
brynpickering: It should be needed actually. We're no longer storing anything in the attributes. I tried putting the default in the attrs to begin with but these helper functions had the habit of destroying them if no protections were in place. Hence why I now rely on the array name

src/calliope/backend/expression_parser.py
brynpickering: It needs to be a try/except as there are instances where `array.name` is a string but isn't a math component name but rather just an abritrary string (as generated by `EvalGenericString`). This string can't be a math component name (that's a rule of the parser), so it can't be confused with a math component here, but it could raise a KeyError.

brynpickering: Although, saying that. These string arrays should never make their way into being evaluated in these parsers (where `_apply_where_array` would be applied), so perhaps it is safest to go for this conditional.

irm-codebase: I like the new approach! Only looking for defaults where relevant is elegant, IMO.

brynpickering:
@irm-codebase after a few changes I came to the conclusion that the decision me made offline cannot work. 

We discussed filling NaNs in `EvalUnslicedComponent.eval` and applying `where` masking _after_ evaluation. 

This leads to `ZeroDivisionError` in cases where the default value is `0` and there is a division in the math. `xarray` cannot handle NaN-filling in these cases and we cannot apply `da.where(da != 0)` before the division since pyomo objects (even parameters!) cannot be subject to comparison operators, as would be the case inside the `where`.

I tried also applying `where` in `EvalUnslicedComponent.eval`, after filling with the default value, but this leads to other issues in the `gurobi` backend in which NaNs get attached as coefficients of decision variables unexpectedly. 

This means that we can fill NaNs in `EvalUnslicedComponent.eval` but we have to apply `where` before applying arithmetic or comparisons. 

The one edge case we get from this is if we have a division inside a function in which `ignore_where=True` since defaults would not be replaced by NaNs in advance of arithmetic application. We have this for functions that need to access data in other parts of the array, so you cannot apply the `where` string straight away. For example, if you set a conditional that in the first timestep you take the final timestep's value then the `where` would be keeping only the first timestep and the final timestep would become NaN when you want to access it 🤪. So you apply the function first without any masking and _then_ mask. The edge case is then if you call `roll(array / parameter, timesteps=1)` and parameter has a default of zero. This is easily circumvented by calling `roll(array, timesteps=1) / roll(parameter, timesteps=1)` but we'd need a troubleshooting item for this!

tests/backend/parsing_test.py
```diff
@@ -1141,8 +1141,13 @@ def test_parse_objective_dict_empty_eq1(self, objective_obj, eval_where_args):
     def test_parse_objective_dict_evaluate_eq2(
         self, objective_obj, eval_where_args, eval_expr_args
     ):
+        """Expression is `sum(only_techs, over=[techs]) + 1`.
+
+        Since we fill with default values in the sum, the NaN `only_techs` value gets included, leading to a sum of 11.
+        """
```
irm-codebase: For maintainability, let's remove the code from the comment. We only need to explain behaviour.

```suggestion
        """Since we fill with default values in the sum, the NaN `only_techs` value gets included, leading to a sum of 11.
        """
```

irm-codebase (dismissed):
Consider this approved, but please look at the two comments I made before merging.

src/calliope/backend/expression_parser.py
```diff
@@ -120,9 +120,12 @@ def _apply_where_array(self, array: xr.DataArray) -> xr.DataArray:
...
         """
-        where_array = self.eval_attrs.where_array
-        masked_array = array.broadcast_like(where_array).where(where_array)
-        return masked_array
+        if not self.eval_attrs.apply_where:
+            return array
+        elif (where_array := self.eval_attrs.where_array).equals(xr.DataArray(True)):
+            return array
+        else:
+            return array.broadcast_like(where_array).where(where_array)
```
irm-codebase: Nitpick: this is a bit hard to read, and the assignment operator is sneaky.
Here is a shorter version with a clear exit and the same execution cost.

```suggestion
        if self.eval_attrs.apply_where:
            where_array = self.eval_attrs.where_array
            if not where_array.equals(xr.DataArray(True)):
                array = array.broadcast_like(where_array).where(where_array)
        return array
```

irm-codebase (dismissed):
Overall the PR looks good! I only identified a minor improvement.

irm-codebase (dismissed):
Looking good!

Feel free to merge as soon as the CI passes.

irm-codebase (approved):

PR #865, opened 2026-05-08
Docs: restructuring

sjpfenninger:
- I've restructured the navigation and enabled the `navigation.tabs` to get a top navbar. Various things moved around, "Advanced" is renamed to "How to" (based on the logic that it's mostly a collection of "how to do X" mini-articles, their titles could be tweaked to match the "How to .... X" framing, and "Advanced" is a bit of a meaningless label).
- "Getting started" is followed by "Examples & tutorials" in the nav, since that's also the order we recommend new users to follow.
- I also modified the landing page to add a "map" of what is where, currently near the top (could be modified, if we think the idea works, or moved further down/elsewhere). Once the current pre-release admonition goes away, this would be at the very top.
- I didn't systematically touch the underlying folder structure, just moved things around in the TOC as a first proposal for how the structure could look. Since I have another doc-fixes branch (and a list of additional fixes to work through on my end) that is based on the current file structure I'd prefer to merge that in first anyway, before rearranging file locations.
- I thought "Math gallery" could also move into the examples, but at the same time it seems sufficiently different to warrant a top-level item (plus the helper script would need to be updated to allow it to live at level 2). Keeping it at the top reduces the click distance to get to entries in the gallery.


irm-codebase:
Broad list of suggested changes:
- [x] The following files should not show up in the search (too many false positives).
    - [x] version history
- [ ] Cyclic references to the docs should be removed (e.g., at the bottom of [this](https://calliope--865.org.readthedocs.build/en/865/examples/urban_scale/notebook/)).
- [ ] "Running in Python" and "Running in the command line" in "Building blocks"  should be moved to the "Running a model" section in "Getting started"
- [ ] Consider a header for "Math" with the following:
    - [ ] In "Building blocks": "Build-in base math", "Other built-in math" and "User-defined math". In particular, "User-defined math" seems pretty important, and yet it is hidden at a 3rd layer level.
    - [ ] "Math Gallery" could be a subsection.
- [x] Reference section
    - [x] Consider a "Schemas" subsection for all the schemas, with more homogenous naming (e.g., "Calliope Model Definition" is also a schema, but it does not say it in its name. Suggested names and ordering:
        - Model configuration
        - Model definition
        - Data tables
        - Model math
- [ ] Link to the Zulip channel subscribe page.

brynpickering:
From my side:

- [x] For examples, something like what they do [in pypsa docs](https://docs.pypsa.org/latest/examples/examples/) might be useful. This could allow for math gallery to go into examples as they would then be a tile group in the overview.
- [ ] "How to" -> "FAQ"? We should maybe then split that into troubleshooting and non-troubleshooting FAQs.
- [x] Maybe only the user-defined math part should be in building blocks (and named simply "math") and the built-in math definitions should be in reference

docs/getting_started/analysing.md
irm-codebase, line 60: This link is botched.

irm-codebase, line 29: Botched xarray link

docs/index.md
```diff
@@ -1,17 +1,37 @@
 # Calliope: energy system modelling made simple
```
irm-codebase, line 2: Nitpick: shouldn't this be energy system**s**?

It sounds off to me, but I'm not a native speaker.

irm-codebase (changes requested):
Gave it a quick glance. Looks good!

Some `xarray` links appear to be broken, though. I'd consider either repairing or (my preference) just removing them. Just link to `xarray` docs generally, not specific functionality that might change.

docs/getting_started/analysing.md
sjpfenninger: Fixed as of 0a0817c

sjpfenninger: Fixed as of 0a0817c

irm-codebase (approved):
Seems good!
CI seems broken, but this code is not functional to those failures, it seems.

PR #868, opened 2026-05-29
Replace AttrDict with pydantic models in model data builder; add tests

brynpickering:
Supersedes #860 

- No more attrdict in the model data builder
- pydantic model dumps on update exclude unset fields
- runtime timings are now a pydantic dict RootModel so contents are seen as set when updating
- IndexedData is now pre-set with empty lists for dims and index items
- Transmission techs are handled separately to other techs in pydantic model

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


src/calliope/preprocess/model_data.py
```diff
@@ -325,59 +367,71 @@ def _input_data_dict_to_array(
...
 
         Returns:
             xr.Dataset: Dataset with arrays indexed over (at least) the input `dim_name`.
         """
+        exclude = {"techs"} if dim_name == "nodes" else None
         input_data_ds = xr.Dataset()
-        for idx_name, idx_inputs in def_dict.items():
+        for idx_name, idx_inputs in definition.root.items():
+            if idx_inputs is None:
+                continue
```
irm-codebase: This effectively does nothing, as techs should've been inherited at this point via `_inherit_defts`.
Skipping seems risky, as it silently drops invalid cases and can lead to us not detecting regressions. This is a case were allowing python to raise an error is OK, IMO, as it would stem from actual regressions on our end and not the user's.

Also, mypy or an IDE might be throwing a warning due to our decision to allow `None` in `CalliopeTechs`. Handing those cases by replacing it with `{}` at the schema level should solve it (which i suggest elsewhere in this review).

src/calliope/preprocess/model_data.py
```diff
@@ -325,59 +367,71 @@ def _input_data_dict_to_array(
...
-    def _definition_dict_to_ds(
+    @overload
+    def _definition_to_ds(
+        self, definition: CalliopeNodes, dim_name: Literal["nodes"]
+    ) -> xr.Dataset: ...
+
+    @overload
+    def _definition_to_ds(
+        self, definition: CalliopeTechs, dim_name: Literal["techs"]
+    ) -> xr.Dataset: ...
```
irm-codebase: What's the reason behind these overloads?
They do not seem to help `mypy` or  `Pylance` in any way, just seem to bloat the code...
```suggestion
```

src/calliope/preprocess/model_data.py
```diff
@@ -150,6 +132,73 @@ def build(self):
...
+
+    @overload
+    @staticmethod
+    def _update_def(
+        to_update: CalliopeDictModel,
+        update_def: CalliopeDictModel,
+        key: None = None,
+        overwrite: bool = False,
+    ) -> CalliopeDictModel: ...
+
```
irm-codebase: A bit too bloaty. I suggest removing this in favour of the fix below.

```suggestion
```

src/calliope/preprocess/data_tables.py
```diff
@@ -73,34 +73,38 @@ def drop(self, name: str):
         """
         self.dataset = self.dataset.drop_vars(name, errors="ignore")
 
-    def tech_dict(self) -> tuple[AttrDict, AttrDict]:
+    def tech_def(self) -> tuple[CalliopeTechs, CalliopeTechs]:
         """Create a dummy technology definition dictionary from the dataset's "techs" dimension.
 
         This definition dictionary will ensure that the minimal YAML content is still possible.
 
         This function should be accessed _before_ `self.node_dict`.
```
irm-codebase: Comment is invalid.
I suggest adjusting it to avoid naming code explicitly, and instead explain functionality.

```suggestion
        This function should be accessed _before_ node definition.
```

src/calliope/preprocess/data_tables.py
```diff
@@ -73,34 +73,38 @@ def drop(self, name: str):
...
-                base_tech_dict = self.dataset[param].to_dataframe().dropna().T.to_dict()
-                base_tech_data.union(base_tech_dict)
+        params = self.PARAMS_TO_INITIALISE_YAML.intersection(self.dataset.data_vars)
+        if params:
+            df = self.dataset[params].to_dataframe().dropna(how="all").T
+            base_tech_dict = {
+                col: series.dropna().to_dict() for col, series in df.items()
+            }
+        else:
+            base_tech_dict = {}
```
irm-codebase: Nitpick: you can avoid the `else` by pre-defining `base_tech_dict = {}` before the `if`.
Same outcome, just a tiny bit more succinct.

src/calliope/preprocess/data_tables.py
```diff
@@ -73,34 +73,38 @@ def drop(self, name: str):
...
 
         return tech_dict, base_tech_data
 
-    def node_dict(self, techs_incl_inheritance: AttrDict) -> AttrDict:
+    def node_def(self, techs_incl_inheritance: CalliopeTechs) -> CalliopeNodes:
         """Create a dummy node definition dictionary from the dimensions defined across all data tables.
 
         This definition dictionary will ensure that the minimal YAML content is still possible.
 
         This function should be run _after_ `self._update_tech_def_from_data_table`.
```
irm-codebase: Invalid docstring.

```suggestion
        This definition dictionary will ensure that the minimal YAML content is still possible.

        This function should be run _after_ tech definition.
```

src/calliope/preprocess/data_tables.py
```diff
@@ -73,34 +73,38 @@ def drop(self, name: str):
...
 
         Args:
-            techs_incl_inheritance (AttrDict):
+            techs_incl_inheritance (CalliopeTechs):
                 Technology definition dictionary which is a union of any YAML definition and the result of calling `self.tech_dict` across all data tables.
```
irm-codebase: Old docstring

```suggestion
            techs_incl_inheritance (CalliopeTechs):
                Technology definition dictionary which is a union of any YAML definition and the result of solving tech definition across all data tables.
```

src/calliope/preprocess/data_tables.py
```diff
@@ -112,32 +116,35 @@ def node_dict(self, techs_incl_inheritance: AttrDict) -> AttrDict:
...
-        for node, tech_dict in node_tech_dict.items():
+        node_tech_dict = CalliopeNodes.model_validate(
+            {i: {"techs": {}} for i in self.dataset.nodes.values}
+        )
+        for node, tech_def in node_tech_dict.root.items():
             try:
                 techs_this_node = is_defined[is_defined].xs(node, level="nodes").index
             except KeyError:
                 continue
             for tech in techs_this_node:
```
irm-codebase: Two things:
- `tech_def` in the first for is not used anywhere.
- The try / except seems avoidable by preparing the definition beforehand. It should also be more efficient, too.

```suggestion
        node_tech_dict = CalliopeNodes.model_validate(
            {i: {"techs": {}} for i in self.dataset.nodes.values}
        )
        defined_node_techs = is_defined[is_defined]
        techs_by_node = {}
        if not defined_node_techs.empty:
            techs_by_node =
                defined_node_techs.index
                .to_frame(index=False)
                .groupby("nodes")["techs"]
                .apply(list)
                .to_dict()
            )
        for node in node_tech_dict.root:
            for tech in techs_by_node.get(node, []):
```

src/calliope/preprocess/data_tables.py
```diff
@@ -112,32 +116,35 @@ def node_dict(self, techs_incl_inheritance: AttrDict) -> AttrDict:
...
-                if (
-                    techs_incl_inheritance[tech].get("base_tech", None)
-                    == "transmission"
-                ):
+                if tech not in techs_incl_inheritance.root:
+                    continue
```
irm-codebase, line 137: Shouldn't this be a hard failure?
This seems to be silently dropping undefined technologies, which would be quite bad.
Am I wrong?

src/calliope/schemas/dimension_data_schema.py
```diff
@@ -33,32 +32,50 @@ class IndexedData(CalliopeBaseModel):
...
+    index: list[UniqueList[IndexValue]] = Field(default_factory=lambda: [[]])
     """
     Model dimension members to apply the data value(s) to.
     If an array of arrays, sub-arrays must have same length as number of `dims`.
     """
 
+    @field_validator("index", mode="before")
+    @classmethod
+    def listify_index(cls, value) -> Self:
+        """Ensure index and dims are consistent."""
```
irm-codebase: Return type is wrong.

```suggestion
    def listify_index(cls, value) -> list[list]:
        """Ensure index and dims are consistent."""
```

src/calliope/schemas/dimension_data_schema.py
```diff
@@ -33,32 +32,50 @@ class IndexedData(CalliopeBaseModel):
...
+            value = [[value]]
+        if isinstance(value, list) and not any(isinstance(i, list) for i in value):
+            value = [[v] for v in value]
+
+        return value
+
+    @field_validator("dims", mode="before")
+    @classmethod
+    def listify_dims(cls, value) -> Self:
+        """Ensure index and dims are consistent."""
```
irm-codebase: Return type is wrong.

```suggestion
    def listify_dims(cls, value) -> list:
        """Ensure index and dims are consistent."""
```

src/calliope/schemas/dimension_data_schema.py
```diff
@@ -86,6 +103,33 @@ class CalliopeTech(DimensionData):
     """
 
 
+class CalliopeTransmissionTech(DimensionData):
+    """Calliope's transmission technology dimension schema."""
```
irm-codebase: This should inherit `CalliopeTech` as parent.
Otherwise, `isinstance(transmission, CalliopeTech)` will fail, which is likely not what we want.

```suggestion
class CalliopeTransmissionTech(CalliopeTech):
    """Calliope's transmission technology dimension schema."""
```

This would also make specifying `CalliopeTransmissionTech` in `CalliopeTechs` below unnecessary.

src/calliope/schemas/dimension_data_schema.py
```diff
@@ -109,20 +153,23 @@ def check_dependent_definitions(self) -> Self:
...
-
-    root: dict[AttrStr, CalliopeTech | None] = Field(default_factory=dict)
+    @field_validator("techs", mode="before")
+    @classmethod
+    def check_techs(cls, value: dict | None) -> dict:
+        """Ensure techs are in correct format."""
+        if value is None:
+            return {}
+        else:
+            return value
```
irm-codebase: Shouldn't this live at the `CalliopeTechs` level instead of here?
This smells like a class being aware of functionality that does not belong to it...

src/calliope/schemas/general.py
```diff
@@ -157,7 +157,7 @@ def update(self, update_dict: dict, deep: bool = False) -> Self:
         new_dict: dict = {}
         # Iterate through dict to be updated and convert any sub-dicts into their respective pydantic model objects.
         # Wrapped in `AttrDict` to allow users to define dot notation nested configuration.
-        for key, val in AttrDict(update_dict).items():
+        for key, val in AttrDict(update_dict).as_dict().items():
             key_class = getattr(self, key, None)
```
irm-codebase: Why the need to back-convert to a regular dictionary?

src/calliope/schemas/runtime_attrs_schema.py
```diff
@@ -6,7 +6,13 @@
...
 from calliope import _version
 from calliope.schemas.config_schema import Resamples, Subsets
-from calliope.schemas.general import AttrStr, CalliopeBaseModel
+from calliope.schemas.general import AttrStr, CalliopeBaseModel, CalliopeDictModel
+
+
+class CalliopeTimings(CalliopeDictModel):
+    """Calliope runtime time logging."""
+
+    root: dict[str, float] = Field(default_factory=dict)
```
irm-codebase, line 15: Why create this unique class if all it'll do is use a dictionary anyhow?

src/calliope/preprocess/model_data.py
```diff
@@ -150,6 +132,73 @@ def build(self):
...
+            pre_update = getattr(to_update, key).model_dump(**kwargs)
+        else:
+            pre_update = to_update.model_dump(**kwargs)
+        update_dict = update_def.update(pre_update).model_dump(**kwargs)
+
+        if key is not None:
+            updated = to_update.update({key: update_dict})
+        else:
+            updated = to_update.update(update_dict)
+        return updated
```
irm-codebase: Most of `_update_def` seems like misplaced functionality: a fair portion of this could be moved to `schemas.general` as an expansion of the core functionality of our schema classes, and then later calling those methods directly (for either the `DictModel` or `BaseModel` case).

The code that calls this function is already aware of the specific class/case it should call, as otherwise this function would raise an error.

Doing this will likely:
- reduce overall code by making this function irrelevant and necessitating less if/else clauses
- avoid the need for the two defensive tests seen here, as using `keys` for `CalliopeDictModel` is impossible by design.

src/calliope/preprocess/model_data.py
```diff
@@ -389,44 +443,60 @@ def _prepare_input_data_dict(
...
+        **connected_dims: str,
+    ) -> CalliopeTechs: ...
+
+    @overload
+    def _inherit_defs(
+        self,
+        dim_name: Literal["nodes"],
+        dim_def: CalliopeNodes | None = None,
+        **connected_dims: str,
+    ) -> CalliopeNodes: ...
```
irm-codebase: Another case were I'd suggest removing the overrides in favour of just passing `CalliopeDictModel` as the type for `dim_def`.

src/calliope/preprocess/model_data.py
```diff
@@ -459,37 +529,43 @@ def _inherit_defs(
...
-            dim_dict = self.model_definition[dim_name]
+        updated_defs = CalliopeTechs() if dim_name == "techs" else CalliopeNodes()
+        if dim_def is None:
+            dim_def = self.model_definition[dim_name]
 
-        for item_name, item_def in dim_dict.items():
+        for item_name, item_def in dim_def.root.items():
             if item_def is None:
-                item_def = AttrDict()
+                item_def = CalliopeTech() if dim_name == "techs" else CalliopeNode()
```
irm-codebase: This is another case where the schema could've handled the re-definition from `None` to an empty default.

src/calliope/preprocess/model_data.py
```diff
@@ -542,20 +611,26 @@ def _links_to_node_format(self, active_node_dict: AttrDict) -> AttrDict:
...
+            )
+            node_to_data = link_data.model_dump(exclude=exclude_to, exclude_unset=True)
+            link_tech_def = link_tech_def.update(
                 {
-                    node_from: {link_name: node_from_data},
-                    node_to: {link_name: node_to_data},
+                    node_from: {"techs": {link_name: node_from_data}},
+                    node_to: {"techs": {link_name: node_to_data}},
                 }
             )
```
irm-codebase: Just for my personal understanding of the update: why the need to pass `exclude_from` and `exclude_to`  to `model_dump` even if `one_way` to is not `True`?

irm-codebase (changes requested):
Overall, a great update!

General comments:
- I think `model_data.py` has some functionality that does not belong to it.
- I do not see the reasoning behind a lot of the `overrides` (and the bot seems to think similarly).

src/calliope/schemas/runtime_attrs_schema.py
brynpickering: Because it inherits the pydantic functionality, namely the knowledge of whether an entry has been set by the user or not. Without it, when doing `model_dump(exclude_unset=True)` in `update`, the timings defined by adding them to the timings dictionary will not be seen by pydantic as "setting" this field. Instead, pydantic will see the "set" field as still being an empty dictionary. Making sure there are no pure dicts in our schemas and only dict RootModels is the safest way to avoid this causing us issues.

src/calliope/preprocess/data_tables.py
brynpickering: Perhaps. I think was just a simple way to handle technologies deactivated at the `techs` level. I should triple check this as I would tend to agree with you that it seems odd.

src/calliope/preprocess/model_data.py
brynpickering: Reasonable assessment. We're overwriting the base pydantic `update` method so we can't change the function signture without creating issues. But we could have a separate updater method that doesn't allow overwriting, to replace this one.

src/calliope/preprocess/model_data.py
brynpickering: There's a benefit to the `overloads` as it aids maintenance since you know what to expect back _and_ you known that `techs` -> `CalliopeTechs`, `nodes` -> `CalliopeNodes`. The only way to truly separate it would be to have `_inherit_tech_defs` and `_inherit_node_defs` as two separate methods

src/calliope/preprocess/model_data.py
brynpickering: We remove `node_from` and `node_to` from the final dataset in favour of defining the tech at those two nodes separately.

src/calliope/preprocess/model_data.py
brynpickering: > They do not seem to help mypy or Pylance in any way, just seem to bloat the code...

They do help writing code as you get type hints and autocomplete that are appropriate the expected input. That is, if you set `dim_name` as `techs` then you know the `definition` arg will be `CalliopeTechs` and its `root` contents will be of type `CalliopeTech`. It's a form of implicit documentation.

src/calliope/schemas/dimension_data_schema.py
brynpickering: > This would also make specifying CalliopeTransmissionTech in CalliopeTechs below unnecessary.

I don't think that's possible. If it defines something as `CalliopeTransmissionTech` and it _isn't_ in `CalliopeTechs` then it will raise an error as it has a `CalliopeTransmissionTech` that it isn't allowed in the tech dict

src/calliope/schemas/runtime_attrs_schema.py
irm-codebase: Got it. In that case it makes sense to follow this approach.

src/calliope/preprocess/model_data.py
irm-codebase: Fair. I guess the overload is ok, but it still feels a bit overcomplicated...
I'd consider using `isinstance` and dropping `dim_name`. It'd achieve the same with less lines, and it'd allow adding guardrails for improper use more easily (although I do not think that's necessary).

```python
    def _definition_to_ds(self, definition: CalliopeNodes | CalliopeTechs) -> xr.Dataset:
        if isinstance(definition, CalliopeTech):
```

src/calliope/preprocess/data_tables.py
irm-codebase: yeah, it has the smell of a future bug...

src/calliope/preprocess/model_data.py
irm-codebase: Sounds good! I do not remember why we overwrote `update`, to be honest...
It might have been better to just add our own version of it. If that's possible, we can just replace our calls with that version and add the overwrite block as a separate parameter. Otherwise, a new method at the schema level is also fine, although it is repetition.

src/calliope/preprocess/model_data.py
irm-codebase: There's an alternative that (I think) would fit this nicely (although I think splitting it two would also be OK). Python supports exclusive typing since 3.11, I think.

Here is some silly code I have to validate dimensionless `pint` fractions or floats.
- float returns float
- Quantity returns Quantity
- mypy "sees" both as exclusive to one another.

```python
def validate_fraction[T: (float, Quantity)](value: T, *, name: str = "fraction") -> T:
    """Validate a dimensionless value bounded by 0 and 1."""
    fraction = value.to("") if isinstance(value, Quantity) else value
    if not 0 <= fraction <= 1:
        raise ValueError(f"Impossible {name} {value!r}.")
    return value
```

src/calliope/backend/helper_functions.py
```diff
@@ -692,6 +689,65 @@ def as_array(self, array: xr.DataArray, condition: xr.DataArray) -> xr.DataArray
...
+        Returns:
+            xr.DataArray: `array` with only the selected member of the dimension.
+
+        Examples:
+            To select a specific node from a `flow` variable, you would do the following:
+
+
+            1. Define `node_1` as an array whose data is node names and whose index is _not_ the `nodes` dimension.
+               It could, for instance, be the techs dimension:
+               ```yaml
```
irm-codebase, line 729: Nitpick:
The example is good, but I suggest changing the name `node_1`. It's a bit confusing (this does not relate to a specific node, after all). Even something as blunt as `special_cases` would help readers understand when this is useful.

src/calliope/preprocess/model_data.py
```diff
@@ -210,32 +142,41 @@ def init_from_data_tables(self, data_tables: Iterable[data_tables.DataTable]):
...
                 lookup_dim = param_config.pivot_values_to_dim
                 if lookup_dim is not None:
                     lookup_def = data_table.lookup_def_from_param(param, lookup_dim)
-                    self.tech_data_from_tables = self._update_def(
-                        self.tech_data_from_tables, lookup_def, overwrite=True
+                    self.tech_data_from_tables = self.tech_data_from_tables.update(
+                        lookup_def
                     )
+                    # We drop it for now; it'll come back via YAML dict later.
                     data_table.drop(param)
```
irm-codebase, line 161: For my understanding:

Am I correct in thinking that the next `for` loop is what "adds this back"?
In any case, it's OK to leave for now, but we should make note to improve this in the future. It's a bit convoluted.

src/calliope/schemas/general.py
```diff
@@ -69,30 +69,42 @@ def __rich_repr__(self):
...
             key_class = self.root.get(key, None)
             if isinstance(key_class, CalliopeBaseModel):
-                new_dict[key] = key_class.update(val, deep=deep)
+                new_dict[key] = key_class.update(val, deep=deep, overwrite=overwrite)
             elif isinstance(key_class, CalliopeListModel):
-                new_dict[key] = key_class.update(val)
+                if overwrite:
+                    new_dict[key] = key_class.update(val)
+                else:
+                    continue
```
irm-codebase, line 104: What behaviour should we have when requesting updates on pre-existing items while `not overwrite`? `AttrDict` would've thrown an error here instead of silently passing.

src/calliope/schemas/general.py
```diff
@@ -144,33 +156,60 @@ def __getitem__(self, item):
...
             key_class = getattr(self, key, None)
             if isinstance(key_class, CalliopeBaseModel | CalliopeDictModel):
-                new_dict[key] = key_class.update(val, deep=deep)
+                new_dict[key] = key_class.update(val, deep=deep, overwrite=overwrite)
             elif isinstance(key_class, CalliopeListModel):
-                new_dict[key] = key_class.update(val)
+                if overwrite:
+                    new_dict[key] = key_class.update(val)
+                else:
+                    continue
```
irm-codebase, line 203: Same question as above: should this `raise` when an update is requested on pre-existing items and `not overwrite`?

src/calliope/schemas/math_schema.py
```diff
@@ -97,7 +97,7 @@ def dtype(self) -> Literal["float"]:
 class Lookup(MathComponent):
     """Schema for named lookup arrays."""
 
-    default: AttrStr | float | int | bool = float("nan")
+    default: str | float | int | bool = float("nan")
     """The default value for the lookup, if not set in the data."""
```
irm-codebase, line 101: Just out of curiosity: why remove `AttrStr` here?

It was there to prevent oddball names, but perhaps we should be more permissive?

irm-codebase (dismissed):
Consider this approved!

I have a couple of questions and some nitpicks. Overall, good to go.

src/calliope/preprocess/model_data.py
brynpickering: No, it's embedded into the model definition pydantic model and then re-introduced via that route later. Safer than adding it now as a table when we need some of that context for decisions made elsewhere in the code.

src/calliope/schemas/math_schema.py
brynpickering: I can revert that, yeah.

src/calliope/schemas/general.py
brynpickering: We have a different default behaviour here compared to AttrDict - overwrite is the default and not overwrite is the exception (opposite in AttrDict). I want to allow partial merges (update without overwrite) without having to handle an exception. Yes, it silently passes but the main reason I want it available is for cases where this would be the case. Otherwise I have to update A with B (allow overwrite) and then B with A (allow overwrite), which is what we were doing before with AttrDict. This is just one step: update B with A (no overwrite).

I can add a debug statement for requesting updates on pre-existing.

irm-codebase (approved):

PR #869, opened 2026-05-29
Refactor `active`

### LINKED ISSUE #725: Allow to use `active` in data tables
### What can be improved?

I would like to be able to use `active` from within data tables. There might be reasons for that to be forbidden because of implementation details. But as a user, I cannot see a reason why active is not allowed in data tables. 

As I am working with models at different spatial resolution, being able to provide all data describing the graph compactly would be of great use. If I could use `active` from within data tables, I could drop the yaml file describing specific links or nodes.

As an example, I currently have to spread data defining each link in two separate places.
**links.csv**
```
techs,distance,from,to
ESP_to_FRA,8,ESP,FRA
ESP_to_PRT,3,ESP,PRT
```
and **links.yaml**
```
techs:
  ESP_to_FRA:
    active: True
  ESP_to_PRT:
    active: True
```
Similar for nodes. Because of this, I need to use Jinja templates to create the yaml, instead of providing all data with spatial granularity in one or a few data tables.

**update** Limited the issue to `active`, as `template` seems to be supported in data tables already, as indicated in the comments.

### Version

v0.7.0.dev4

brynpickering:
Closes #725

Cleans up the internals, makes it possible to define `active` in a data table, and adds config to keep deactivated techs/nodes around in the final input dataset, in case there is a desire to reactivate components later.

Comes with the risk of model bloat, something I haven't tested for.

## TODO
- [ ] Test model size on disk compared to `main` on large model when there are lots of techs/nodes deactivated
- [ ] Add docs

- `definition_matrix` table -> `active` (no carrier dim)
- active allowed in data tables
- option to remove deactivated at init

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


docs/user_defined_math/components.md
```diff
@@ -110,7 +110,7 @@ variables:
 These are not required, but are useful metadata for later reference.
 1. It can have a top-level `foreach` list and `where` string.
 Without a `foreach`, it becomes an un-indexed variable.
-Without a `where` string, all valid members (according to the `definition_matrix`) based on `foreach` will be included in this decision variable.
+Without a `where` string, all valid members (according to the `active`) based on `foreach` will be included in this decision variable.
```
irm-codebase: Small tuning so that this reads a bit more natural.

```suggestion
Without a `where` string, all `active` members in `foreach` will be included in this decision variable.
```

docs/user_defined_math/components.md
```diff
@@ -148,7 +148,7 @@ Global expressions are by no means necessary to include, but can make more compl
 These are not required, but are useful metadata for later reference.
 1. It can have a top-level `foreach` list and `where` string.
 Without a `foreach`, it becomes an un-indexed expression.
-Without a `where` string, all valid members (according to the `definition_matrix`) based on `foreach` will be included in this expression.
+Without a `where` string, all valid members (according to the `active`) based on `foreach` will be included in this expression.
```
irm-codebase: Another small tuning for readability.

```suggestion
Without a `where` string, all `active` members in `foreach` will be included in this decision variable.
```

docs/user_defined_math/components.md
```diff
@@ -209,7 +209,7 @@ constraints:
 These are not required, but are useful metadata for later reference.
 1. It can have a top-level `foreach` list and `where` string.
 Without a `foreach`, it becomes an un-indexed constraint.
-Without a `where` string, all valid members (according to the `definition_matrix`) based on `foreach` will be included in this constraint.
+Without a `where` string, all valid members (according to the `active`) based on `foreach` will be included in this constraint.
```
irm-codebase:
```suggestion
Without a `where` string, all `active` members in `foreach` will be included in this decision variable.
```

docs/user_defined_math/components.md
```diff
@@ -242,7 +242,7 @@ piecewise_constraints:
 This is not required, but is useful metadata for later reference.
 1. It can have a top-level `foreach` list and `where` string.
 Without a `foreach`, it becomes an un-indexed constraint.
-Without a `where` string, all valid members (according to the `definition_matrix`) based on `foreach` will be included in this constraint.
+Without a `where` string, all valid members (according to the `active`) based on `foreach` will be included in this constraint.
```
irm-codebase:
```suggestion
Without a `where` string, all `active` members in `foreach` will be included in this decision variable.
```

docs/user_defined_math/components.md
```diff
@@ -302,7 +302,7 @@ postprocessed:
    These are not required, but are useful metadata for later reference.
 1. It can have a top-level `foreach` list and `where` string.
    Without a `foreach`, it becomes an un-indexed expression.
-   Without a `where` string, all valid members (according to the `definition_matrix`) based on `foreach` will be included in this expression.
+   Without a `where` string, all valid members (according to the `active`) based on `foreach` will be included in this expression.
```
irm-codebase:
```suggestion
   Without a `where` string, all `active` members in `foreach` will be included in this decision variable.
```

src/calliope/backend/helper_functions.py
```diff
@@ -244,6 +244,14 @@ class Defined(ParsingHelperFunction):
...
     ALLOWED_IN = ["where"]
 
+    def __init__(  # noqa: D107, override
+        self, return_type: Literal["array", "math_string"], attrs: "EvalAttrs"
+    ):
+        super().__init__(return_type, attrs)
+        self._defined = self._attrs.input_data.active & (
+            self._attrs.input_data.carrier_in | self._attrs.input_data.carrier_out
+        )
+
```
irm-codebase: Could you detail why this became necessary?
The previous version of this was agnostic to the math (for what I can see).
This new version breaks this and will fail if the math renames or replaces 'carrier_in' or 'carrier_out'.

Additionally, seems like this function is not even used any more in our examples. The one that claims to use it (`docs/user_defined_math/examples/net_import_share.yaml`) has a depreciated preamble.

Some suggestions:
- drop this helper?
- tune it slightly to accept a list of names and check that
    1. they are all `variables` or `parameters`
    2. they all have the same dimensionality (to make the comparison stable), and make `self._defined` carry that comparison instead.
- rename it to `DefinedCarrier`, and accept it will break if math is altered.

I suggest either of the first two to reduce cross-contamination between helpers and math.

src/calliope/backend/parsing.py
```diff
@@ -711,19 +711,16 @@ def extend_equation_list_with_expression_group(
...
-
-        The `exists` array is a boolean array defining the structure of the model and
-        is True for valid combinations of technologies consuming/producing specific
-        carriers at specific nodes.
+        constraint is to be built (`foreach`) and the model `active`, `carrier_in`, and `carrier_out` arrays.
 
-        It is indexed over ["nodes", "techs", "carriers"].
+        The `active` array is a boolean array defining the structure of the model.
+        Values are True for valid combinations of technologies at specific nodes.
+        We extend it to include the carriers they produce and consume (i.e. `carrier_in` and `carrier_out`) to ensure that constraints are only built where relevant.
```
irm-codebase: Shouldn't this detection be in the constraints section of the backend?
Seems like this might be patching some logic flaw in that part of the backend... it also introduces unstable behaviour with math redefinition (i.e., if someone renames `carrier_in` to `energy_input` or something).

src/calliope/schemas/config_schema.py
```diff
@@ -62,6 +62,15 @@ class Init(CalliopeBaseModel):
     Defaults to False to mitigate unexpected broadcasting when applying overrides.
     """
 
+    keep_deactivated: bool = Field(default=False)
+    """
```
irm-codebase: A bit confusing. I'd read this as 'never turn this on' rather than 'keep this in memory even if it is turned off'.


How about `retain_inactive`? it matches the name of the setting this affects, which should help in avoiding confusion.

irm-codebase (changes requested):
Looks good overall, with one caveat.

I think this introduces a potential regression in how `carrier_in` and `carrier_out` are always needed for the `active` matrix to work.

This was not there before, and it cross-contaminates the backend and helpers with math context.

Could you detail why this was added, and whether or not it is avoidable by improving the logic in how constraints are triggered in the backend?

src/calliope/backend/helper_functions.py
brynpickering: It was not agnostic, really, as it used a derived dataset `definition_matrix` which combined `active`/`carrier_in`/`carrier_out`. This hardcoded those on the calliope side so that it didn't have to expose them in the math. I've now updated it so that whenever carrier_in/_out is needed, the math `where` strings include them explicitly. Adds verbosity but makes things explicit.

src/calliope/backend/parsing.py
brynpickering: Fixed to just use `active` as a base mask and everything else is explicit in the math.

docs/user_defined_math/examples/monthly_peak_flow_charge.yaml
```diff
@@ -53,7 +54,7 @@ variables:
     default: 0
     unit: power
     foreach: [nodes, techs, carriers, months]
-    where: monthly_peak_mode
+    where: carrier_out AND monthly_peak_mode
     bounds:
```
irm-codebase: I'm assuming this solves an unseen bug?

docs/user_defined_math/helper_functions.md
```diff
@@ -36,12 +36,20 @@ nodes:
...
 `defined(techs=[tech1, tech2], within=nodes)` would yield a list of `[True, True]` as both nodes define _at least one_ of `tech1` or `tech2`.
 
 `defined(techs=[tech1, tech2], within=nodes, how=all)` would yield a list of `[False, True]` as only `node2` defines _both_ `tech1` and `tech2`.
 
+By default, `defined` searches the model `active` array, which is indexed over `nodes` and `techs` only.
+To check membership of any other dimension, extend it with the `using` argument, giving the boolean input array(s) that define that dimension.
+Multiple arrays are combined with a boolean OR, so must share their dimensions.
+
+`defined(carriers=electricity, within=techs, using=[carrier_in, carrier_out])` would yield a list of `[True, True]` as both technologies define electricity.
+
```
irm-codebase: This is not very clear. From the user side, you are giving arrays of strings. They will become booleans. You are not giving boolean arrays as input, that is a backend thing that users do not need to be aware of.

Just telling people to give a list of elements to check is enough.

docs/user_defined_math/helper_functions.md
```diff
@@ -36,12 +36,20 @@ nodes:
...
+!!! note
+    `using` keeps the helper function agnostic to your math.
+    If you rename or replace the lookups that define a dimension in your own math, update the `using` argument to match; `defined` itself needs no change.
+
 ## sum
```
irm-codebase, line 15: Not needed and self evident. I suggest deleting this `note`.

src/calliope/backend/helper_functions.py
```diff
@@ -264,7 +301,12 @@ def as_math_string(  # noqa: D102, override
             return rf"\bigwedge({', '.join(substrings)})"
```
irm-codebase: Shouldn't `using` be featured in the latex string too?

src/calliope/backend/helper_functions.py
```diff
@@ -274,9 +316,15 @@ def as_array(
         Args:
             within (str): the model dimension to check.
             how (Literal[all, any]): Whether to return True for `any` match of nested members or for `all` nested members.
+            using (xr.DataArray | list[xr.DataArray] | None):
+                Boolean model input array(s) with which to extend the model `active` array along further dimensions.
+                Multiple arrays are combined with a boolean OR, so must share their dimensions.
+                Required to check membership of any dimension other than `nodes`/`techs`,
+                e.g. `using=[carrier_in, carrier_out]` to check `carriers` membership.
+                Defaults to None (only `active` is used).
```
irm-codebase: - 'Boolean model input arrays' is similarly confusing here. Aren't these just lookups? Why not call them that?
- "Multiple arrays are combined with a boolean OR, so **they** must have **the same** dimensions."?

irm-codebase (dismissed):
Consider this approved! I'm highlighting minor comments and improvements.

The only one that might require code tweaks is the latex string in the altered helper function.

irm-codebase (approved):

PR #883, opened 2026-07-25
Feat/uv pixi pr fixes

irm-codebase:
Fixes stuff identified in the review

* Limited python support to 3.12+
* No upper pins
* Moved uv dep to dev environment
* Fixed some comments in the docs
* Removed deprecated files
* Fixed ReadTheDocs build command

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


docs/contributing.md
```diff
@@ -35,28 +35,26 @@ To create a development environment for calliope, use [pixi](https://pixi.sh/lat
...
-1. Run a task in a `pixi` environment
+1. Install the development environment: `pixi install --environment dev`
 
 All together:
 
 ``` shell
 git clone git@github.com:calliope-project/calliope.git
 cd calliope
-pixi run ...
+pixi install --environment dev
```
brynpickering, line 45: you don't _need_ to actively install it. It will be installed on first run that uses it. But if that is the workflow you think makes most sense then I'm happy to go with it

docs/contributing.md
```diff
@@ -193,7 +191,7 @@ We have chosen to use [`ruff`](https://docs.astral.sh/ruff/) for code formatting
 When run from the root directory of this repository, `pyproject.toml` should ensure that formatting and linting fixes are in line with our custom preferences (e.g., 88 character maximum line length).
 The philosophy behind using the `ruff` linter is that it's based on [`black`](https://black.readthedocs.io/en/stable/), an uncompromising formatter that ensures uniform style throughout the project.
 Since `black` is designed to minimise diffs, and make patches more human readable, this also makes code reviews more efficient.
-To make this a smooth experience, you should run `pixi run pre-commit install` after setting up your development environment, so that `ruff` makes all the necessary formatting fixes to your code each time you commit, and so that it also highlights any style issues in your code.
+To make this a smooth experience, you should run `pixi run pre-commit-install` after setting up your development environment, so that `ruff` makes all the necessary formatting fixes to your code each time you commit, and so that it also highlights any style issues in your code.
```
brynpickering: why the dash? I've never seen it like that before

pyproject.toml
```diff
@@ -197,18 +192,17 @@ python = "3.13.*"
 python = "3.14.*"
 
 [tool.pixi.tasks]
-init-dev-env = { cmd = "pre-commit install", default-environment = "dev" }
-test = { cmd = "pytest --no-cov", default-environment = "dev", depends-on = ["init-dev-env"] }
-test-ci = { cmd = "pytest --no-cov -m 'not time_intensive' -m 'not needs_gurobi_license'", default-environment = "dev", depends-on = ["init-dev-env"]  }
-test-cov = { cmd = "pytest --cov", default-environment = "dev", depends-on = ["init-dev-env"] }
+pre-commit-install = { cmd = "pre-commit install", default-environment = "dev" }
```
brynpickering: Answers my question from above. No need for this I don't think. Just delete it and revert to requesting users to call `pre-commit install`

pyproject.toml
```diff
@@ -147,17 +146,15 @@ documentation = "https://calliope.readthedocs.io"
...
-name = "calliope"
 channels = ["conda-forge", "gurobi"]
-platforms = ["linux-64", "osx-64", "osx-arm64", "win-64"]
+platforms = ["linux-64", "osx-arm64", "win-64"]
 preview = ["pixi-build"]
 
 [tool.pixi.dependencies]
-libnetcdf = "*"
-gurobi = "*"
-uv = "*"
```
brynpickering, line 158: My understanding is that pixi needs uv as an explicit dependency to install the local version of calliope. Would not having it here cause issues in the `docs` env? I guess if the RTD has built then that answers the question for me.

README.md
```diff
@@ -37,7 +37,7 @@ Model results can be explored interactively with [Calligraph](https://calligraph
 
 ## Quick start
 
-Calliope can run on Windows, macOS and Linux. Installing from PyPI is quickest with `uv`: `uv pip install calliope`. For local development and full solver/binary environment management, use pixi as described in the documentation.
+Calliope can run on Windows, macOS and Linux. It can be installed using `pixi` (`pixi add calliope`), `conda` (`conda install calliope`), or `uv` (`uv pip install calliope`). For local development, use pixi as described in the documentation.
```
brynpickering: Whenever we make changes in places like this, move new sentences to new lines since that's our preferred way of managing `.md` files

pyproject.toml
irm-codebase: Then we'd have to remove the dependency from other tasks. Are you OK with that?
I do not mind this being done every time, and it's quite benign (it just prints a message every time).

docs/contributing.md
irm-codebase: I just thought it made more sense didactically. I think most folks just want to get set-up, not run full tests or build the docs.

pyproject.toml
irm-codebase: The issue is that you'd likely pass this dependency onto the `conda-forge` item, no?
It's not needed anyhow, and the docs build well :)

pyproject.toml
brynpickering: Yeah, that's fine. I take your point that some people don't want to be burdened with pre-commit locally if they don't have to be.

irm-codebase: @brynpickering 
Honestly, I've come around on this. A particular advantage of forcing this is that there is no `ruff` dependency in dev, which might deviate from the one in `pre-commit`. It saves a _different_ type of headache.

So far I've worked using pre-commit on this repo this way, and it's fine. I'd keep it as-is, honestly

brynpickering: OK!

brynpickering: It's still a bit of a confusing run call given how close it is to what people will know from installing pre-commit in other projects. Maybe don't explicitly mention it as a step in the process? Or change it to `pre-commit install` and we can still be calling pre-commit-install in other tasks in the pyproject.toml

irm-codebase: Implemented!
- Docs now just state the familiar `pre-commit install`.
- the `pre-commit-install` task now redirects stdout to `/dev/null` to silence the output. See more here https://docs.deno.com/runtime/reference/cli/task/#redirects (this is the task shell that `pixi` uses, which works on all platforms). Errors will still show up just fine, too.

brynpickering (approved):

PR #884, opened 2026-07-25
Feat/uv pixi pr warnings

irm-codebase:
Fixes a _lot_ of soon to be depreciated code. We had _25 thousand_ warnings!

I am submitting this PR separately from #883 to keep ourselves sane @brynpickering 
Changes in `src/` need deeper attention. I've added comments to those as an aid.

* Made `xarray` broadcasting more explicit while retaining our behaviour, as the tool will remove / alter some defaults
* Changed tests that used `scope="class"` to be class methods, as suggested in: https://docs.pytest.org/en/stable/deprecations.html#class-scoped-fixture-as-instance-method

## Results

Warnings go from this:

image

To this. The last remaining warning is not our fault, and has already been fixed on `xarray`'s main branch. We'll have to live with it until that is released, then it should go away without any changes on our side.

image

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/backend/backend_model.py
+++ b/src/calliope/backend/backend_model.py
@@ -792,9 +792,12 @@ class BackendModelGenerator(ABC, metaclass=SelectiveWrappingMeta):
         if kwargs:
             func = partial(func, **kwargs)
         vectorised_func = np.frompyfunc(func, len(args), n_out)
-        da = vectorised_func(
-            *(arg.broadcast_like(where) for arg in args), where=where.values
+
+        broadcast_where, *broadcast_args = xr.broadcast(where, *args)
+        outputs = tuple(
+            np.empty(broadcast_where.shape, dtype=object) for _ in range(n_out)
         )
+        da = vectorised_func(*broadcast_args, where=broadcast_where.values, out=outputs)
         if isinstance(da, xr.DataArray):
             da = da.fillna(np.nan)
         else:
--- a/src/calliope/backend/expression_parser.py
+++ b/src/calliope/backend/expression_parser.py
@@ -402,7 +402,16 @@ class EvalComparisonOp(EvalArrayOrMath):
                 op = np.less_equal
             case ">=":
                 op = np.greater_equal
-        constraint = op(lhs_where, rhs_where, where=where.values, dtype=np.object_)
+        broadcast_where, lhs_where, rhs_where = xr.broadcast(
+            where, lhs_where, rhs_where
+        )
+        constraint = op(
+            lhs_where,
+            rhs_where,
+            where=broadcast_where.values,
+            out=(np.empty(broadcast_where.shape, dtype=object),),
+            dtype=np.object_,
+        )
         return xr.DataArray(constraint)
 
 
--- a/src/calliope/backend/expression_parser.py
+++ b/src/calliope/backend/expression_parser.py
@@ -392,9 +392,6 @@ class EvalComparisonOp(EvalArrayOrMath):
                 raise self.error_msg(
                     f"The {side}-hand side of the equation is indexed over dimensions not present in `foreach`: {extra_dims}"
                 )
-        lhs_where = self._apply_where_array(lhs)
-        rhs_where = self._apply_where_array(rhs)
-
         match self.op:
             case "==":
                 op = np.equal
@@ -402,9 +399,7 @@ class EvalComparisonOp(EvalArrayOrMath):
                 op = np.less_equal
             case ">=":
                 op = np.greater_equal
-        broadcast_where, lhs_where, rhs_where = xr.broadcast(
-            where, lhs_where, rhs_where
-        )
+        broadcast_where, lhs_where, rhs_where = xr.broadcast(where, lhs, rhs)
         constraint = op(
             lhs_where,
             rhs_where,
--- a/src/calliope/model.py
+++ b/src/calliope/model.py
@@ -552,7 +552,9 @@ class Model(ModelStructure):
             iteration_results.coords["timesteps"] = new_ts
 
         results_list.append(iteration_results.sel(timesteps=slice(windowstep, None)))
-        results = xr.concat(results_list, dim="timesteps", combine_attrs="drop")
+        results = xr.concat(
+            results_list, dim="timesteps", data_vars="all", combine_attrs="drop"
+        )
         results.attrs["termination_condition"] = ",".join(
             set(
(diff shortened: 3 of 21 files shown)


src/calliope/backend/backend_model.py
```diff
@@ -792,9 +792,12 @@ def _apply_func(
...
         vectorised_func = np.frompyfunc(func, len(args), n_out)
-        da = vectorised_func(
-            *(arg.broadcast_like(where) for arg in args), where=where.values
+
+        broadcast_where, *broadcast_args = xr.broadcast(where, *args)
+        outputs = tuple(
+            np.empty(broadcast_where.shape, dtype=object) for _ in range(n_out)
         )
+        da = vectorised_func(*broadcast_args, where=broadcast_where.values, out=outputs)
         if isinstance(da, xr.DataArray):
```
irm-codebase, line 801: Highlighting this as an important check for the review.

src/calliope/backend/expression_parser.py
```diff
@@ -402,7 +402,16 @@ def as_array(self) -> xr.DataArray:  # noqa: D102, override
...
+        broadcast_where, lhs_where, rhs_where = xr.broadcast(
+            where, lhs_where, rhs_where
+        )
+        constraint = op(
+            lhs_where,
+            rhs_where,
+            where=broadcast_where.values,
+            out=(np.empty(broadcast_where.shape, dtype=object),),
+            dtype=np.object_,
+        )
```
irm-codebase: Highlighting this as an important check for the review.

src/calliope/model.py
```diff
@@ -552,7 +552,9 @@ def _solve_operate(self, solver_config: config_schema.Solve) -> xr.Dataset:
             iteration_results.coords["timesteps"] = new_ts
 
         results_list.append(iteration_results.sel(timesteps=slice(windowstep, None)))
-        results = xr.concat(results_list, dim="timesteps", combine_attrs="drop")
+        results = xr.concat(
+            results_list, dim="timesteps", data_vars="all", combine_attrs="drop"
+        )
```
irm-codebase, line 557: Highlighting this as an important check for the review.

src/calliope/preprocess/model_data.py
```diff
@@ -363,7 +363,7 @@ def _definition_to_ds(
                 )
                 input_data_das.append(self._input_data_to_array(name, validated_data))
             input_data_ds = xr.merge(
-                [input_data_ds, xr.combine_by_coords(input_data_das)],
+                [input_data_ds, xr.combine_by_coords(input_data_das, join="outer")],
                 join="outer",
```
irm-codebase, line 367: Highlighting this as an important check for the review.

src/calliope/backend/backend_model.py
brynpickering: Any performance hit from this?

irm-codebase: I did not notice much, but I also have not done thorough testing.
The best I have ATM is 'tests ran on the same amount of time' when eyeballed.

I could try plugging `memray` and `timeit` later to see if we took a hit.

src/calliope/backend/expression_parser.py
```diff
@@ -402,7 +402,16 @@ def as_array(self) -> xr.DataArray:  # noqa: D102, override
                 op = np.less_equal
             case ">=":
                 op = np.greater_equal
-        constraint = op(lhs_where, rhs_where, where=where.values, dtype=np.object_)
+        broadcast_where, lhs_where, rhs_where = xr.broadcast(
+            where, lhs_where, rhs_where
+        )
```
brynpickering: If doing this broadcast here then we don't need to do it above. Pass `lhs` and `rhs` in directly here

src/calliope/backend/expression_parser.py
brynpickering: again, would be good to know if we have any changes in peak memory due to this. I doubt it since we have about as many large arrays sitting around as we did previously. However, here we have an `np.empty` array that we didn't have before.

src/calliope/model.py
brynpickering: Go for `minimal` so we don't get the `timestep` dimension added to arrays where we don't need to have it

irm-codebase: Perhaps we should open an issue for this explicitly.
I'm just matching what the software used to do (the default was `all`). Setting minimal causes `operate` failures:

Image

brynpickering: ok, stick with it then. I must have forgotten how it works in this case. This is why making kwargs explicit is quite good!

src/calliope/backend/expression_parser.py
irm-codebase: Tested with a broadcast of 8,388,608 elements (128, 256, 256) using pytest-memray.

Memory went _down_, actually!
[aggro_test_new.txt](https://github.com/user-attachments/files/30407746/aggro_test_new.txt)
[aggro_test_old.txt](https://github.com/user-attachments/files/30407747/aggro_test_old.txt)

src/calliope/backend/backend_model.py
irm-codebase: Done. No differences in memory when testing a 'where' array with 4194304 elements (64, 256, 256).


[aggro_test_backend_model_new.txt](https://github.com/user-attachments/files/30408067/aggro_test_backend_model_new.txt)
[aggro_test_backend_model_old.txt](https://github.com/user-attachments/files/30408068/aggro_test_backend_model_old.txt)

irm-codebase:
Did some thorough testing. Overall, it seems like this change is actually _slightly_ beneficial.
Here are the test files i used for this, with the help of `pytest-memray`.

Unfortunately this does not work for Windows machines, so you'll have to remove it from `pyproject.toml` if you want to test it.
Place these scripts in `tests/backend/`, then run like so:

```bash
CALLIOPE_MEMRAY_STRESS=1 pixi run -e dev pytest -n0 --no-cov --memray \
    tests/backend/backend_model_memory_test.py::test_backend_model_generator_apply_func_memory > output/test.txt
```

[backend_model_memory_test.py](https://github.com/user-attachments/files/30408166/backend_model_memory_test.py)
[expression_parser_memory_test.py](https://github.com/user-attachments/files/30408168/expression_parser_memory_test.py)

brynpickering (approved):

PR #889, opened 2026-08-15
build: bump numpy lower bound to 2.0

### LINKED ISSUE #888: [Compatibility Issue] Missing numpy>=2 constraint — np.typing crashes on numpy<2
## Affected File

`src/calliope/backend/where_parser.py:20`

## Current Code

```python
from __future__ import annotations  # PEP 563 — only protects annotations

BOOLEANTYPE = np.bool_ | np.typing.NDArray[np.bool_]  # runtime | expression
```

The `|` operator on types is a **runtime expression**, not a type annotation. PEP 563 does not protect it.

## Root Cause

`numpy.typing` is only exposed as a module attribute in **numpy >= 2.0.0**. On numpy < 2.0, `np.typing.NDArray` triggers:

```
AttributeError: module 'numpy' has no attribute 'typing'
```

`numpy.typing` becomes accessible on numpy<2 if any code in the import chain has executed `import numpy.typing` before the access — once imported, it is injected into `numpy.__dict__` and all subsequent `np.typing.XXX` accesses work.

## Dependency

`numpy >= 1` — allows any numpy version. Transitive dependencies do not force numpy>=2:

```
$ pip install --dry-run -r deps.txt numpy<2
  → numpy 1.26.4 ✓  (all constraints satisfied)
```

Reproduced on numpy 1.26.4:

```
>>> BOOLEANTYPE = np.bool_ | np.typing.NDArray[np.bool_]
AttributeError: module 'numpy' has no attribute 'typing'
```

## Upstream Dependency Protection

In normal calliope usage, this crash is incidentally prevented because **xarray** (`xarray >= 2026.4`) executes `from numpy.typing import ArrayLike/DTypeLike` at module level in `xarray.core.types`, `xarray.core.indexing`, and `xarray.groupers` — all loaded at `import xarray` time. Since calliope imports xarray before loading the backend, `np.typing` is already available when `where_parser.py` runs.

However, this protection is not guaranteed: it depends on xarray's internal implementation, which could change in a future version. Direct imports of `where_parser` before xarray is loaded would also bypass the protection.

## Solution

Adding `import numpy.typing` before the `BOOLEANTYPE` assignment provides a self-contained defense regardless of import order or upstream dependency behavior:

```python
import numpy.typing
BOOLEANTYPE = np.bool_ | np.typing.NDArray[np.bool_]
```

Or tighten the numpy constraint in `pyproject.toml`:

```python
"numpy >= 2",  # was: "numpy >= 1"
```

## References

- NumPy 2.0 release notes — `numpy.typing` exposed as module attribute
- Similar fix: [numexpr#540](https://github.com/pydata/numexpr/issues/540)

xyf5432:
Fixes #888

## Problem

`where_parser.py` evaluates a `np.typing.NDArray` expression at import time (`BOOLEANTYPE = np.bool_ | np.typing.NDArray[np.bool_]`). On numpy < 2.0, `np.typing` is only accessible after an explicit `import numpy.typing`, so the declared floor `numpy >= 1` admits environments where that expression would only resolve through an incidental import.

As discussed in #888, this is mostly latent in practice — `where_parser` imports xarray, which imports `numpy.typing` at startup — but the declared constraint does not reflect the numpy versions the project actually supports and tests against.

## Fix

Bump the declared floor to match the numpy versions the project already locks and tests against:

```diff
-  "numpy >= 1",
+  "numpy >= 2.0",
```

This aligns `pyproject.toml` with the current `pixi.lock` (numpy 2.5.1), so no lock regeneration is needed. Pip resolvers now reject numpy < 2.0 up front instead of relying on the incidental `numpy.typing` import provided by xarray.

## Verification

- The project's `pixi.lock` already pins numpy 2.5.1, which satisfies the new floor — the lock remains valid and no environment actually resolves to numpy < 2.0.
- This is a constraint-only change; no code paths are affected.


brynpickering (dismissed):

brynpickering:
Before I can merge this @xyf5432 you will need to refresh the pixi.lock file

xyf5432:
Done — I ran `pixi lock` and pushed the refreshed lock file (commit 80f646c). The only change is the in-lock snapshot of calliope's dependency metadata: `numpy>=1` → `numpy>=2.0` (the lock's `requires_dist` for the local editable calliope package). No package versions changed.

Confirmed the lock now matches the manifest — `pixi lock` produces no further diff.

irm-codebase (approved):

irm-codebase:
@allcontributors please add @xyf5432 for bug

PR #893, opened 2026-09-01
Add operate mode to the docs ref page

brynpickering:
Fixes missing operate mode in ref docs

* Slight change to latex math to handle cases where an item has been removed but we want to keep its description around with a "removed" flag.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/src/calliope/math/operate.yaml
+++ b/src/calliope/math/operate.yaml
@@ -75,14 +75,6 @@ global_expressions:
     equations:
       - expression: sum(cost_operation_variable, over=timesteps)
 
-objectives:
-  min_cost_optimisation:
-    description: >-
-      Minimise the total cost of operating all technologies in the system.
-      If multiple cost classes are present (e.g., monetary and co2 emissions),
-      the weighted sum of total costs is minimised.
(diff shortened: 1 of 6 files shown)


src/calliope/math/operate.yaml
```diff
@@ -75,6 +75,14 @@ global_expressions:
...
       - expression: sum(cost_operation_variable, over=timesteps)
 
+objectives:
+  min_cost_optimisation:
+    description: >-
+      Minimise the total cost of operating all technologies in the system.
+      If multiple cost classes are present (e.g., monetary and co2 emissions),
+      the weighted sum of total costs is minimised.
+      Cost class weights can be defined in the indexed parameter `objective_cost_weights`.
+
```
irm-codebase: can you clarify why this was added?
Seems benign, I just find it strange.

irm-codebase (dismissed):
Looks good. I just have a small question regarding the modification of the math file.

irm-codebase (approved):
Looks good!
A nice to have would be to perhaps sort the extra math in alphabetical order, but it's not necessary for release.

PR #894, opened 2026-09-01
Various minor fixes in migration guide

brynpickering:
## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


irm-codebase (approved):
Changes look good, but maybe in the future we should switch to a link checker that has more people behind.
But only if it's worth it.

PR #898, opened 2026-09-09
Feat/highs speedups

sjpfenninger:
Fixes #

* Shadow price extraction in the HiGHS backend no longer iterates over all constraint elements. This was causing severe slowdowns, making the HiGHS backend unusable for large models with shadow price extraction.
* On top of that, this branch also batches variable and constraint construction as an extra optimisation. The HiGHS backend is still slightly slower than the Gurobi one, and the slowdown worsens the larger the model. For the national-scale built-in model (full time series) it's about 5% slower.

## Reviewer checklist

- [ ] Documentation updated
- [ ] Changelog updated


src/calliope/backend/highs_backend_model.py
```diff
@@ -83,6 +84,53 @@ def _set_instance_options(self) -> None:
...
 
+    @staticmethod
+    def _broadcast_to_mask(
+        where: xr.DataArray, *args: xr.DataArray
+    ) -> tuple[xr.DataArray, np.ndarray, list[np.ndarray]]:
+        """Broadcast `where` against `args` and flatten the results.
+
+        Mirrors the broadcasting done by `_apply_func`, so that components added to
+        HiGHS in batches end up in arrays with the same shape, dims and coords as an
+        element-wise apply would produce.
```
irm-codebase, line 95: Two things: 
- Wouldn't other backends benefit from this approach, if it is more efficient?
If it can be demonstrated (e.g., via `memray`, we should consider moving this to the parent class.
- Let us avoid referencing other functions in comments (`_apply_func`). Names change and this is bound to become obsolete.

src/calliope/backend/highs_backend_model.py
```diff
@@ -83,6 +84,53 @@ def _set_instance_options(self) -> None:
...
+            broadcast_where,
+            broadcast_where.values.ravel(),
+            [arg.values.ravel() for arg in broadcast_args],
+        )
+
+    @staticmethod
+    def _scatter_objects(
+        template: xr.DataArray, mask: np.ndarray, objs: list
+    ) -> xr.DataArray:
+        """Place backend objects at the `mask` positions of an otherwise-NaN array.
```
irm-codebase, line 118: Another generic function that perhaps move upwards if other backends would benefit from it.

src/calliope/backend/highs_backend_model.py
```diff
@@ -93,11 +141,38 @@ def _add_variable(  # noqa: D102, override
...
+        first_col = self._instance.getNumCol()
+        status = self._instance.addCols(
+            n_new,
+            np.zeros(n_new),
+            np.asarray(lb_vals[mask], dtype=float),
+            np.asarray(ub_vals[mask], dtype=float),
+            0,
+            np.zeros(n_new, dtype=np.int32),
+            np.empty(0, dtype=np.int32),
+            np.empty(0),
```
irm-codebase, line 158: Trying to summarise this to see if I got the behaviour right.
- total new variables to add
- objective coefficient (0).
- lower bound
- upper bound
- coefficient in constraints (0) 
- constraint coefficient rows indices (none)
- coefficient values (none).

Basically, this puts the variables in the model, does not initialise or set coefficients for them in the objective or the constraint matrix. Kind of 'non-parametric' variable initialisation.

src/calliope/backend/highs_backend_model.py
```diff
@@ -93,11 +141,38 @@ def _add_variable(  # noqa: D102, override
     ) -> xr.DataArray:
         lb = self._get_variable_bound(bounds.min, name, references)
         ub = self._get_variable_bound(bounds.max, name, references)
-        var = self._apply_func(
-            self._instance.addVariable, where, 1, lb, ub, type=domain_type
+        template, mask, (lb_vals, ub_vals) = self._broadcast_to_mask(where, lb, ub)
+
+        # Variables are added in one batch: `addVariable` allocates arrays and queries
+        # the model on every call, which dominates build time on large models.
```
irm-codebase, line 147: Do not refer to code in comments (`addVariable`) to avoid confusion in the future.
It's fine to just state that this is more efficient in HiGHS, and state what this actually does (create non-parametric variables in the model).

src/calliope/backend/highs_backend_model.py
```diff
@@ -108,19 +183,90 @@ def _add_global_expression(  # noqa: D102, override
...
+    def _to_csr(
+        lengths: np.ndarray, idxs: list[list[int]], vals: list[list[float]]
+    ) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+        """Flatten per-row column indices and coefficients into a HiGHS CSR matrix.
+
+        Entries are sorted by (row, column) and duplicate columns within a row are
+        summed, mirroring `highs_linear_expression.unique_elements`. HiGHS rejects a
+        row that references the same column twice, so the merge is required, not an
+        optimisation.
+
```
irm-codebase, line 196: Please adapt AI over explained text into actual docstrings.


```suggestion
         Entries are sorted by (row, column) and duplicate columns within a row are summed  (3x + 3x + 3y = 6x + y).
        Otherwise, HiGHS would reject a row that references the same column twice.
```

irm-codebase: Github suggestions are so, so broken nowadays :(

src/calliope/backend/highs_backend_model.py
```diff
@@ -108,19 +183,90 @@ def _add_global_expression(  # noqa: D102, override
...
+        starts, cols, coefs = self._to_csr(lengths, idxs, vals)
+        # Constraints are added in one batch: `addConstr` re-sorts each row's
+        # coefficients and re-queries the model on every call.
+        first_row = self._instance.getNumRow()
+        status = self._instance.addRows(
+            n_new, lower, upper, coefs.size, starts, cols, coefs
+        )
+        if status != highspy.HighsStatus.kOk:
+            # E.g. a coefficient whose absolute value is below `small_matrix_value`,
+            # which HiGHS drops, flagging it with a warning status.
```
irm-codebase, line 257: remove. Self evident.

src/calliope/backend/highs_backend_model.py
```diff
@@ -108,19 +183,90 @@ def _add_global_expression(  # noqa: D102, override
...
+        template, mask, (expr_vals,) = self._broadcast_to_mask(where, expression)
+        exprs = expr_vals[mask]
+        n_new = exprs.size
+
+        lower = np.empty(n_new)
+        upper = np.empty(n_new)
+        lengths = np.empty(n_new, dtype=np.int64)
+        idxs: list[list[int]] = []
+        vals: list[list[float]] = []
+        for idx, expr in enumerate(exprs):
```
irm-codebase, line 236: Another attempt of trying to summarise what this is doing.

This is the row equivalent of `_add_variable` above (which uses `addCols`).
- fetches the bonds from the expression and builds its compressed sparse row matrix equivalent (to fit HiGHS). This is a 'squashed' version of the matrix constructed by the model, built for higher efficiency. Note that if the problem is very 'dense' this will be less efficient than regular notation. In most applications sparse will win, though.
  - starts: where each constraint 'begins'
  - cols: which variables are referenced ($3x_1 + 2x_3$ -> [1, 3])
  - coefs: the coefficients ($3x_1 + 2x_3$ -> [3, 2])
- get the current number of rows (constraints) already in the model to make sure we do not forget / overwrite them.
- Add the set of new constraints in one go via `addRows`
  - n_new: number of constraints to add
  - lower / upper: constraint bounds
  - coefs.size: total number of coefficients being added.
  - starts: where each constraint starts
  - cols: variable index for the coefficient in each constraint
  - coefs: value of each coefficient
- Check if all is well, blow up otherwise.
- Return the location of the new constraints in the backend so we can extract results later.

irm-codebase (changes requested):
Looks good. Had to do some brain churning to understand the method.
Overall it looks good. I trust our tests catch poor implementation thanks to our static result checkers.

Requests:
- Modify comments and docstrings to get rid of chaff
- Make sure test coverage is complete.

